PR #10, opened 2018-12-23
Mainly in-place operations with some flake8

prisae:
I made a couple of changes to `LinearOperator.py` and `basicoperators/*`. Feel free to use them or ignore them. All tests passed on my machine.

### In-place statements

There where many statements of the type
```
x = x + y
```
which I replaced by 
```
x += y
```
The first one creates a new object `x`, whereas the second one does the operation in-place, saving potentially time and memory. You can see in this little example that the ID of `x` changes if you do `x = x+1`, but it doesn't if you do `x += 1`:
```
In [1]: x = np.arange(10)

In [2]: id(x)
Out[2]: 140703335921344

In [3]: x += 1

In [4]: id(x)
Out[4]: 140703335921344

In [5]: x = x+1

In [6]: id(x)
Out[6]: 140703344097440
```
(The same applies for `-`, `*`, and `/`.)

### Replace `flatten()` by `ravel()`
`flatten` always returns a copy, whereas `ravel` only returns a copy if necessary. So I think it is preferably to use `ravel`.

### Some pep8 stuff
I took the opportunity to run it through `flake8`. The only error-reporting left in said files are the unused imports in `basicoperators/__init__.py`. Changes made:
- File length. You told me you plan to change this anyway in the future.
- Comments starting with `#text` instead of `# text` (space after `#`).
- Missing white space around operators (e.g., `x=y` instead of `x = y`).
- One empty line when two are required and vice versa.
- Multiple statements on one line, e.g.:
  `if x: y = np.sqrt(x)`
  changed to
  ```
  if x:
      y = np.sqrt(x)
  ```


mrava87:
Looks good to me. Thanks for making these changes :)

PR #11, opened 2018-12-23
Change default dtype behaviour

prisae:
As discussed.

My suggestion is two-fold.

## 1. If input contains an `ndarray`
Set default to `None`, and if `None`, take the `dtype` of the input array. Example where the input contains the matrix `A`.
```
if dtype is None:
    self.dtype = A.dtype
else:
    self.dtype = np.dtype(dtype)
```

## 2. If input does not contain an `ndarray`
Change default from `float32` to `float64`. This is the default in `NumPy`/`SciPy`, so users most likely expect `float64` unless they specifically specify it otherwise.

### DIFF
--- a/pylops/basicoperators/CausalIntegration.py
+++ b/pylops/basicoperators/CausalIntegration.py
@@ -71,7 +71,7 @@ class CausalIntegration(LinearOperator):
 
     """
     def __init__(self, N, dims=None, dir=-1, sampling=1,
-                 halfcurrent=True, dtype='float32'):
+                 halfcurrent=True, dtype='float64'):
         self.N = N
         self.dir = dir
         self.sampling = sampling
--- a/pylops/basicoperators/FirstDerivative.py
+++ b/pylops/basicoperators/FirstDerivative.py
@@ -39,7 +39,7 @@ class FirstDerivative(LinearOperator):
         y[i] = (0.5x[i+1] - 0.5x[i-1]) / dx
 
     """
-    def __init__(self, N, dims=None, dir=0, sampling=1., dtype='float32'):
+    def __init__(self, N, dims=None, dir=0, sampling=1., dtype='float64'):
         self.N = N
         self.dir = dir
         self.sampling = sampling
--- a/pylops/basicoperators/Laplacian.py
+++ b/pylops/basicoperators/Laplacian.py
@@ -2,7 +2,7 @@ import numpy as np
 from pylops.basicoperators import SecondDerivative
 
 
-def Laplacian(dims, dirs=(0, 1), weights=(1, 1), sampling=(1, 1), dtype='float32'):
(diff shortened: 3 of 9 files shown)


mrava87:
Overall agree with the idea (and definitely with the swap from float32 to float64 to be consistent with ``Numpy/Scipy``). 

I need to give it a deeper thought to *1. If input contains an ndarray* as in some cases this may not always be true. Take for example the ``MatrixMult`` operator,  the matrix ``A`` may be ``float32`` for example but this does not mean that the input could not be ``complex64`` and so ``dtype`` should be ``complex64``. Not sure what is better in this case?

Let me take a more detail look at the actual ``scipy.sparse.linalg.LinearOperator`` as I am a bit unsure how important is ``dtype`` (feel free to do the same) and then we can decide which way to go :)

prisae:
From what I saw is that if `dtype` is `None` or is not defined in a `LinearOperator`, `scipy` runs the `LinearOperator` when it's created to define the `dtype`. So you have an overhead of one call to the operator. If this is OK then you can leave it completely out. Otherwise you would have to implement something along the lines of what I suggested, or simply set `float64` as default.

mrava87:
I looked into it and effectively the only time where an extra forward is carried out is for the _CustomLinearOperator class which is used when you provide explicit functions for matvec and rmatvec at initialisation (not when you subclass LinearOperator like we do in PyLops).

However, if you look at how the adjoint is implemented (shortcut .H), LinearOperator will implicitly create a _CustomLinearOperatorclass under the hood and perform an extra computation EVERY time you call the adjoint unless you call it directly with rmatvec/_rmatvec. This is very bad and so I agree we should avoid allowing None to be used for dtype. Luckily scipy solvers do not invoke the adjoint with .H* like I do in some example to make things more expressive and so they do not add this extra overhead even if dtype=None.

Based on this I can make a new pull request with the following (since this one has now a few conflicts):

- always set to float64 as default and let the user do more clever things if needed (we avoid extra checks and computations that may not always do the right thing anyways)
- look into the optimisation module and replace * with matvec and .H* with rmatvec for speedup (readability is not the priority here as the inner working of these methods is not exposed to the users)
- keep * and .H* in examples and tutorials acknowledging for readability purposes (although it has an extra cost). Show the various ways of calling forward and adjoint in a tutorial.

Good we got to look into this deeper, we may be able to get some speedup in some of the solvers almost for free ;)

prisae:
> Based on this I can make a new pull request with the following
Sounds good to me!

mrava87:
Just implemented the changes discussed above here https://github.com/mrava87/pylops/commit/d31eee5c6a988fcc795895af5a902ceeb8ae7fc3 and added tutorial to the functionalities of LinearOperator (scipy + additional ones in PyLops) here https://github.com/mrava87/pylops/commit/ea03d881c2e14726a027da761c320a80d41a662e#diff-b6ffad6ca319c003ad770ca6592a6911

PR #38, opened 2019-01-14
Creates FunctionOperator, a useful wrapper for operators defined by generic functions

cako:
Create LinearOperators quickly by calling:

    FunctionOperator(f, n)
    FunctionOperator(f, nr, nc)
    FunctionOperator(f, fc, nc)

or

    FunctionOperator(f, fc, nr, nc)

### DIFF
--- a/pylops/LinearOperator.py
+++ b/pylops/LinearOperator.py
@@ -17,7 +17,6 @@ class LinearOperator(spLinearOperator):
 
     """
     def __init__(self, explicit=False):
-        super(LinearOperator, self).__init__()
         self.explicit = explicit
 
     def div(self, y):
--- a/pylops/__init__.py
+++ b/pylops/__init__.py
@@ -8,6 +8,8 @@ from .basicoperators import Diagonal
 from .basicoperators import Flip
 from .basicoperators import Symmetrize
 
+from .basicoperators import FunctionOperator
+
 from .basicoperators import VStack
 from .basicoperators import HStack
 from .basicoperators import Block
--- a/pylops/basicoperators/FunctionOperator.py
+++ b/pylops/basicoperators/FunctionOperator.py
@@ -90,7 +90,6 @@ class FunctionOperator(LinearOperator):
             self.fc = args[0]
             self.shape = args[1:3]
 
-    
     def _matvec(self, x):
         return self.f(x)
 
--- a/pylops/basicoperators/FunctionOperator.py
+++ b/pylops/basicoperators/FunctionOperator.py
@@ -1,0 +1,102 @@
+import numpy as np
+from numbers import Integral
+from pylops import LinearOperator
+
+
+class FunctionOperator(LinearOperator):
+    r"""Function Operator.
+
+    Simple wrapper to functions for forward `f` and adjoint `fc`
+    multiplication.
+
+    Functions :math:`f` and :math:`fc` respect
+    :math:`f:\mathbb{F}^n \to \mathbb{F}^m` and
+    :math:`fc:\mathbb{F}^m \to \mathbb{F}^n` where :math:`\mathbb{F}` is
+    the appropriate underlying type.
+
+    Parameters
+    ----------
+    f : :obj:`callable`
+        Function for forward multiplication.
+    fc : :obj:`callable`, optional
+        Function for adjoint multiplication.
+    nr : :obj:`int`
+        Number of "rows" (length of output vector).
+    nc : :obj:`int`, optional
+        Number of "columns" (length of input vector).
+    dtype : :obj:`str`, optional
+        Type of elements in input array.
+
+    It can be called in the following ways:
+
+    FunctionOperator(f, n)
+    FunctionOperator(f, nr, nc)
+    FunctionOperator(f, fc, nc)
+    FunctionOperator(f, fc, nr, nc)
+
+    The first two methods will return `NotImplementedError` is the adjoint is
+    The first and third method assume the matrix (or matrices) are square.
+    called.
+    All methods can be called with the `dtype` keyword argument.
+
+    Attributes
+    ----------
+    shape : :obj:`tuple`
+        Operator shape (nr, nc)
+    explicit : :obj:`bool`
+        Operator contains a matrix that can be solved explicitly (``True``) or
+        not (``False``)
+
+    Examples
+    --------
+    >>> from pylops.basicoperators import FunctionOperator
+    >>> def forward(v):
+    ...     return np.array([2*v[0], 3*v[1]])
+    ...
+    >>> A = FunctionOperator(forward, 2)
+    >>> A
+    <11x11 FunctionOperator with dtype=float64>
+    >>> A.matvec(np.ones(2))
+    array([2.,  3.])
+    >>> A @ np.ones(2)
+    array([2.,  3.])
+    """
+
+    def __init__(self, f, *args, **kwargs):
+        try:
+            self.dtype = kwargs['dtype']
+        except KeyError:
+            self.dtype = 'float64'
+        self.explicit = False
+
+        super().__init__()
+
+        self.f = f
+
+        # call is FunctionOperator(f, m)
+        if len(args) == 1:
+            self.shape = (args[0], args[0])
+            self.fc = None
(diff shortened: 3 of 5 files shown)


mrava87:
Nice :) how does this compare to the actual scipy one which has the option to pass functions instead of subclassing, see in examples in this page https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.LinearOperator.html?

I guess your one gets all the extra bit of our linear operator (doc, eigs...)?

cako:
1. 

> Nice :) how does this compare to the actual scipy one which has the option to pass functions instead of subclassing, see in examples in this page https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.LinearOperator.html?
> 
> I guess your one gets all the extra bit of our linear operator (doc, eigs...)?

Pretty much just has the extra bits. The functionality of the class could also be simply implemented as a utility function (all the hard work is done in `__init__`), but it was a good exercise to try and implement this! I still haven't decided whether I should call `super()super().__init__` inside my class (thereby initializing `scipy.sparse.linalg.LinearOperator` in order to validate `shape` and `dtype` but at the moment I am leaning against this.


> I guess you fixed to code but not the comment was ok before, with ‘overdetermined’?

Yup, you are right, I'll change it.

mrava87:
Sounds good :) I think this is useful to have especially as it is like the scipy one (and agree about leaving super.super out for the moment)

We can add it in the documentation next to LinearOperator as it is very generic, can be used for anything, agree? But I would suggest that in the docstring or somewhere we explain that this can be used for very simple operators where forward and adjoint are for example lambda function.  Things like this 

G = np.matrix(np.random.normal(0, 1, (par['nr'], par['nc'])) +
                  np.random.normal(0, 1, (par['nr'], par['nc']))*par['imag'],
                  dtype=par['dtype'])

     def forward_f(x):
        return G @ x
    def adjoint_f(y):
        return G.H @ y
can actually be very nasty to debug because all of a sudden you have things outside of the operator modifying the operator, which is why I found the subclassing part nicer than passing functions directly. I just want to ensure that none develops an operator not using subclassing... what do you think?

Btw, I am going to change LinearOperator into _LinearOperator soon as this is not effectively for users, you can only use it as parent class and not call it directly, makes sense?

cako:
> But I would suggest that in the docstring or somewhere we explain that this can be used for very simple operators where forward and adjoint are for example lambda function.

Yes, lambda functions are very good for this case, but also not great at debugging. That's why I decided to go with a `def` in this case. Now that the code works, we can revert to lambda :stuck_out_tongue_winking_eye: 

> all of a sudden you have things outside of the operator (G) which if you modify you also are modifying the operator behavior (?)

True, I had not thought of this. Overall I would argue it can be very practical, especially for cases where your linear function depends on other arguments, so you can do `FunctionOperator(lambda x: f(x, arg1, arg2).flatten(), n)` for example. I think I will add something in the documentation to this effect.

> Btw, I am going to change LinearOperator into _LinearOperator soon as this is not effectively for users, you can only use it as parent class and not call it directly, makes sense?

Julia library [LinearMaps](https://github.com/Jutho/LinearMaps.jl) makes a call to `LinearMaps` (equivalent to `LinearOperator`) actually be a call to `FunctionMap`. I don't really like this approach because it masks what is going on and you have do dig deep sometimes to find who initializes what. On the other hand, I am not sure `_LinearOperator` should be "hidden", the scipy one isn't, for example. Maybe renaming to `PyLinearOperator` could avoid confusion. I think it is really up to the documentation to let people know that they don't have to create their linear operators from scratch (aka, subclass LinearOperator), and can use the ready-made ones.

mrava87:
Just a stupid question, if you have ``FunctionOperator(f, n)`` can we not just assume fc=f, otherwise why would you allow a user to call this if we just return an error?

For ``FunctionOperator(f, n, m)`` instead I would say that if m!=n then self-adjoint doesn't make sense so we raise an error but if n==m again we can force fc=f?

mrava87:
Final question, any reason for having a main in ``test_functionoperator`` file. ``pytest`` compared to ``unittest`` does not require mains running the classes, unless you have a reason for this I suggest we remove it :)

cako:
> Just a stupid question, if you have `FunctionOperator(f, n)` can we not just assume fc=f, otherwise why would you allow a user to call this if we just return an error?
> 
> For `FunctionOperator(f, n, m)` instead I would say that if m!=n then self-adjoint doesn't make sense so we raise an error but if n==m again we can force fc=f?

My first iteration of the function did exactly that. But hard-coding this means that whenever methods find `_rmatmul`, they will be tempted to use it, implicitly assuming that the operator is symmetric. Instead, I think better behavior is to simply say: the user has not coded this, this method cannot be used. "Explicit is better than implicit" :stuck_out_tongue:

cako:
> Final question, any reason for having a main in `test_functionoperator` file. `pytest` compared to `unittest` does not require mains running the classes, unless you have a reason for this I suggest we remove it :)

I thought I had gotten rid of this! It is pretty useful for running only those specific tests with:

    python test_functionoperator.py

I am sure pytests has a better way to do it–I just don't know it!

mrava87:
Ahhaah no worries, I can get rid of it :)

``pytest pytests/test_functionoperator.py`` does the trick

cako:
> > Just a stupid question, if you have `FunctionOperator(f, n)` can we not just assume fc=f, otherwise why would you allow a user to call this if we just return an error?
> > For `FunctionOperator(f, n, m)` instead I would say that if m!=n then self-adjoint doesn't make sense so we raise an error but if n==m again we can force fc=f?
> 
> My first iteration of the function did exactly that. But hard-coding this means that whenever methods find `_rmatmul`, they will be tempted to use it, implicitly assuming that the operator is symmetric. Instead, I think better behavior is to simply say: the user has not coded this, this method cannot be used. "Explicit is better than implicit"

Btw, although I haven't tested this in Python, ARPACK `eigs` should still work (at least with the LM option), as well as several other Krylov-based methods (e.g. CG)

mrava87:
> > Just a stupid question, if you have `FunctionOperator(f, n)` can we not just assume fc=f, otherwise why would you allow a user to call this if we just return an error?
> > For `FunctionOperator(f, n, m)` instead I would say that if m!=n then self-adjoint doesn't make sense so we raise an error but if n==m again we can force fc=f?
> 
> My first iteration of the function did exactly that. But hard-coding this means that whenever methods find `_rmatmul`, they will be tempted to use it, implicitly assuming that the operator is symmetric. Instead, I think better behavior is to simply say: the user has not coded this, this method cannot be used. "Explicit is better than implicit" 😛

Agree. So in docstring shall we just tell the user that they can run:

``FunctionOperator(f, fc, n)`` for symmetric operators
``FunctionOperator(f, fc, n, m)`` for rectangular operators

cako:
The possible calls are like this:
 
    FunctionOperator(f, n) – square operators with an unknown transpose
    FunctionOperator(f, nr, nc) – rectangular operators with an unknown transpose
    FunctionOperator(f, fc, n) – square operators with a known transpose
    FunctionOperator(f, fc, nr, nc) – rectangular operators with a known transpose

The last call is the "ideal" call. Oh, and all calls can also take `dtype`. If not provided, `float64` is used.

mrava87:
What I mean is last and third work (or at least every call can break at runtime if the method is not consistent with nr and nc...) but the first two we basically know already that they won't work... so why even suggesting them to the user, or am I missing something?

cako:
Basically if you don't know the adjoint (or don't care about it), you can use the first two. It's not the most common thing, but thinking practically, it seems very restrictive to require users to know adjoints all the time. Sure, they are limited in using methods which don't require the adjoint, but there are more than a few of these. The idea of throwing the error is that calls to the adjoint are not passed silently, rather, they alert the user. I was even thinking of creating a specific error (subclassed from NotImplementedError) just for adjoints.

mrava87:
Ok I get it now. I always think that ultimately you want to solve an inverse problem and every solver requires forward and adjoint (of course this is same if self-adjoint but you need both anyways). But I agree, no harm to keep the first two if you just want to use FunctionOperator to apply an operator :)

PR #116, opened 2019-11-06
Add favicon

prisae:
Add a favicon to the manual. Feel free to replace the actual `favicon.ico` with a better one.

Having a favicon can help if you have many tabs open, so you see at once where you have the docs ;-)

Selection_002


mrava87:
Thanks, good idea :)

PR #127, opened 2019-12-14
Fix versioning

prisae:
Currently, installing pylops either with
```
pip install git+https://github.com/equinor/pylops
```
or cloning the repo and running in the repo
```
pip install -e .
```
results in the following in Python
```
>>> import pylops
>>> pylops.__version__
'unknown-20191214'
```

This is fixed (at least for me) by using setup from `setuptools` instead of from `distutils.core`.


prisae:
There is another possible solution. Change all content in `requirements.txt` and replace it by `.`.

PR #156, opened 2020-05-28
Tosparse

twhughes:
This PR adds a `tosparse` method to the `LinearOperator` class.

The basic  premise is the same as `todense`, except the method returns a `scipy.sparse.csr_matrix` instead of a dense `numpy.ndarray`.

Having an explicit, sparse matrix representation of a linear operator is necessary for performing direct solves of sparse linear systems.

The sparse matrix of linear operator 'A' is created by looping through the unit vectors for the columns of A, multiplying each of these unit vectors by A, and storing the non zero values and their indices.  In this way, no dense representation of A is ever created.


pylops/LinearOperator.py
```diff
@@ -143,6 +144,55 @@ def todense(self):
...
+
+        Returns
+        -------
+        matrix : :obj:`scipy.sparse.csr_matrix`
+            Sparse matrix.
+
+        """
+
+        Op = aslinearoperator(self)
+        (m, n) = self.shape
```
mrava87, line 163: Codacy Issue found: [Unused variable 'm'](https://app.codacy.com/manual/mrava87/pylops/pullRequest?prid=5603693)

mrava87:
Fantastic, this looks good to me and very useful :)

PR #170, opened 2020-09-30
Substitute x / np.abs(x) by np.sign(x)

cako:
Ran into floating point exeptions using ISTA and FISTA because of the division in `_softthreshold`. Changed `_shrinkage` as well which did have protection from division by zero, but was ugly.

### DIFF
--- a/pylops/optimization/sparsity.py
+++ b/pylops/optimization/sparsity.py
@@ -181,20 +181,6 @@ def _halfthreshold_percentile(x, perc):
     #return _halfthreshold(x, (2. / 3. * thresh) ** (1.5))
     return _halfthreshold(x, (4. / 54 ** (1. / 3.) * thresh) ** 1.5)
 
-def _shrinkage(x, thresh):
-    r"""Shrinkage.
-
-    Applies shrinkage to vector ``x``.
-
-    Parameters
-    ----------
-    x : :obj:`numpy.ndarray`
-        Vector
-    thresh : :obj:`float`
-        Threshold
-    """
-    return np.maximum(np.abs(x) - thresh, 0.) * np.exp(1j * np.angle(x))
-
 
 def _IRLS_data(Op, data, nouter, threshR=False, epsR=1e-10,
                epsI=1e-10, x0=None, tolIRLS=1e-10,
@@ -1323,7 +1309,7 @@ def SplitBregman(Op, RegsL1, data, niter_outer=3, niter_inner=5, RegsL2=None,
                                         x0=x0 if restart else xinv,
                                         **kwargs_lsqr)
             # Shrinkage
-            d = [_shrinkage(RegsL1[ireg] * xinv + b[ireg], epsRL1s[ireg])
+            d = [_softthreshold(RegsL1[ireg] * xinv + b[ireg], epsRL1s[ireg])
                  for ireg in range(nregsL1)]
         # Bregman update
         b = [b[ireg] + tau * (RegsL1[ireg] * xinv - d[ireg]) for ireg in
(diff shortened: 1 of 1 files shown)


mrava87:
Hi Carlos,
thanks for the PR.

Just looking at it and I think there is a reason why I did not use sign, and this is because the input can be a complex number.
Try this:
```
x = np.array([-0.5, 0.5])
x / np.abs(x), np.sign(x)
```
same result. And this:
```
x = np.array([-0.5 + 1j, 0.5 + 1j])
x / np.abs(x), np.sign(x)
```
different results.

I have left in the code some lines commented out from a previous implementation that was based on this link https://stats.stackexchange.com/questions/357339/soft-thresholding-for-the-lasso-with-complex-valued-data#%20%20%20%20#%20for-the-lasso-with-complex-valued-data. Later I realized that I could have a single implementation for the real and complex case 
```
x = np.array([-0.5 + 1j, 0.5 + 1j])
thresh = 0.4
x - thresh * np.sign(x), x - thresh * x / np.abs(x), np.maximum(np.abs(x) - thresh, 0.) * np.exp(1j * np.angle(x))
```

I think this is also why two tests fail, because they are running on a model vector that has complex numbers.

Maybe there is a better way of doing it than ``x/abs(x)`` which I can't think of? Or better to revert to the commented out implementation as it would avoid floating point exception? What do you think? 

PS: I guess the same applies to `_shrinkage` which I start to think it is a copy for `_softthreshold`, could probably be removed. Do you agree?

cako:
Hi Matteo,

Good point on the complex values! I think the best solution is to use the complex-friendly implementation for both types. This is in fact the only version which seems to work in all cases. I tried the different implementations in a simple example:

```
x = np.array([0.1, 1, 1j])
thresh = 0.5

x1 = x - thresh * np.sign(x)
x1[np.abs(x1) <= thresh] = 0

x2 = x - thresh * x / np.abs(x)
x2[np.abs(x2) <= thresh] = 0

x3 = np.maximum(np.abs(x) - thresh, 0.) * np.exp(1j * np.angle(x))

x4 = np.maximum(np.abs(x) - thresh, 0.) * np.sign(x)
for arr in [x1, x2, x3, x4]:
    print("%+.1f%+.1fj    %+.1f%+.1fj    %+.1f%+.1fj" % tuple(np.ravel([[x.real, x.imag] for x in arr])))
```
which prints out
```
+0.0+0.0j    +0.0+0.0j    -0.5+1.0j
+0.0+0.0j    +0.0+0.0j    +0.0+0.0j
+0.0+0.0j    +0.5+0.0j    +0.0+0.5j
+0.0+0.0j    +0.5+0.0j    +0.5+0.0j
```
All implementations except the third one (complex version) actually give the wrong value. The current implementation (second line) does very poorly, and sets everything to zero. If you change any of the coordinates to 0, it also creates a NaN.

As I'm writing this, however, I have noticed that the output type of `np.exp(1j * np.angle(x))` is obligatorily complex, even if `np.angle(x)` is 0 (which makes sense). So, I think in the end the best solution is really what you had commented out in the first place.

In terms of commits, I'd say you close this without accepting the changes, and then one of us can reopen with a clean set of commits. Since this version actually gives different results, it will probably change a few tests as well.

Wr.t. `_shrinkage`, I agree that it is pretty much the same thing.

mrava87:
Agree :) 

Happy if you make a new PR with these two changes: - delete `_shrinkage` and replace its use with `_softthreshold`, - swap back to commented out implementation of  `_softthreshold`.

I quickly checked tests and they should pass no problem as they used to be running with the old implementation :)

PR #172, opened 2020-10-02
Use complex-friendly _softthreshold

cako:
Use complex-friendly implementation for _softthreshold and substitute _shrinkage for _softthreshold

See #170


mrava87:
Great, thanks a lot @cako :)

PR #186, opened 2020-11-17
Add two missing variables in README example

pgdr:
The first PyLops example in the README is not self-contained, which it probably should be.  Trying to run it results in a `NameError`.

I chose to use a `range` instead of `np.arange`, but I can change that if that would be better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,9 @@ and similarly using PyLops commands:
 ```python
 from pylops import FirstDerivative
 
+nx = 7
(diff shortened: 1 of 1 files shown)


mrava87:
Hi @pgdr,
thanks, this looks good. When you say not-self contained, what error you get precisely. I just tried to run it in a new ipython console and it seems to run fine?

mrava87:
Ops, sorry I see what you mean. I was looking at the code snippet above. I agree to also keep this one self consistent, thanks for pointing it out

PR #205, opened 2020-12-12
dottest embeds domain and range sizes

fpicetti:
Instead of passing the domain size and range size, they are taken directly from the operator shape.

In my opinion, one might want to add these extra arguments in order to check if the operator has been initialized correctly.
However, I found the proposed change a bit more comfy.

### DIFF
--- a/pylops/utils/dottest.py
+++ b/pylops/utils/dottest.py
@@ -2,7 +2,7 @@ import numpy as np
 from pylops.utils.backend import get_module
 
 
-def dottest(Op, tol=1e-6, complexflag=0, raiseerror=True, verb=False,
+def dottest(Op, nr=None, nc=None, tol=1e-6, complexflag=0, raiseerror=True, verb=False,
             backend='numpy'):
     r"""Dot test.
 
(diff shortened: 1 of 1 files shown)


mrava87:
Hi @fpicetti,
thanks for the input 👍 

This makes lots of sense. In practice the main reason I had chosen ``nr`` and ``nc`` to be mandatory input arguments is exactly like you say to add an extra layer of check... but we could allow the case where someone just doesn't want to do it and these are inferred from the operator.

However, removing ``nr`` and ``nc`` will be a breaking change which I think it is too small to justify a version bumb to v2.0.0 (as you can see all tests fail as we always pass ``nr`` and ``nc``). My suggestion is, what if we make this change to the function instead:

```
def dottest(Op, nr=None, nc=None, tol=1e-6, complexflag=0, raiseerror=True, verb=False,
            backend='numpy')
      ncp = get_module(backend)
      if nr is None or nc is None:
          nr, nc = Op.shape
```

I didn't see your PR and just merged some code with a small change to the dottest. Do you mind pulling the latest version from upstream and making a new PR with this change?

fpicetti:
Ok, the modification is here but I'll open a new PR.

Thanks!

PR #207, opened 2020-12-14
Dottest nr and nc are default None

fpicetti:
The domain size and range size, `nr` and `nc`, are derived from the operator shape if not provided explicitly.


mrava87:
Great @fpicetti, thanks a lot. I am merging this 👍

PR #222, opened 2021-04-13
Fix broken initialization in CG

BurningKarl:
If `x0=None` is supplied as an argument to `pylops.optimization.solver.cg` (which seems to be intended) `x` is initialized as the zero vector. In this case the residual vector `r` should be `r = y - Op * x = y` and not `r = Op.T * y` (where * is matrix multiplication and .T is transposition).


mrava87:
Hello @BurningKarl, you are completely right! 

Thanks for spotting this. I realized we did not have unittest with ``x0=None``, which is probably why this mistake wasn't caught before ;)

Issue #229, opened 2021-05-07
Dot product test checkimag same as checkreal


ar4:
`checkimag` seems to be the same as `checkreal`, but I suspect that the `np.real`s should be changed to `np.imag`s:
https://github.com/PyLops/pylops/blob/9c7ad688c3333ba68bb2ccbbf3e6f3a6999e5e66/pylops/utils/dottest.py#L104

mrava87:
Hi @ar4, 
good catch!

I quickly checked rerunning all our tests after fixing this locally and only one dottest failed in `test_Bilinear_2dsignal`). I needed to change the dtypes in the dictionaries at the top of the `test_interpolation.py` from float32->float64 and complex64->complex128. Other than that all other tests pass (meaning we never checked the imag component but things were still done correctly, phew! ;) )

Happy to make a PR or shall I do?

ar4:
Hi @mrava87,

That's good news that the test didn't reveal any problems. It sounds like you have all of the work done already, so it's probably fastest for you to submit the fix.

mrava87:
Sounds good, I am doing it and you can have a look if you agree with the change :)

mrava87:
We should think about changing print statements to f-string so we don't need to handle real-complex prints differently. See discussion in https://github.com/PyLops/pylops/pull/231

mrava87:
Added dedicated issue for f-string https://github.com/PyLops/pylops/issues/283

PR #231, opened 2021-05-17
Fix bug in dottest for imag

mrava87:
Resolve https://github.com/PyLops/pylops/issues/229.

### DIFF
--- a/pylops/utils/dottest.py
+++ b/pylops/utils/dottest.py
@@ -101,8 +101,8 @@ def dottest(Op, nr=None, nc=None, tol=1e-6, complexflag=0, raiseerror=True, verb
     else:
         checkreal = np.abs((np.real(yy) - np.real(xx)) /
                            ((np.real(yy) + np.real(xx)+1e-15) / 2)) < tol
-        checkimag = np.abs((np.real(yy) - np.real(xx)) /
-                           ((np.real(yy) + np.real(xx)+1e-15) / 2)) < tol
+        checkimag = np.abs((np.imag(yy) - np.imag(xx)) /
+                           ((np.imag(yy) + np.imag(xx)+1e-15) / 2)) < tol
         if checkreal and checkimag:
             if verb:
                 print('Dot test passed, v^T(Opu)=%f%+fi - u^T(Op^Tv)=%f%+fi'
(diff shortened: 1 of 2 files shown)


ar4:
Looks good.

Is there a reason to use `dot` instead of `vdot`, and to have a separate closeness check, when the values are real? Would the code written for the complex cases work as well (since the imaginary parts will just be zero)?

mrava87:
Great :) No, I suspect this should be the same, probably `dot ` is a bit faster as it doesn't do checks but I guess the overhead is tiny. For the closeness check it is more to have a print statement that doesn't have the images where there is no imag - so it is a bit cleaner. Of course, we could do the check at print stage but I think there is no way out having a check somewhere unless we always want to print real and imag even when we know the latter is 0.

Tests pass with `vdot`, I can make a PR taking that `if... else` away.

ar4:
Does it not print the real and imaginary parts the way that you want if you just print the value (without the `.real` and `.imag`)? If it did, then the same print statement might work for both the real and complex cases.

mrava87:
Yes :)

What I meant was that so far we have this for real numbers:
```
print('Dot test failed, v^T(Opu)=%f - u^T(Op^Tv)=%f' % (yy, xx))
```
and this for complex:
```
print('Dot test passed, v^T(Opu)=%f%+fi - u^T(Op^Tv)=%f%+fi' % (yy.real, yy.imag, xx.real, xx.imag))
```

if we have only one print statement we will always get something like ``v^T(Opu)=3+0j - u^T(Op^Tv)=2+0j`` instead of ``v^T(Opu)=3 - u^T(Op^Tv)=2``. I am not sure if this is worth of extra checks or if we should just let it print always with the imag part even if it is zero. What do you think?

ar4:
Ah, do you wish to keep using that style of Python print statement for backwards compatibility? With the newer ones the print statement will automatically print without the imaginary part if the value is real:
```
a = 1 + 2j
b = 3
print('a={}, b={}'.format(a, b)) # a=(1+2j), b=3
print(f'a={a}, b={b}') # a=(1+2j), b=3
```

mrava87:
Oh right, good point.

It is more an historical reason, I think we should switch to f-strings but since we have quite some prints it is a matter of time and priority. In that case, you are right the all real/complex story will become much easier to handle.

Let's reopen https://github.com/PyLops/pylops/issues/229 as a reminder of this discussion and hopefully we can go for a switch in print statements in coming months :)

mrava87:
Created a new dedicated issue https://github.com/PyLops/pylops/issues/283

PR #233, opened 2021-05-20
Adds R-linear operators

ar4:
Many useful linear operators are only linear over real numbers, and
complex numbers when they are treated as being elements of R^2
(e.g., (a+ib) -> (a, b)). These are known as R-linear operators, to
distinguish them from operators that are also linear over complex
numbers (C-linear operators). Examples include operators that
extract the real or imaginary components of the input, and
complex conjugation.

This commit introduces R-linear operators to PyLops. An operator
is specified to be only R-linear (and thus not C-linear) by
setting the new `clinear` attribute on the LinearOperator class
to be False.

When operators are combined, such as by composition or stacking,
the combined operator will be R-linear if any of the component
operators are. This commit implements this in the `dot` method
of the `LinearOperator` class, and also in the `VStack`, `HStack`,
and `BlockDiag` operators.

As complex values are considered to be elements of R^2 when
using R-linear operators, the dot product test for these operators
uses the R^2 inner product rather than the complex inner product.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,3 +220,4 @@ A list of video tutorials to learn more about PyLops:
 * Tyler Hughes, twhughes
 * Lyubov Skopintseva, lskopintseva
 * Francesco Picetti, fpicetti
+* Alan Richardson, ar4
--- a/docs/source/api/index.rst
+++ b/docs/source/api/index.rst
@@ -57,6 +57,9 @@ Basic operators
     Block
     BlockDiag
     Kronecker
+    Real
+    Imag
+    Conj
 
 Smoothing and derivatives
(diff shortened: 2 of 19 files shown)


mrava87:
Fantastic! I added some small changes here and there - the most important one, casting to outputs of the forward of Real and Imag to a the real-valued equivalent dtype of that of the operator, so those outputs are of float type instead of complex, do you agree?

I'll wait to hear from you and then I think we are ready to merge this PR.

ar4:
It looks good. Thank you for reviewing it. I haven't done the RFFT part yet, but will hopefully submit it someday.

mrava87:
Great, thanks a lot Alan! You can push the RFFT at any other time :)

Issue #268, opened 2021-12-09
RFFT


ar4:
I mentioned in an earlier message that the RFFT operator can be modified to have the correct adjoint. Here is some code:

```python
import numpy as np
from pylops import LinearOperator

class FFTND(LinearOperator):
    def __init__(self, dims, dtype='complex128'):
        self.dims = np.array(dims)
        self.dims_fft = self.dims.copy()
        self.dims_fft[-1] = self.dims_fft[-1] // 2 + 1
        self.shape = (int(np.prod(self.dims_fft)), int(np.prod(self.dims)))
        self.rdtype = np.real(np.ones(1, dtype)).dtype
        self.cdtype = (np.ones(1, dtype=self.rdtype) +
                       1j * np.ones(1, dtype=self.rdtype)).dtype
        self.dtype = self.cdtype
        self.explicit = False


class FFTND_fwd(FFTND):
    def _matvec(self, x):
        x = np.reshape(x, self.dims)
        y = np.fft.rfftn(x, norm='ortho')
        return y.ravel()

    def _rmatvec(self, x):
        x = np.reshape(x.copy(), self.dims_fft)
        x[..., 1:1 + (self.dims[-1] - 1)//2] /= 2 # <-- Scaling
        y = np.fft.irfftn(x, norm='ortho', s=self.dims)
        return y.ravel()


class FFTND_sym(FFTND):
    def _matvec(self, x):
        x = np.reshape(x, self.dims)
        y = np.fft.rfftn(x, norm='ortho')
        y[..., 1:1 + (self.dims[-1] - 1)//2] *= np.sqrt(2) # <-- Scaling
        return y.ravel()

    def _rmatvec(self, x):
        x = np.reshape(x.copy(), self.dims_fft)
        x[..., 1:1 + (self.dims[-1] - 1)//2] /= np.sqrt(2) # <-- Scaling
        y = np.fft.irfftn(x, norm='ortho', s=self.dims)
        return y.ravel()


def dottest(op):
  x = np.random.rand(*op.dims).ravel()
  y = op * x
  yp = np.random.rand(*y.shape) + 1j * np.random.rand(*y.shape)
  xp = op.H * yp

  assert np.allclose(np.real(np.conj(yp) @ y), xp @ x)


for Op in [FFTND_fwd, FFTND_sym]:
  for dims in [(3, 4), (4, 3)]:
    op = Op(dims)
    dottest(op)
```

[...]

PR #273, opened 2021-12-26
In 1D FFT, implements `ifftshift_before` and `fftshift_after`

cako:
This pull request implements `ifftshift_before` as the new version of the `fftshift` in `FFT`, which is now flagged as deprecated.

In addition, it adds the option to apply fftshift after the forward call to the fft library. This option is given by `fftshift_after` and defaults to True.

See #268 for more discussion on these two options.

If the everything is OK with the patch, the same code will be ported to `FFT2` and `FFTN`.


mrava87:
This looks great!

But again, it has conflicts due to the linter changes. Can you either resolve them or make a new PR please :)

cako:
Hi Matteo, great, thanks for the review. I decided to reduce some code duplication by introducing a `_BaseFFT` class which `_FFT_numpy` and `_FFT_pyfftw` now derive from. This should also make it easier to introduce a wrapper for the scipy fft. I also included in the backend utils a couple of functions for obtaining real and complex dtypes from a given input dtype. I feel like we will be using this function a lot in the future, so it's better to have it handy.

With regards to `FFT2/N`, I think every argument that refers to a dimension (dirs, ifftshift_before, etc.) should either be a value or a list of values the size of dirs. The default can then be False for these new options and all tests should pass without a problem. I do want to try and refactor some code so we don't have a lot of duplication accross FFT/2/N but I'd like to familiarize myself with the code a bit more.

When all of this is done, I can then tackle the norm.

mrava87:
Sounds good, I like the idea of a base class.

I merged all the other PRs you created. This one seems to still require some changes to be merged (I guess it’s to do with the black autopep8 conflicts)

PR #276, opened 2021-12-27
Change every flatten to ravel

### LINKED ISSUE #24: Change all np.flatten() to np.ravel()
Not exactly a bug, but it should be avoided np.flatten() unless needed (both codebase and documentation examples).

cako:
Closes #24.


mrava87:
Mmmh looks like this does not play nice with the linter PR. 

Can I ask you to make a new PR now that I merged the other one. When you make it can you avoid changing flatten to ravel in docs/source/roadmap.rst:

```
* Change all ``np.ravel()`` into ``np.ravel()`` 
```

and instead have:

```
* :strike:`Change all ``np.flatten()`` into ``np.ravel()``` -
  `Issue #24 <https://github.com/Statoil/pylops/issues/24>`_.
```

cako:
I think it's good now!

PR #279, opened 2021-12-28
Fix bug in FFT with fftw backend for 1d signals

mrava87:
Modified handling of 1d signals in _FFT_fftw which led to an error
when passing a single number (of int type) for the dims parameter.

### DIFF
--- a/pylops/signalprocessing/FFT.py
+++ b/pylops/signalprocessing/FFT.py
@@ -75,7 +75,9 @@ class _BaseFFT(LinearOperator):
         if self.fftshift_after:
             if self.real:
                 warnings.warn(
-                    "Using fftshift_after with real=True. fftshift should only be applied after a complex FFT. This is rarely intended behavior but if it is, ignore this message."
+                    "Using fftshift_after with real=True. fftshift should only be "
+                    "applied after a complex FFT. This is rarely intended behavior "
+                    "but if it is, ignore this message."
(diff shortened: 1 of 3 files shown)


mrava87:
@cako I found a little bug in _FFT_fftw, when passing a single int to dims, the len(dims) throws an error. I fixed it but just wanting to check you agree (also suggest if you think of a better way to write the if statement

cako (approved):
For now I am going to approve, but I think we should use reconsider using `isinstance` in this manner. Here is an example where this would not work properly:
```python
arr = np.array([1], dtype=np.int64)
isinstance(arr[0], int) # False
print(type(arr[0])) # numpy.int64
```
Until I come up with a better way, I am doing this:
```python
try:
    len(value_or_list_like)
    out = np.array(value_or_list_like)
except TypeError:
    out = np.array([value_or_list_like])
return out
```
It works for any value type (not necessarily int) and let's NumPy cast in its preferred manner. If you then want to ensure that the case was to an int/float/complex-like dtype, you can then use
```python
np.issubdtype(out.dtype, np.integer)
np.issubdtype(out.dtype, np.floating)
np.issubdtype(out.dtype, np.complexfloating)
```
If you want to see how I use these, check out my `fftn` branch in my local version. I will be making it into a pull request soon!

cako (approved):

PR #292, opened 2022-01-07
Feature: Norm option for FFT operator

### LINKED ISSUE #268: RFFT

cako:
New features:
* Adds a `norm` parameter to `FFT`, `FFT2D` and `FFTND` (and base respective base classes). Possible values are "ortho" (default), "backward" and "forward" (it might be a good idea to change these names... opinions?).
* Adds `__truediv__` for all FFT operators, obtained by multiplying by the adjoint with the proper scalar
* Improves documentation
* Closes #268 

Implementational details:
* FFTW backend: FFTW supports an equivalent implementation to "ortho" and "backward" through use of keywords. For norm="forward", we just need to scale the forward FFT with 1/N.
* NumPy and SciPy backends: In order to support old versions of these libraries, we only implement "ortho" through the use of the `norm` option in `fft`. "forward" and "backward" are both implemented manually.

### DIFF
--- a/pylops/signalprocessing/FFT.py
+++ b/pylops/signalprocessing/FFT.py
@@ -110,6 +110,11 @@ class _FFT_numpy(_BaseFFT):
         y = y.astype(self.rdtype)
         return y
 
+    def __truediv__(self, y):
+        if self.norm != "ortho":
+            return self._rmatvec(y) / self._scale / self._scale
+        return self._rmatvec(y)
+
 
 class _FFT_scipy(_BaseFFT):
     """One dimensional Fast-Fourier Transform using numpy"""
@@ -192,6 +197,11 @@ class _FFT_scipy(_BaseFFT):
             y = scipy.fft.fftshift(y, axes=self.dir)
         y = y.ravel()
         return y
+
+    def __truediv__(self, y):
+        if self.norm != "ortho":
+            return self._rmatvec(y) / self._scale / self._scale
+        return self._rmatvec(y)
 
 
 class _FFT_fftw(_BaseFFT):
@@ -339,6 +349,11 @@ class _FFT_fftw(_BaseFFT):
         if not self.clinear:
             y = np.real(y)
         return y.ravel()
+
+    def __truediv__(self, y):
+        if self.norm == "ortho":
+            return self._rmatvec(y)
+        return self._rmatvec(y) / self._scale
 
 
 def FFT(
--- a/pylops/signalprocessing/FFT.py
+++ b/pylops/signalprocessing/FFT.py
@@ -273,21 +273,21 @@ class _FFT_fftw(_BaseFFT):
             warnings.warn(
                 f"FFTW option 'ortho' will be overwritten by norm={self.norm}"
             )
+            del kwargs_fftw["ortho"]
         if "normalise_idft" in kwargs_fftw:
             warnings.warn(
                 f"FFTW option 'normalise_idft' will be overwritten by norm={self.norm}"
             )
+            del kwargs_fftw["normalise_idft"]
+        # Use FFTW without norm-related keywords above. In this case, FFTW standard
+        # behavior is to scale with 1/N on the inverse transform. The _scale below
+        # converts the default bevavior to that of ``norm``.
         if self.norm == "ortho":
-            kwargs_fftw["ortho"] = True
-            kwargs_fftw["normalise_idft"] = False
+            self._scale = np.sqrt(1.0 / self.nfft)
         elif self.norm == "backward":
             self._scale = self.nfft
-            kwargs_fftw["ortho"] = False
-            kwargs_fftw["normalise_idft"] = False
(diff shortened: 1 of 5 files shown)


mrava87:
Hi @cako, great addition :)

I agree that we may come up with a better naming but I wonder if it’s better to stick to numpy/scipy ones so users can easily understand what we do if they are already familiar with the numpy/scipy equivalents. What names do you suggest instead?

cako:
I'm mostly worried about the adjoint scaling, since in order to be a real adjoint, it doesn't respect the same scaling as `ifft` when used with the same keyword. I tried to make this crystal clear in the documentation, but I'm just worried that it will be confusing to users who might be expecting FFTOp.H to have a scaling.

The reason why I left it with "forward" and "backward" is because I honestly couldn't come up with anything better. I was thinking something like "ortho", "noscaling", "oneoverN" which are pretty self explanatory even if a little different. Any ideas?

mrava87:
Oh right, I see what you mean. I am not sure myself about another definition which is more self-explanatory (the one you suggest is for me, but likely not for someone that was not involved in this PR). 

Looking at the documentation, I think what you write is very clear and does not leave any room to misinterpretation (and people need to read the documentation if they want to understand every little detail). If you agree, I would say let's go ahead with this?

cako:
How about "ortho", "none" and "1/n"? In addition, I could implement a warning a nice error on "forward", "backward" explaining that they should use "none" and "1/n".

mrava87:
Sure, that sounds like a good one :)

cako:
Updates on the details of the PR:

New features:

* Adds a norm parameter to FFT, FFT2D and FFTND (and base respective base classes). Possible values are "ortho" (default), "none" and "1/n"
* Adds __truediv__ for all FFT operators, obtained by multiplying by the adjoint with the proper scalar
* Improves documentation
* Closes RFFT #268

Implementational details:
* FFTW backend: while FFTW supports an equivalent implementation to "ortho" and "backward" through use of keywords, this option does not seem to be stable across versions, so we do the scaling manually like in the scipy and numpy backends.
* NumPy and SciPy backends: In order to support old versions of these libraries, we only implement "ortho" through the use of the norm option in fft. "none" and "1/n" are both implemented manually.

PR #296, opened 2022-01-16
Patch docs - math edition

cako:
This PR improves documentation, mostly through bettering the math notation, but also fixing minor bugs and adding improved descriptions. In addition, the second commit removes the "Shared" and "Others" submenus under PyLops Utilities in favor of a single section for all utilities.

I have opened this PR as a draft for two reasons. First, the second commit may be removed if anyone disagrees with it. Second, there were some issues I identified during the preparation of this PR which need to be looked over.

* In [`pylops/avo/avo.py#448` ](https://github.com/PyLops/pylops/blob/bd3a919655ddbf91d9c6c63c456d99d3926b7517/pylops/avo/avo.py#L448), no `G_1` appears in the equation despite appearing in the returns. Please make sure this is correct.
* ISTA and FISTA require `alpha` which according to the documentation must be less than 1/λ(Op^H Op). It is not clear if this means 1 / [λ(Op^H Op)] or (1/λ) Op^H Op.

### DIFF
--- a/docs/source/api/others.rst
+++ b/docs/source/api/others.rst
@@ -7,11 +7,8 @@ Alongside with its *Linear Operators* and *Solvers*, PyLops contains also a numb
 performing universal tasks that are used by several operators or simply within one or more :ref:`tutorials` for
 the preparation of input data and subsequent visualization of results.
 
-Shared
-------
-
 Dot-test
-~~~~~~~~
+------
 
 .. currentmodule:: pylops.utils
 
@@ -21,7 +18,7 @@ Dot-test
     dottest
 
 Estimators
-~~~~~~~~
+------
 
 .. currentmodule:: pylops.utils
 
@@ -33,7 +30,7 @@ Estimators
     trace_nahutchpp
 
 Scalability test
-~~~~~~~~~~~~~~~~
+------
 
 .. currentmodule:: pylops.utils
 
@@ -42,11 +39,8 @@ Scalability test
 
     scalability_test
 
-Others
-------
-
 Synthetics
-~~~~~~~~~~
+------
 
 .. currentmodule:: pylops.utils
 
@@ -70,7 +64,7 @@ Synthetics
 
 
 Signal-processing
-~~~~~~~~~~~~~~~~~
+------
 
 .. currentmodule:: pylops.utils
 
@@ -83,7 +77,7 @@ Signal-processing
 
 
 Tapers
-~~~~~~
+------
 
 .. currentmodule:: pylops.utils
 
@@ -95,7 +89,7 @@ Tapers
 
 
 Wavelets
-~~~~~~~~
+------
 
 .. currentmodule:: pylops.utils
 
@@ -106,8 +100,8 @@ Wavelets
     wavelets.gaussian
 
 
-Geophysicical Reservoir characterization
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Geophysical Reservoir characterization
+------
 
 .. currentmodule:: pylops.avo
 
--- a/pylops/avo/avo.py
+++ b/pylops/avo/avo.py
@@ -299,39 +299,44 @@ def akirichards(theta, vsvp, n=1):
     theta : :obj:`np.ndarray`
         Incident angles in degrees
     vsvp : :obj:`np.ndarray` or :obj:`float`
-        VS/VP ratio
+        :math:`V_S/V_P` ratio
     n : :obj:`int`, optional
-        number of samples (if ``vsvp`` is a scalare)
+        number of samples (if ``vsvp`` is a scalar)
 
     Returns
     -------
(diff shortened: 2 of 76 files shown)


mrava87:
Nice :) let me go through this slowly and make some comments.

For the 2 issues you mentioned:

* In pylops/avo/avo.py#448 , no G_1 appears in the equation despite appearing in the returns. Please make sure this is correct.

This is because all other AVO modelling relations are for PP-waves and have 3 terms (`akirichards` and `fatti`). For PS there is no first term (PS reflection at zero angle is zero). The PS addition was asked by a user after the rest of the code had been already in place for long time. So to be able to call any of these methods interchangeably here https://github.com/PyLops/pylops/blob/014441d1ff271dc071cd31ec589ee6c9e216adda/pylops/avo/prestack.py#L140 I wanted to have the same amount of outputs. So my idea (maybe it is stupid) was to still have a G1 just with zeros. This is definitely not the most efficient way but it avoided restructuring completely the code. Maybe I can add some comments in the code to make it clear, or we can think on how to adapt it so that it checks how many terms are in the linearization when building the modelling operator

* ISTA and FISTA require alpha which according to the documentation must be less than 1/λ(Op^H Op). It is not clear if this means 1 / [λ(Op^H Op)] or (1/λ) Op^H Op.

This is the max eigenvalue of what is inside the parenthesis, also called the Lipschitz constant of the operator. I have seen it written like this in papers but if you think it is unclear maybe we can specify it in words. Definitely not this (1/λ) Op^H Op as to me this is a scaled matrix (whilst we need a scalar), if you think 1 / [λ(Op^H Op)] makes it less ambiguous I am happy with it :)

cako:
Ok all that makes sense. Let me make some changes and get back to you.

pylops/avo/prestack.py
```diff
@@ -234,11 +236,13 @@ def PrestackWaveletModelling(
...
+        :math:`V_S/V_P` ratio
     linearization : :obj:`str`, optional
-        choice of linearization, ``akirich``: Aki-Richards,
-        ``fatti``: Fatti, ``ps``: PS, or any function on the form of
-        ``pylops.avo.avo.akirichards``
+        * "akirich": Aki-Richards
+
+        * "fatti": Fatti
+
+        * "PS": PS, or any function on the form of ``pylops.avo.avo.akirichards``
```
mrava87, line 245: I see that the original description may have been misleading. It should be

* "akirich": Aki-Richards

* "fatti": Fatti

* "PS": PS, 

* any function on the form of ``pylops.avo.avo.akirichards``

pylops/optimization/solver.py
```diff
@@ -273,6 +273,9 @@ def lsqr(
     Solve an overdetermined system of equations given an operator ``Op`` and
     data ``y`` using LSQR iterations.
 
+    .. math::
```
mrava87, line 276: What is this?

docs/source/api/others.rst
```diff
@@ -7,11 +7,8 @@ Alongside with its *Linear Operators* and *Solvers*, PyLops contains also a numb
 performing universal tasks that are used by several operators or simply within one or more :ref:`tutorials` for
 the preparation of input data and subsequent visualization of results.
 
-Shared
-------
-
 Dot-test
-~~~~~~~~
+------
```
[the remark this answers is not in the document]
cako: Old:
image

New:
image

pylops/avo/avo.py
```diff
@@ -299,39 +299,44 @@ def akirichards(theta, vsvp, n=1):
...
         third coefficient of three terms Aki-Richards approximation
-        :math:`[n_{theta}  \times  n_{vsvp}]`
+        :math:`[n_\theta  \times  n_\text{vsvp}]`
 
     Notes
     -----
     The three terms Aki-Richards approximation [1]_ is used to compute the
     reflection coefficient as linear combination of contrasts in
-    :math:`V_P`, :math:`V_S`, and :math:`\rho`. More specifically:
+    :math:`V_P`, :math:`V_S`, and :math:`\rho.` More specifically:
```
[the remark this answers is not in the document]
cako: When the period is outside, the line can break on the period like this:

... and ρ
. More specifically.

Another option is to use the [U+2060 WORD JOINER](https://en.wikipedia.org/wiki/Non-breaking_space#Width_variation) but I thought this might be overkill hahahaha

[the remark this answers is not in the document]
mrava87: Ah right :)

pylops/avo/prestack.py
cako: Oh, I see! Changed in the newest commit.

pylops/optimization/solver.py
cako: It declares `\cond\` as a math operator (like `\max`, `\tan`, etc.). Since "cond" is used quite a bit in the documentation, I just declared it there. It does not change the rendering at that location (the LaTeX command is hidden).

PR #297, opened 2022-01-17
feature: Vectorize slope_estimate

cako:
This PR vectorizes the loops in `slope_estimate`. The seislet example improves by a factor of ~90x on my machine:

```python
%%timeit -n 100
slope = -pylops.utils.signalprocessing.slope_estimate(d.T, dt, dx, smooth=6)[0]
# New: 8.29 ms ± 144 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
# Old: 754 ms ± 24.3 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```


mrava87:
Oh good catch, I didn't realize I had implemented this with for loops!

Issue #295, opened 2022-01-16
Add warning to FFT2D


cebirnie92:
When initialising FFT2D it would be nice to have a warning or error thrown if using an nt that is bigger than nfft. 

For example, the following code should throw a warning: 

```
    nt = d.shape[0]  # 2**12
    nfft_k = nfft_f = 2**10
    F = FFT2D(dims=(nt, nr), nffts=(nfft_f, nfft_k), sampling=[dt, dx], dtype=np.float)

**Comment by mrava87** [comment]
Thanks!

@cako : Claire was using FFT2D and accidentally chose nfft<nt and got this error:

```
y = np.take(y, range(self.dims[self.dirs[1]]), axis=self.dirs[1])
...
IndexError: index 128 is out of bounds for axis 1 with size 128
```

I kind of do this all the time and soon realize the mistake but I can see that for a normal user the error thrown right now may be not obvious. I wanted to check what you think:

- 1. shall we immediately raise an error during initizialization if any of the nfft is smaller than the corresponding dimension in the original domain?
- 2. can you think of any case where someone may want an fft shorter than the original domain? By doing so you lose info as under the hood classic FFT implementations truncate the input and then apply FFT, so I would be tempted to say no... if this is the case then we should not bother accommodating for this and revert to raising an error, but if this is the case it may be better to change the way we truncate (checking if nfft is bigger and do it only when this is the case (and possibly pad in the other scenario).

I can easily fix all FFTs in either way, but just wanted a second opinion before I move on :)

cako:
I personally cannot see a reason for anyone to use a smaller nfft than nt. I'm also not too worried about accommodating this case as it can easily be achieved by cropping the input vector as you mentioned. At the same time I'm not a fan of the idea of simply banning the user from doing so: I think we need to keep in line with the philosophy that our users know what they are doing.

Maybe a good middle ground is to allow nfft < nt but warn that input signals will be cropped in accordance to NumPy/SciPy behavior.

mrava87:
Sounds like a good deal!

So, we basically keep all as it is but we add a warning to tell users that what they are doing (nfft<nt) will risk losing some of the signal as FFT truncates upfront... and we can simply check nfft vs the dims and when this is smaller replace np.take with np.pad... that should work :)

I'll give it a go and see

PR #299, opened 2022-01-18
Bug: fix issue with nfft<nt

mrava87:
This `PR` handles https://github.com/PyLops/pylops/issues/295 for all FFT operators.

Two changes are made:

1. A warning is raised when user selects `nfft<nt`;
2. When the above condition arises, the output of the adjoint is padded after ifft such as it has size `nt`. This way a user can select `nfft<nt` without getting an error when the adjoint is invoked

### DIFF
--- a/pylops/signalprocessing/FFT.py
+++ b/pylops/signalprocessing/FFT.py
@@ -98,8 +98,12 @@ class _FFT_numpy(_BaseFFT):
             y = np.fft.ifft(x, n=self.nfft, axis=self.dir, **self._norm_kwargs)
         if self.norm is _FFTNorms.NONE:
             y *= self._scale
-        if self.nfft != self.dims[self.dir]:
+
+        if self.nfft > self.dims[self.dir]:
             y = np.take(y, np.arange(0, self.dims[self.dir]), axis=self.dir)
+        elif self.nfft < self.dims[self.dir]:
+            y = np.pad(y, self.ifftpad)
+
         if not self.clinear:
             y = np.real(y)
         if self.ifftshift_before:
@@ -185,8 +189,12 @@ class _FFT_scipy(_BaseFFT):
             y = scipy.fft.ifft(x, n=self.nfft, axis=self.dir, **self._norm_kwargs)
         if self.norm is _FFTNorms.NONE:
             y *= self._scale
-        if self.nfft != self.dims[self.dir]:
+
+        if self.nfft > self.dims[self.dir]:
             y = np.take(y, np.arange(0, self.dims[self.dir]), axis=self.dir)
+        elif self.nfft < self.dims[self.dir]:
+            y = np.pad(y, self.ifftpad)
+
         if not self.clinear:
             y = np.real(y)
         if self.ifftshift_before:
@@ -298,9 +306,11 @@ class _FFT_fftw(_BaseFFT):
             x = np.real(x)
         if self.dopad:
             x = np.pad(x, self.pad, "constant", constant_values=0)
+        elif self.doifftpad:
+            x = np.take(x, np.arange(0, self.nfft), axis=self.dir)
 
         # self.fftplan() always uses byte-alligned self.x as input array and
-        # returns self.y as output array. As such, self.y must be copied so as
+        # returns self.y as output array. As such, self.x must be copied so as
         # not to be overwritten on a subsequent call to _matvec.
         np.copyto(self.x, x)
         y = self.fftplan().copy()
@@ -341,8 +351,11 @@ class _FFT_fftw(_BaseFFT):
         elif self.norm is _FFTNorms.NONE:
             y *= self._scale
 
-        if self.nfft != self.dims[self.dir]:
+        if self.nfft > self.dims[self.dir]:
             y = np.take(y, np.arange(0, self.dims[self.dir]), axis=self.dir)
+        elif self.nfft < self.dims[self.dir]:
+            y = np.pad(y, self.ifftpad)
+
         if self.ifftshift_before:
             y = np.fft.fftshift(y, axes=self.dir)
         if not self.clinear:
--- a/pylops/signalprocessing/FFT.py
+++ b/pylops/signalprocessing/FFT.py
@@ -100,7 +100,7 @@ class _FFT_numpy(_BaseFFT):
             y *= self._scale
 
         if self.nfft > self.dims[self.dir]:
-            y = np.take(y, np.arange(0, self.dims[self.dir]), axis=self.dir)
+            y = np.take(y, range(0, self.dims[self.dir]), axis=self.dir)
         elif self.nfft < self.dims[self.dir]:
             y = np.pad(y, self.ifftpad)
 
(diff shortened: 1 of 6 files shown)


mrava87:
@cako, can you take a look at this.

It was quite easy to handle, as we discussed I both added a warning as most likely users will choose nfft<nt by mistake more than for a real reason. But, if someone has a real reason, now the adjoint works by padding the output of ifft to nt.

pylops/signalprocessing/_BaseFFTs.py
```diff
@@ -102,6 +102,22 @@ def __init__(
             nfft = nffts[0]
         self.nfft = nfft
 
+        # Check if the user provided nnft smaller than n (size of signal in
```
cako, line 105: Could you fix the typo in `nfft`?

mrava87: Done

cako:
Looks good, I was just working on a patch myself and came across very similar solutions. I'm still reviewing the files but so far it looks good. One thing: my tests flagged this line as problematic: https://github.com/PyLops/pylops/blob/2ac336e9a9b5b40456aecbfcc7e976dbcfdcb1b5/pylops/signalprocessing/FFT.py#L315

when using engine="fftw" and real=True. It might be a good idea to add that test in your suite (it should fail). My fix to this issue was:
```python
np.copyto(self.x, np.take(x, range(self.nfft), axis=self.dir))
```

mrava87:
> Looks good, I was just working on a patch myself and came across very similar solutions. I'm still reviewing the files but so far it looks good. One thing: my tests flagged this line as problematic:
> 
> https://github.com/PyLops/pylops/blob/2ac336e9a9b5b40456aecbfcc7e976dbcfdcb1b5/pylops/signalprocessing/FFT.py#L315
> 
> when using engine="fftw" and real=True. It might be a good idea to add that test in your suite (it should fail). My fix to this issue was:
> 
> ```python
> np.copyto(self.x, np.take(x, range(self.nfft), axis=self.dir))
> ```

Sure. On which test, the `test_FFT_1dsignal`? If I do 

```
par7 = {
    "nt": 41,
    "nx": 31,
    "ny": 10,
    "nfft": 16,
    "real": True,
    "engine": "fftw",
    "ifftshift_before": False,
    "dtype": np.complex128,
}  # nfft<nt, real input, fftw engine
```

it seems to be still fine with the old np.copy. What am I missing?

cako:
Probably nothing hahahaha maybe your implementation is more robust that mine. Just to make sure we are hitting all the right combinations, can you try this test:

```python
par_lists_small_nfft = dict(
    shape=[
        np.random.randint(4, 9, size=(1,)),
        # np.random.randint(4, 9, size=(2,)),
        # np.random.randint(3, 6, size=(3,)),
    ],
    real=[True, False],
    engine=["numpy", "scipy", "fftw"],
)
# Generate all combinations of the above parameters
pars_small_nfft = [
    dict(zip(par_lists_small_nfft.keys(), value))
    for value in itertools.product(*par_lists_small_nfft.values())
]


@pytest.mark.parametrize("par", pars_small_nfft)
def test_small_nfft(par):
    """Dot-test when nfft < nt"""
    shape = par["shape"]
    real = par["real"]
    engine = par["engine"]
    x = np.random.randn(*shape)
    if not real:
        x = x.astype(np.complex128) + 1j * np.random.randn(*shape)

    nffts = [(n + 1) // 2 for n in shape]
    if len(shape) == 1:
        FFTOp = FFT(x.shape, nfft=nffts[0], real=real, engine=engine)
    elif len(shape) == 2:
        if engine == "fftw":
            return
        FFTOp = FFT2D(x.shape, nffts=nffts, real=real, engine=engine)
    else:
        if engine == "fftw":
            return
        FFTOp = FFTND(x.shape, nffts=nffts, real=real, engine=engine)

    y = FFTOp * x.ravel()
    y = y.reshape(FFTOp.dims_fft)
    print("New")
    print("x:     ", x.shape)
    print("nffts: ", nffts)
    print("y:     ", y.shape)
    xadj = FFTOp.H * y.ravel()
    xadj = xadj.reshape(x.shape)
    print("xadj:  ", xadj.shape)
    nr, nc = FFTOp.shape
    if par["real"]:
        assert dottest(FFTOp, nr, nc, complexflag=2)
    else:
        assert dottest(FFTOp, nr, nc, complexflag=2)
        assert dottest(FFTOp, nr, nc, complexflag=3)

```

mrava87:
Mmmh I just copy-pasted this at the bottom of the `test_ffts.py` file and all tests pass :)

pylops/signalprocessing/FFT.py
```diff
@@ -185,8 +189,12 @@ def _rmatvec(self, x):
             y = scipy.fft.ifft(x, n=self.nfft, axis=self.dir, **self._norm_kwargs)
         if self.norm is _FFTNorms.NONE:
             y *= self._scale
-        if self.nfft != self.dims[self.dir]:
+
+        if self.nfft > self.dims[self.dir]:
             y = np.take(y, np.arange(0, self.dims[self.dir]), axis=self.dir)
```
cako, line 194: I think this is a good opportunity to change all of these lines to the simpler:
```python
np.take(y, range(self.dims[self.dir]), axis=self.dir)
```

pylops/signalprocessing/FFT2D.py
```diff
@@ -93,8 +93,14 @@ def _rmatvec(self, x):
...
             y *= self._scale
-        y = np.take(y, range(self.dims[self.dirs[0]]), axis=self.dirs[0])
-        y = np.take(y, range(self.dims[self.dirs[1]]), axis=self.dirs[1])
+        if self.nffts[0] > self.dims[self.dirs[0]]:
+            y = np.take(y, range(self.dims[self.dirs[0]]), axis=self.dirs[0])
+        if self.nffts[1] > self.dims[self.dirs[1]]:
+            y = np.take(y, range(self.dims[self.dirs[1]]), axis=self.dirs[1])
+        if (self.nffts[0] < self.dims[self.dirs[0]]) or (
+            self.nffts[1] < self.dims[self.dirs[1]]
+        ):
```
cako, line 102: Why note use `self.doifftpad` here?

pylops/signalprocessing/FFTND.py
```diff
@@ -84,8 +84,16 @@ def _rmatvec(self, x):
...
-            y = np.take(y, range(self.dims[direction]), axis=direction)
+        for direction, nfft in zip(self.dirs, self.nffts):
+            if nfft > self.dims[direction]:
+                y = np.take(y, range(self.dims[direction]), axis=direction)
+        if any(
+            [
+                nfft < self.dims[direction]
+                for direction, nfft in zip(self.dirs, self.nffts)
+            ]
+        ):
```
cako, line 95: Why note use `self.doifftpad` here?

pylops/signalprocessing/FFTND.py
```diff
@@ -169,8 +177,16 @@ def _rmatvec(self, x):
...
```
cako, line 188: And here

pylops/signalprocessing/FFT.py
mrava87: Agreed!

pylops/signalprocessing/FFT2D.py
mrava87: Good one :)

pylops/signalprocessing/FFTND.py
mrava87: Same

mrava87:
Added all! Let me know if you agree and I merge it :)

cako (approved):
I'm approving but there are some changes not related to the bug. Since they are comments I don't think it's a big issue.

Issue #306, opened 2022-01-23
Alter development flow


cako:
## Motivation
The current development flow is to develop entirely on the `master` branch, with stable versions being tagged as releases. On ReadTheDocs, the latest documentation always refers to the latest `master` commit. As breaking changes are introduced when developing the next version, the documentation can rapidly change, break and generally become confusing for users. In addition, this causes an issue when introducing bugfixes: since the development version will have many other changes that are not just bugfixes, it is hard to release minor versions that simple patch existing versions. Currently one would have to branch from the release version, import the commit as a cherry pick, and then tag that branch as the new release. Merge conflicts are usually harder to solve in these cases.

## Proposal
I propose a few changes to the process. First, we move the development to a `dev` branch, which can potentially even be broken as we develop the library. Second, we maintain master as the latest stable, which should essentially be equivalent to the last released version. This will allow us to quickly introduce bugfixes to the latest stable without having to worry about all the other changes in `dev`.  In addition, new users cloning the master branch won't have to deal with rapidly changing code. Finally, to make the documentation compatible with this change, we should change the [latest docs](https://pylops.readthedocs.io/en/latest/) to reflect the `dev` branch, [stable](https://pylops.readthedocs.io/en/stable/) to `master` and set stable as the default landing page for documentation.

## Migration
Currently, `master` introduces several breaking changes compared to the latest release (v1.16.0), but is still very far from v2.0.0. With that said, the breaking changes are not enormous and mostly related to bugfixes/some new features. I think we can release v1.17.0 with the current master (maybe with some cleanup on documentation to ensure that changes are documented), tag it as v1.17.0 and move development to `dev`. From then on, we can focus on delivering v2.0.0.

mrava87:
Alright, let me highlight a few points and we can come up with a final plan after that:

1. Readthedocs: fair point, I agree it is a good idea for people to land on a 'stable' documentation, which we likely would like to the latest release (see second screenshot, Stable points to that by default). I am attaching a screenshot of Readthedocs settings, this could be very easily fixed by changing the first dropdown to Stable instead of Latest. This would not need any Github related changes. If you agree I can make this change instantly?

<img width="815" alt="Screen Shot 2022-01-25 at 12 39 48 PM" src="https://user-images.githubusercontent.com/22675848/150951473-81847079-0e5a-45f0-9d0e-c003876984c7.png">
[...]

PR #307, opened 2022-01-23
Patch docs - v1.17 edition

cako:
This PR contains several improvements to documentation, including math formatting and some language fixes. Most importantly, however, it documents with versionadded/note/warning tags all the changes introduced since the latest stable release v1.16.0.

Following the rationale in #306, I have used version v1.17 as the following major, but this can be amended if #306 is not accepted.

### DIFF
--- a/docs/source/api/others.rst
+++ b/docs/source/api/others.rst
@@ -8,7 +8,7 @@ performing universal tasks that are used by several operators or simply within o
 the preparation of input data and subsequent visualization of results.
 
 Dot-test
-------
+--------
 
 .. currentmodule:: pylops.utils
 
@@ -18,7 +18,7 @@ Dot-test
     dottest
 
 Describe
-------
+--------
 
 .. currentmodule:: pylops.utils.describe
 
@@ -28,7 +28,7 @@ Describe
     describe
 
 Estimators
-------
+----------
 
 .. currentmodule:: pylops.utils
 
@@ -40,7 +40,7 @@ Estimators
     trace_nahutchpp
 
 Scalability test
-------
+----------------
 
 .. currentmodule:: pylops.utils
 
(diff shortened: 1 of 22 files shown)


mrava87:
Hi @cako,
sounds good, I think we have quite some new things to make a new release.

A few comment:
- to follow the current standard of versioning (https://semver.org) could you please change all v1.17 into v1.17.0 which is also how I will later tag the release.
- any time we make a new release we have to update https://github.com/PyLops/pylops/blob/master/CHANGELOG.md and https://github.com/PyLops/pylops/blob/master/docs/source/changelog.rst. No worries, I can merge this PR once ready and do that as last thing before releasing.
- I'll reply separately to the release flow proposal :)

Btw, for historical reasons I currently release pylops manually but from next time we starting using Github actions as done here (https://github.com/DIG-Kaust/pymarchenko/blob/main/.github/workflows/deploy.yaml). The flow would be locally release to https://test.pypi.org to check it is all working fine, then make the release on Github which will automatically trigger the release to PyPI. I can share privately some scripts I use to check the first step is working (once a release is made on PyPI you can't overwrite so its good to make sure everything works otherwise we would need to immediately release a patch :)

cako:
That sounds good, I will make the version numbering change. Regarding the test failing, this is actually something I've noticed is happening quite a lot. `test_PhaseShift_*` tests often fail due to machine precision. I've change a few here or there (including one in this patch, but it's not the one which is failing), but I think more need to have the "decimal" option in the dot-test. However, I think that should be in another PR.

It would be great to have the GitHub actions for release, go for it and I will help in any way I can.

mrava87:
Alright, let me merge the PhaseShift PR first. Yes, most tests run no matter what but when there is some lower precision (f32) or complex involved sometimes the precision we get may vary. With time I started to realize you need random seeds inside a single test not at top of the file but I think here and there it is still the old way and failed tests may pop up when we add new ones as they change the pseudorandom behavior. Let’s see if you can fix this one and hopefully it becomes less and less common for this to happen :)

Issue #123, opened 2019-12-05
Simplify redundant `N` and `dims`


prisae:
Some calls take an `N` and a `dims` parameter, e.g., `pylops.FirstDerivative`. The `N` is redundant, as it can be derived from `dims`.

This is there for legacy reasons and should be cleaned-up for v2.0.0 (@mrava87 you can tag that for the milestone/project board).

Also, as it happens in the same instance call, replacing the `dir` with `axis` (like used in NumPy/SciPy).

mrava87:
@cako this is done right, ready to be closed?

prisae:
Oh wow, Dec 2019, I definitely do not remember having opened this issue :-)

cako:
Yes, this is closed by #310 and #332!

PR #310, opened 2022-01-26
Convert `dir` to `axis`

cako:
## Motivation

As described in #123, the choice of `dir` for direction in many linear operators is not compatible with NumPy/Scipy notation. In addition, it encroaches on the [`dir`](https://docs.python.org/3/library/functions.html#dir) built-in function. As such, it is highly desirable to move from the use of `dir` to `axis` instead. In addition, some other related keywords such as `dirs` and `nodir` can both be substituted by `axes`. While there are other improvements that can be made in this respect, the scope of this PR is simply to move away from `dir` without altering the current functionality of these modified operators.

## Strategy

The main issue to keep in mind is to retain the old behavior as much as possible, so as to allow users to migrate to ``axis`` and not break any existing code. The approach taken to ensure this was as follows

1. Change function signatures from `..., dir=0,...` → `..., axis=-1, dir=None, ...` and similar for `dirs`/`axes`.
2. Ensure that when `dir` is supplied, it supersedes the use of `axis` but also issues a deprecation warning:
```python
if dir is not None:
    warnings.warn(
        "dir is deprecated in version 2.0.0, use axis instead.", category=DeprecationWarning, stacklevel=2,
    )
    axis = dir
else:
    axis = axis
```
3. The documentation is updated to always use `axis` and note that `dir` is deprecated like so:
```rst
axis : :obj:`int`, optional
    .. versionadded:: 2.0.0
    Axis along which derivative is applied.
dir : :obj:`int`, optional
    .. deprecated:: 2.0.0
        Use ``axis`` instead. Note that the default for ``axis`` is -1
        instead of 0 which was the default for ``dir``.
```
4. Changes to the linear operators were first made prior to changing tests: this ensured during development that using `dir` always worked. After this check, tests were updated to also call `axis`.

## Considerations
I am creating this as a draft because there I'd like some feedback on this strategy. For example, I strongly favor setting the default as `axis=-1` for v2.0.0, but this *is* a breaking change. I am also open to amending warnings/documentation on this change.


mrava87:
Let me get back to this once we got v1.18.0 out :)

mrava87:
I have not forgotten about this, just being thinking a little about it.

Overall thought:

- My main fear is that we are polluting the input variables with an another parameter and this may create more confusing than benefit in the short run. Postponing this to `v2.0.0` and describe this as one of the major code-breaking changes may be a better option?

Here a some more specific thoughts:

- In general I agree the choice of `dir` was a bad one I regretted early. But I decided that a worst choice would have been to switch half way through... probably doing what you have done early on would have been the best choice ;)

- Similarly I agree that in v2.0.0 we should move to default `axis=-1`. I can think of a few cases where having `dir=0` as default makes some sense for how the operator is operating, but in most cases working on the last axis is the best idea.

- One thing I am not sure I understand `Change function signatures from ..., dir=0,... → ..., axis=-1, dir=None, ... and similar for dirs/axes.`. I understand that a code currently working that has an operator where `dir=0` will keep working. But what about a currently written code where an operator is made but **dir is not specified**, would the behavior suddenly change to `axis=-1`? Unfortunately if this is the case I think we need to introduce `axis` more softly... one way could be to have `axis=None` and if this is provided it will take priority... but maybe I don't understand the strategy you designed?

- In warning messages I think it is better to say `dir will be deprecated` as we don't have v2.0.0 yet.

cako:
Hi @mrava87 thanks for the comments. Let's go over your them:

* I absolutely agree that this should go in v2.0.0. The question is, are you expecting to release other versions before v2.0.0? If so, we could move this to a feature branch tagged v2.0.0.
* Yup, maintaining compatibility is important!
* Also agree here, there may be some operators where `axis=0` makes more sense as a default. I know NumPy has a few of those. If you think that any of the ones I changed should be changed, flag them and let's discuss!
* The signature change is kind of tricky. When setting a value as a default e.g. `dir=0` in the function signature, there is no way to distinguish between calling the function with `myfunction()` or with `myfunction(dir=0)`. In order to distinguish between the two cases, you have two options: 
  ```python
  def myfunction(dir=None): # Option 1
  def myfunction2(**kwargs): # Option 2
  ``` 
   Calling `myfunction()` will create `dir=None` in the function body and calling `myfunction(dir=0)` will create `dir=0` in the function body. This way we can distinguish in non-`None` values are passed to the function. When calling `myfunction2()`, `dir` will not exist in `kwargs` and calling `myfunction2(dir=any_value)` will create `kwargs[dir] = any_value`.
   So if we want to ensure that `dir` is respected whenever it is supplied, we need to choose one of two options. Since we don't use `dir=None` as a valid keyword, the first option is preferred. Using `kwargs` is generally the last resort since it makes code really hard to read, maintain and debug (for example, calling `myfunction2(arg_that_doesnt_exist=0)` does not raise an error. That's the reasoning behind the `dir=None`.
   To answer your question, yes, not specifying `dir` will use `axis=-1`. But that's what a default of `axis=-1` is. One possibility I considered was to first introduce `axis=0` (the same as `dir`) and only later switch the default to `axis=-1`. But in this case, the user will have to know exactly which version introduces the change of defaults. On the other hand, doing it all at once ensures the user that if `axis` is a valid keyword, then `axis=-1` will always be the default. In terms of priority, `dir` will always precede `axis` until it is retired as a keyword.
* Good point, I will change it!

cako:
This PR, together with #332 closes #123.

mrava87:
Great! Let me take a look in details in coming days (I just want to be sure if there is any operator where we don't want to switch convention 0->1) then I can merge it :)

MIGRATION_V1_V2.md
```diff
@@ -8,6 +8,7 @@ should be used as a checklist when converting a piece of code using PyLops from
 - Several operators have deprecated `N` as a keyword. To migrate, pass only `dims` if both `N` and `dims` are currently
   being passed. If only `N` is being passed, ensure it is being passed as a value and not a keyword argument (e.g.,
   change `Flip(N=100)` to `Flip(100)`).
+- `dir`, `dirs` and `nodir` have been deprecated in favor of `axis` and `axes`.
```
mrava87, line 11: I would add. 
```
When previously `nodir` was required, you must now provide the directions along which the operator is applied in `axes`.
```

And more importantly, we should remark:

```
The default value for `axis` is chosen to be -1 (before the default value for `dir` was 0). Be careful to check all operators where `dir` was not provided as the direction along which they are applied is now -1.
```

pylops/basicoperators/CausalIntegration.py
```diff
@@ -7,14 +9,21 @@
...
     ----------
     dims : :obj:`list` or :obj:`int`
         Number of samples for each dimension
+    axis : :obj:`int`, optional
+        .. versionadded:: 2.0.0
+
+        Axis along which the model is integrated.
     dir : :obj:`int`, optional
-        Direction along which array is integrated.
+
```
mrava87, line 23: General comment for all operators. As we are doing with PR targeting v2.0.0 (this will never make it into any v1 versions, shall we really keep `dir` as an input parameter and allow people to still use it?

pylops/basicoperators/FirstDerivative.py
```diff
@@ -95,50 +114,50 @@ def __init__(
     def _matvec_forward(self, x):
         ncp = get_array_module(x)
         x = ncp.reshape(x, self.dims)
-        if self.dir > 0:  # need to bring the dim. to derive to first dim.
-            x = ncp.swapaxes(x, self.dir, 0)
+        if self.axis > 0:  # need to bring the dim. to derive to first dim.
```
mrava87, line 117: This could have been done before as it is an internal change, but since we are already working with moving all to axis=-1, I think it makes more sense to have this code pattern:

```
        if self.axis not in (-1, len(self.dims)-1):
            x = ncp.swapaxes(x, self.axis, -1)
        y = ncp.zeros(x.shape, self.dtype)
        y[..., :-1] = (x[..., 1:] - x[..., :-1]) / self.sampling
        if self.axis not in (-1, len(self.dims)-1):
            y = ncp.swapaxes(y, -1, self.axis)
```
as every operation should be faster (or at least not slower) in the last dimension. If you agree, the same needs changed in quite a few places :)

I can implement it but I just want to first check that you agree :)

pylops/basicoperators/Restriction.py
```diff
@@ -151,14 +169,14 @@ def mask(self, x):
 
         y = np_ma.array(np.zeros(self.dims), mask=np.ones(self.dims), dtype=self.dtype)
         x = np.reshape(x, self.dims)
-        if self.dir > 0:
```
mrava87, line 154: Same as the `FirstDerivative` comment

pylops/basicoperators/SecondDerivative.py
```diff
@@ -89,52 +115,52 @@ def __init__(
     def _matvec_forward(self, x):
         ncp = get_array_module(x)
         x = ncp.reshape(x, self.dims)
-        if self.dir > 0:  # need to bring the dim. to derive to first dim.
-            x = ncp.swapaxes(x, self.dir, 0)
+        if self.axis > 0:  # need to bring the dim. to derive to first dim.
```
mrava87, line 118: Same as the `FirstDerivative` comment

pylops/basicoperators/Symmetrize.py
```diff
@@ -69,24 +88,24 @@ def _matvec(self, x):
...
+            y = ncp.swapaxes(y, 0, self.axis)
         y = y.ravel()
         return y
 
     def _rmatvec(self, x):
         ncp = get_array_module(x)
         x = ncp.reshape(x, self.dimsd)
-        if self.dir > 0:  # bring the dimension to symmetrize to first
-            x = ncp.swapaxes(x, self.dir, 0)
+        if self.axis > 0:  # bring the dimension to symmetrize to first
```
mrava87, line 104: Same comment as `FirstDerivative`

pylops/signalprocessing/Interp.py
```diff
@@ -16,23 +17,23 @@ def _checkunique(iava):
...
-def _nearestinterp(dims, iava, dir=0, dtype="float64"):
+def _nearestinterp(dims, iava, axis=0, dtype="float64"):
     """Nearest neighbour interpolation."""
     iava = np.round(iava).astype(int)
     _checkunique(iava)
-    return Restriction(dims, iava, dir=dir, dtype=dtype), iava
+    return Restriction(dims, iava, axis=axis, dtype=dtype), iava
 
 
-def _linearinterp(dims, iava, dir=0, dtype="float64"):
```
mrava87, line 26: Not a big thing but we can put also here axis=-1 (and in the _sincinterp :)

mrava87 (commented):
First half, done up to Interp

mrava87 (commented):
Second part

pylops/basicoperators/CausalIntegration.py
cako: That is a good point. I say we just get rid of `dir`, `dirs`  and `nodir` entirely then. I will start working on it this week.

pylops/basicoperators/FirstDerivative.py
cako: Actually, we can get rid of the `if` everywhere since `swapaxes` is extremely fast (less than a microsecond on my machine).  In case we both agree to just get rid of it, I can work on it but for another PR :)

pylops/signalprocessing/Sliding1D.py
```diff
@@ -103,7 +103,7 @@ def Sliding1D(Op, dim, dimd, nwin, nover, tapertype="hanning", design=False):
 
     combining = HStack(
         [
-            Restriction(dimd, np.arange(win_in, win_end), dtype=Op.dtype).H
+            Restriction(dimd, np.arange(win_in, win_end), axis=0, dtype=Op.dtype).H
```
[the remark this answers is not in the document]
cako: You're right, this one is not required!

pylops/signalprocessing/Interp.py
cako: Yes! Done.

pylops/signalprocessing/Sliding2D.py
```diff
@@ -135,7 +135,7 @@ def Sliding2D(Op, dims, dimsd, nwin, nover, tapertype="hanning", design=False):
 
     combining = HStack(
         [
-            Restriction(dimsd, range(win_in, win_end), dtype=Op.dtype).H
+            Restriction(dimsd, range(win_in, win_end), axis=0, dtype=Op.dtype).H
```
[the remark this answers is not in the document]
cako: My main worry was not breaking any tests and `axis=-1` breaks a few. With regard to the design, I'm happy to keep it as is, since sliding along the slowest does make sense to me.

pylops/signalprocessing/_BaseFFTs.py
```diff
@@ -24,7 +24,7 @@ class _BaseFFT(LinearOperator):
     def __init__(
         self,
         dims,
-        dir=0,
+        axis=0,
```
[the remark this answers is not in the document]
cako: Fixed.

pylops/basicoperators/FirstDerivative.py
mrava87: Agree, if it is such a minor overhead let's not bother checking and just do it :) another PR sounds good

cako:
@mrava87, could you have a final look? By the way, we should probably squash all this into one commit when we merge :)

mrava87:
Let me take a final look and I let you know if I have anything more :) Then I am happy if you want to squash all in one commit as there is quite a lot here and it all fits within the same idea of removing `dir` :)

mrava87:
Alright, apart from the small comment on the MIGRATION file, everything else looks good to me and ready to go :)

cako:
> Alright, apart from the small comment on the MIGRATION file, everything else looks good to me and ready to go :)

Have a look [here](https://github.com/PyLops/pylops/pull/310/commits/fad324ebca650230995e40cbcbd5428810db7f9a). If it's ok, feel free to merge!

mrava87:
Wait, maybe my comment was confusing. In the MIGRATION file you say:

```
the default `dir` and `dirs` was -1 and (0, 1),
```

shouldn't that be:

```
the default `dir` and `dirs` was 0 and (0, 1),
```

Apart from that, you want to squash all in one commit before I merge or not? I don't really have strong feeling here ;)

PR #321, opened 2022-02-05
Minor: Extend Laplacian to Nd-arrays

mrava87:
This PR extends the `Laplacian` operator to perform derivatives over N directions of a N-dimensional array.

Previously the operator could only apply derivatives over up to two directions of a N-dimensional array.


mrava87:
@cako I just bumped into this and realized we could make it more general, do you agree?

cako (approved):

PR #326, opened 2022-02-12
Bugfix: Modifications to handle scipy 1.8.0

mrava87:
Some of the private routines in scipy.sparse.linalg.interface have
been moved to scipy.sparse.linalg._interface. Moreover, scipy lsqr
atol and btol have changed in this version to more conservative values,
the old values are forced in PyLops to ensure tests to pass without
loosening up on tolerance.

### DIFF
--- a/pylops/LinearOperator.py
+++ b/pylops/LinearOperator.py
@@ -11,8 +11,13 @@ from scipy.sparse.linalg import eigs as sp_eigs
 from scipy.sparse.linalg import eigsh as sp_eigsh
 from scipy.sparse.linalg import lobpcg as sp_lobpcg
 from scipy.sparse.linalg import lsqr, spsolve
-from scipy.sparse.linalg.interface import _ProductLinearOperator
 
+# need to check scipy version since the interface submodule changed into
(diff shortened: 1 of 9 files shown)


pylops/LinearOperator.py
```diff
@@ -11,8 +11,13 @@
 from scipy.sparse.linalg import eigsh as sp_eigsh
 from scipy.sparse.linalg import lobpcg as sp_lobpcg
 from scipy.sparse.linalg import lsqr, spsolve
-from scipy.sparse.linalg.interface import _ProductLinearOperator
 
+# need to check scipy version since the interface submodule changed into
+# _interface from scipy>=1.8.0
+if int(sp.__version__.split(".")[1]) < 8:
```
cako, line 17: Need to check major as well (`int(sp.__version__.split(".")[0]) <= 1`

mrava87: Oh yeah good point, even though we have a harsh requirement on scipy being >=1.4.. but I guess it’s good just in case scipy goes up to 2?

PR #332, opened 2022-02-14
Patch `N`/`dims`

cako:
As noted in #123, the use of `N` and `dims` is redundant. This patch introduces a unified format for using `dims` in operators:
```python
# Old
def __init__(self, N, dims=None, dir=0):

# New (actual code differs)
def __init__(self, dims, dir=0):
    self.dims = [dims] if is_value(dims) else array(dims)
```

With this standard, every operator which receives `N` and `dims` may now receive only `dims`, with `N` being derived from `dims` internally if required (i.e. `N = np.prod(self.dims)`). The list of operators changed follows:

* Flip
* Roll
* Symmetrize
* CausalIntegration
* FirstDerivative
* SecondDerivative
* Restriction
* Convolve1D
* Convolve2D
* ConvolveND
* Interp

Notes:
* Internally many other operators are affected, e.g., Gradient, Laplacian, Patch2D to name a few
* The only two remaining operators which take N are Identity and Zero. They do not take dims, so using the above pattern would need a rewrite which I am not sure is a good idea.


mrava87:
This is finally good to go, I think.

 Just one thing to add to the PR: I have created a file called MIGRATION_V1_V2.md where we should document to our best all the changes a user would need to do to their code to v2 ready, a sort of check list… this is where we should document this changes pretty much copying your description in this PR

cako:
Done!

PR #335, opened 2022-02-19
Refactor dottest

prisae:
Just refactoring, removing a lot of duplicate code. No functional change.

The only change is that the message always says `^H`, also for real values. I feel that is OK and not wrong, but if preferred I can add that it prints `^T` for real values.

Warning: I have _not_ tested it! I assume the tests that run when I create the PR will do that...

(I am using f-strings, f"{var}", which were introduced in Python 3.6; should be fine I think, or does PyLops still support Python 3.5?)

### DIFF
--- a/pylops/utils/dottest.py
+++ b/pylops/utils/dottest.py
@@ -112,7 +112,7 @@ def dottest(
 
     def passes(xx, yy, tol=tol):
         """True if xx and yy are  the same within tolerance."""
-        return np.abs((yy - xx) / ((yy + xx + 1e-15) / 2)) < tol
+        return abs((yy - xx) / ((yy + xx + 1e-15) / 2)) < tol
 
     # evaluate if dot test is passed (both real and imag parts)
     passed = passes(np.real(xx), np.real(yy)) and passes(np.imag(xx), np.imag(yy))
--- a/pylops/utils/dottest.py
+++ b/pylops/utils/dottest.py
@@ -110,50 +110,21 @@ def dottest(
     # complex numbers in subsequent prints also when using cupy arrays.
     xx, yy = np.array([to_numpy(xx)])[0], np.array([to_numpy(yy)])[0]
 
+    def passes(xx, yy, tol=tol):
+        """True if xx and yy are  the same within tolerance."""
+        return np.abs((yy - xx) / ((yy + xx + 1e-15) / 2)) < tol
+
     # evaluate if dot test is passed
     if complexflag == 0:
-        if np.abs((yy - xx) / ((yy + xx + 1e-15) / 2)) < tol:
-            if verb:
-                print("Dot test passed, v^T(Opu)=%f - u^T(Op^Tv)=%f" % (yy, xx))
-            return True
-        else:
-            if raiseerror:
-                raise ValueError(
-                    "Dot test failed, v^T(Opu)=%f - u^T(Op^Tv)=%f" % (yy, xx)
-                )
-            if verb:
-                print("Dot test failed, v^T(Opu)=%f - u^T(Op^Tv)=%f" % (yy, xx))
-            return False
+        passed = passes(xx, yy)
     else:
         # Check both real and imag parts
-        checkreal = (
-            np.abs(
-                (np.real(yy) - np.real(xx)) / ((np.real(yy) + np.real(xx) + 1e-15) / 2)
-            )
-            < tol
-        )
-        checkimag = (
-            np.abs(
-                (np.imag(yy) - np.imag(xx)) / ((np.imag(yy) + np.imag(xx) + 1e-15) / 2)
-            )
-            < tol
-        )
-        if checkreal and checkimag:
-            if verb:
-                print(
-                    "Dot test passed, v^T(Opu)=%f%+fi - u^T(Op^Tv)=%f%+fi"
-                    % (yy.real, yy.imag, xx.real, xx.imag)
-                )
-            return True
-        else:
-            if raiseerror:
-                raise ValueError(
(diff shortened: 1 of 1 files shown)


prisae:
This
```
    if complexflag == 0:
        passed = passes(xx, yy)
    else:
        # Check both real and imag parts
        passed = passes(np.real(xx), np.real(yy)) and passes(np.imag(xx), np.imag(yy))
```
could actually be simplified to
```
        passed = passes(np.real(xx), np.real(yy)) and passes(np.imag(xx), np.imag(yy))
```
as `passes(np.imag(xx), np.imag(yy))` will simply be `True` for real numbers, so it can be used for real numbers too.

But I first want to see if you are interested in these changes at all or not.

mrava87:
Thanks! 

This looks good to me, I agree the `dottest` had quite some repetitions. I like the idea of making a `passes` function :)

I think we can merge this in, do you agree @cako ?

pylops/utils/dottest.py
```diff
@@ -110,50 +110,17 @@ def dottest(
...
+        """True if xx and yy are  the same within tolerance."""
+        return np.abs((yy - xx) / ((yy + xx + 1e-15) / 2)) < tol
+
+    # evaluate if dot test is passed (both real and imag parts)
+    passed = passes(np.real(xx), np.real(yy)) and passes(np.imag(xx), np.imag(yy))
+
+    if not passed and raiseerror:
+        raise ValueError(f"Dot test failed, v^H(Opu)={yy} - u^H(Op^Hv)={xx}")
+    elif verb:
+        passed_str = ['failed', 'passed'][passed]
```
cako, line 123: Could you change this to:
```python
passed_status = "passed" if passed else "failed" 
```

cako:
Hi guys, looks really good! I just added a minor suggestion, but in any case I happy with it.

prisae:
Btw: `xx` and `yy` are always floats or complex values, no arrays, right? Then I would change `np.abs((yy - xx) ...` to `abs((yy - xx) ...`.

prisae:
By the way, instead of checking for real and for imaginary separate, I would do the following instead:
```
2 * abs(yy - xx) / (abs(yy) + abs(xx) + 1e-15) < tol
```
I don't think you need to threat them separately. This is how I always do it in CSEM.

cako:
I think that is also a valid way of checking but for compatibility I'd stick to the old way. Regarding the removal of `np.abs` in favor or `abs`, I don't see a problem with it!

prisae:
I see you are in favour @cako - great! I have one more thing. The `+ 1e-15` can screw it up if you work with numbers which are very low, as is often the case in EM. If you compare numbers in the level of 1e-21 it will almost always pass the current test.

I would use this instead:
```
if abs(xx) + abs(yy) != 0.0:
    passed = 2 * abs(xx - yy) / (abs(xx) + abs(yy)) < tol
else:  # Only if xx == yy == 0.0
    passed = True
```
This works for any number of any magnitude, and for floats as well as for complex numbers, without having a `+1e-15`.

What do you think?

prisae:
Ah, I saw your comment too late. Just let me know what you think. Otherwise I'll only do the `np.abs`->`abs`.

cako:
> I see you are in favour @cako - great! I have one more thing. The `+ 1e-15` can screw it up if you work with numbers which are very low, as is often the case in EM. If you compare numbers in the level of 1e-21 it will almost always pass the current test.
> 
> I would use this instead:
> 
> ```
> if abs(xx) + abs(yy) != 0.0:
>     passed = 2 * abs(xx - yy) / (abs(xx) + abs(yy)) < tol
> else:  # Only if xx == yy == 0.0
>     passed = True
> ```
> 
> This works for any number of any magnitude, and for floats as well as for complex numbers, without having a `+1e-15`.
> 
> What do you think?

Doing equality check for (non-negative) float is essentially equivalent to doing `abs(xx) + abs(yy) > eps` where `eps` is the machine epsilon for the type. The issue with this is that you are implicitly relying on this epsilon. I know you didn't introduce it, but division also has the issue that very small numbers will blow up the denominator, and even if the numerator is equivalently as small, you lose precision on the division. So even removing it would still be troublesome for very small numbers. The solution to the division would be to change the code to something like
```python
2 * abs(xx - yy) < tol * (abs(xx) + abs(yy))
```
 
The issue with this solution is that, for very small numbers for `xx` and `yy`, when you multiply them by another very small number (`tol`), you require an absurd precision not available for floats or even doubles.

The canonical solution to all these problems is to use `rtol` and `atol`: `rtol` is the relative tolerance (as used in the code above) and absolute tolerance could look something like this: 
```python
abs(xx - yy) < atol 
```
I think in hindsight this is probably the best solution, but now need two tolerances.

So we could use `tol` an alias for `rtol` and implement `atol` separately. Or we make `tol` be `rtol` if, for example, `(abs(xx) + abs(yy)) > 1e2 * eps` or be `atol` in the other case. In any case, it will a little fiddling to ensure all tests still pass, so maybe we can leave this for another PR?

prisae:
You are of course right, and it is absolutely fine. I just thought I'd raise it, but as you mention, it would be a bigger thing and be better handled in a separate issue/PR.

So, from my point of view this PR is ready and if you are happy with it you can go ahead and merge.

prisae:
(The best and easiest would probably be to simply use `np.isclose(rtol, atol)`...

prisae:
We could just set `tol` to `rtol`, and `atol=0`, but add it to the options. Anyway, it is OK if that is in another PR.

cako (approved):

PR #336, opened 2022-02-19
Check passed with np.isclose

prisae:
As a follow-up to #335.

We can have the discussion here.

The defaults I set (`rtol=1e-6, atol=1e-21`) corresponds more or less to the status quo I think. Backwards compatibility is given through the kwargs option. So both, keyword and positional arguments will keep working the same way:
```
dottest(Op, nr, nc, tol)
dottest(Op, nr, nc, tol=tol)
```

### DIFF
--- a/MIGRATION_V1_V2.md
+++ b/MIGRATION_V1_V2.md
@@ -8,3 +8,5 @@ should be used as a checklist when converting a piece of code using PyLops from 
 - XX
 - XX
 - XX
+
+- `utils.dottest`: The relative tolerance is new set via `rtol` (before `tol`), and absolute tolerance is new supported via the keyword `atol`. When calling it with purely positional arguments, note that after `rtol` comes now first `atol` before `complexflag`. When using `raiseerror=True` it now emits an `AttributeError` instead of a `ValueError`.
--- a/examples/plot_nmo.py
+++ b/examples/plot_nmo.py
@@ -296,7 +296,7 @@ class NMO(LinearOperator):
 # and adjoint transforms truly are adjoints of each other.
 
 NMOOp = NMO(t, x, vel_t)
-dottest(NMOOp, *NMOOp.shape, tol=1e-4)
+dottest(NMOOp, *NMOOp.shape, rtol=1e-4)
 
 ###############################################################################
 # NMO using :py:class:`pylops.Spread`
@@ -370,7 +370,7 @@ SpreadNMO = Spread(
     dtable=nmo_dtable,  # Table of weights for linear interpolation
     engine="numba",  # numba or numpy
 ).H  # To perform NMO *correction*, we need the adjoint
-dottest(SpreadNMO, *SpreadNMO.shape, tol=1e-4)
+dottest(SpreadNMO, *SpreadNMO.shape, rtol=1e-4)
 
 ###############################################################################
 # We see it passes the dot test, but are the results right? Let's find out.
--- a/pylops/utils/dottest.py
+++ b/pylops/utils/dottest.py
@@ -90,22 +90,15 @@ def dottest(
         raise AssertionError("Provided nr and nc do not match operator shape")
 
     # make u and v vectors
-    if complexflag != 0:
-        rdtype = np.real(np.ones(1, Op.dtype)).dtype
+    rdtype = np.ones(1, Op.dtype).real.dtype
 
-    if complexflag in (0, 2):
-        u = ncp.random.randn(nc).astype(Op.dtype)
-    else:
-        u = ncp.random.randn(nc).astype(rdtype) + 1j * ncp.random.randn(nc).astype(
-            rdtype
-        )
+    u = ncp.random.randn(nc).astype(rdtype)
+    if complexflag not in (0, 2):
+        u = u + 1j * ncp.random.randn(nc).astype(rdtype)
 
-    if complexflag in (0, 1):
-        v = ncp.random.randn(nr).astype(Op.dtype)
-    else:
-        v = ncp.random.randn(nr).astype(rdtype) + 1j * ncp.random.randn(nr).astype(
-            rdtype
-        )
+    v = ncp.random.randn(nr).astype(rdtype)
+    if complexflag not in (0, 1):
+        v = v + 1j * ncp.random.randn(nr).astype(rdtype)
 
     y = Op.matvec(u)  # Op * u
     x = Op.rmatvec(v)  # Op'* v
--- a/pylops/utils/dottest.py
+++ b/pylops/utils/dottest.py
@@ -7,11 +7,13 @@ def dottest(
     Op,
     nr=None,
     nc=None,
-    tol=1e-6,
+    rtol=1e-6,
+    atol=1e-21,
     complexflag=0,
     raiseerror=True,
     verb=False,
     backend="numpy",
+    **kwargs,
 ):
     r"""Dot test.
 
@@ -28,8 +30,10 @@ def dottest(
         Number of rows of operator (i.e., elements in data)
     nc : :obj:`int`
         Number of columns of operator (i.e., elements in model)
-    tol : :obj:`float`, optional
-        Dottest tolerance
+    rtol : :obj:`float`, optional
+        Relative dottest tolerance
+    atol : :obj:`float`, optional
+        Absolute dottest tolerance
     complexflag : :obj:`bool`, optional
         Generate random vectors with
 
@@ -67,6 +71,9 @@ def dottest(
         \mathbf{u}^H(\mathbf{Op}^H\mathbf{v})
 
     """
+    # Backwards compatibility for ``tol`` keyword:
+    rtol = kwargs.pop("tol", rtol)
+
     ncp = get_module(backend)
 
     if nr is None:
@@ -110,12 +117,8 @@ def dottest(
     # complex numbers in subsequent prints also when using cupy arrays.
     xx, yy = np.array([to_numpy(xx)])[0], np.array([to_numpy(yy)])[0]
 
-    def passes(xx, yy, tol=tol):
-        """True if xx and yy are  the same within tolerance."""
-        return abs((yy - xx) / ((yy + xx + 1e-15) / 2)) < tol
-
-    # evaluate if dot test is passed (both real and imag parts)
-    passed = passes(np.real(xx), np.real(yy)) and passes(np.imag(xx), np.imag(yy))
+    # evaluate if dot test passed
+    passed = np.isclose(xx, yy, rtol, atol)
 
     if not passed and raiseerror:
         raise ValueError(f"Dot test failed, v^H(Opu)={yy} - u^H(Op^Hv)={xx}")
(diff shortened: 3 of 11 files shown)


prisae:
I wonder if, with this implementation, the following lines are still required:
```
    # convert back to numpy (in case cupy arrays were used), make into a numpy
    # array and extract the first element. This is ugly but allows to handle
    # complex numbers in subsequent prints also when using cupy arrays.
    xx, yy = np.array([to_numpy(xx)])[0], np.array([to_numpy(yy)])[0]
```

I have the impression we could remove them.

mrava87:
This looks nice @prisae. Just a few comments:
- which status quo do you refer to? As the default for np.allclose seems to be different? Also I think it is not easy to choose universal values as working with single vs double precision would affect what you expect. Said that, seeing all tests pass makes me happy and I agree to have such strict tolerances :)
- For the cupy thingy, I can easily test it on my machine with a gpu, I suspect you are right but let’s not remove this until we are 100% sure
- I am happy to make breaking changes when this really gives something to the user, in this case I think we can adopt the strategy you suggest (adding kwargs) as making a breaking change to have r in front of tol sounds a bit too aggressive. @cako, what do you think?

prisae:
With status quo I meant what PyLops used. Because you had the 1e-15, with `tol=1e-6`, that meant that any two values below 1e-21 would pass. For this reason I set the `atol=1e-21`, as I think it is more or less the ballpark of the "status quo".

- `np.isclose/np.allclose` has `rtol=1e-05, atol=1e-08`
- `np.testing.assert_allclose` has `rtol=1e-07, atol=0`

I first thought to go with the `np.testing`-scenario with `atol=0`, which is super restrict, but then I thought it might be better to use something that resembles more or less of what you had before.

No hurry here, check with your GPU whenever you have time!

cako:
I am happy with the values, don't see a problem with them. I do think we should avoid using `kwargs` though. One of the biggest issues with using `kwargs` is that when you call a function with the "wrong" arguments, it will not complain. Example
```python
def fun_without_kwargs(arg1, arg2=None):
    pass
def fun_with_kwargs(arg1, **kwargs):
    pass

fun_without_kwargs(1, arg3=10) # raises error
fun_with_kwargs(1, arg3=10) # rails silently
```

Instead, I would suggest keeping `tol` and deprecating it. The implementation would be similar to how it is done here: #310 
Basically something like this:
```python
def dottest(..., tol=None, ...):
"""
[...]
    tol : :obj:`float`, optional
        Dottest tolerance
        .. deprecated:: 2.0.0
            Use ``axis`` instead.
[...]
"""
[...]
  if tol is not None:
      warnings.warn(
          "tol will be deprecated in version 2.0.0, use rtol or atol instead.",
          category=DeprecationWarning,
          stacklevel=2,
      )
      rtol = tol
```

prisae:
Two comments:
- First, I would actually replace the `ValueError` by an `AssertionError`, which is more descriptive, what do you think?
- To your point @cako : if you are concerned about wrong kwargs then we can add simply this
```
    rtol = kwargs.pop("tol", rtol)
    if kwargs:
        raise TypeError(f"Unexpected **kwargs: {list(kwargs.keys())}.")
```
(I actually had it and removed it before making the PR. In a function it makes sense to check that kwargs is empty. In a class you often do not want it, in able to allow inheritance and just pass kwargs up to the next in line.)

prisae:
Having said that, adding the deprecation is anyway a good idea.

I have no hard feelings with regards to using kwargs or not - I can put whatever you two decide, just let me know.

cako:
* That is a very good point. Why don't we create a separate `assert_dottest` (which raises `AssertionError`) to be used in testing? This is how NumPy does it afaik... but maybe it is overkill for us. @mrava87 thoughts? Either way, I think yes, it should raise AssertionError. But note that this would be breaking behavior, which means it will have to go in v2.
* If it's the same to you, I would rather keep the kwargs out of it :)

prisae:
Ha, I didn't think of it, changing the Error type would indeed break backwards compatibility. So no problem, I don't think it is a big deal whether it is `ValueError` or `AssertionError`, more semantics than other.

I made three small commits, I can undo any if you don't like any of them (please squash it when you merge, these commits don't have to be separated):

1. I replaced the `kwarg` by an `arg` for `tol`. I did not document it, as I personally would still not document deprecated things, but I can add the documentation if you prefer.
2. I replaced the
    ```
    assert (nr, nc) == Op.shape, "Provided nr and nc do not match operator shape"
    ```
    by
    ```
        if (nr, nc) != Op.shape:
            raise AssertionError("Provided nr and nc do not match operator shape")
    ```
    as `assert` is removed by some optimization flags such as `-O` and `-OO` (https://archive.is/5GfiG), and should only be used in tests, not in code.
3. I simplified the creation of the random vectors. 

Let me know what you think.


It _might_ be worth to think if it would be a good point, as we are reworking the entire thing, to switch to the "new" version of random number creation in NumPy (https://numpy.org/doc/stable/reference/random/generator.html) - however, I do not know how that would interply with CuPy, but I assume they take that over too.

prisae:
I just realized that introducing `atol` after `rtol` will be backwards incompatible for folks using only positional arguments, so I move it to the end.

E.g., 
```
dottest(Op, nr, nc, tol, complexflag)
```
might give unexpected results, as `complexflag` is taken as `atol` now!

Moving `atol` to after `verb` resolves that issue.

mrava87:
Few more minor comments before this is good to go:
- Happy with the tolerance values
- I agree with @cako here, as there is no consensus all over Python anyways on anything, I would rather keep `tol` for the rtol and introduce `atol` as you do. So we don't break existing codes for such a minor modification in a parameter name. I would also suggest add
- Leave the cupy thingy in, I will test the new `dottest` with and without using cupy codes and if nothing pops up I can make a new quick PR removing it.

prisae:
This just as information, thought it might be helpful for your current re-work: If you plan to re-model the entire code-base, and will have many many deprecations/removals etc, it might be worth do implement a machinery for it. In `discretize` @jcapriot did a fantastic job with it, when renaming most things to more pep8-like stuff and re-organize the code base. He introduced the functions `deprecate_class`, `deprecate_function`, `deprecate_property`, `deprecate_module`, `deprecate_method`, see e.g., https://github.com/simpeg/discretize/blob/main/discretize/utils/code_utils.py. These make it then very simple to deprecate things.

mrava87:
> This just as information, thought it might be helpful for your current re-work: If you plan to re-model the entire code-base, and will have many many deprecations/removals etc, it might be worth do implement a machinery for it. In `discretize` @jcapriot did a fantastic job with it, when renaming most things to more pep8-like stuff and re-organize the code base. He introduced the functions `deprecate_class`, `deprecate_function`, `deprecate_property`, `deprecate_module`, `deprecate_method`, see e.g., https://github.com/simpeg/discretize/blob/main/discretize/utils/code_utils.py. These make it then very simple to deprecate things.

Nice, this looks like a good idea. So I guess they keep the old methods/properties but call those routines when they are invoked? Forever or for a grace period?

prisae:
Some really convenient (too short for pep*, e.g. `nC` instead of `n_cells`) potentially forever, but most just for a certain period.

Seems that I broke the tests, I'll look into it.

prisae:
But then I do not really understand the reasoning of you two:

> We are moving towards v2 with the idea of doing breaking changes once forever (hopefully) to fix all the bad things we have from early days design. One thing is to rearrange parameters in a logical way (especially for those methods where we added new params at the end to avoid breaking compatibility. So in this case I would be in favour of logical order even if it breaks, otherwise we will never fix that. @cako what do you think?

Because if this is the case, so we can change the order as we do not care for backwards compatibility, then we should also simply remove `tol` and not care about that either...? Or do I miss something?

mrava87:
Correct, I didn't think about the fact you have added `atol` and changed the order... but since this is the case let's also change `tol` into `rtol` and have a new clean method. Then you need to change all tests :)

My idea is that we should document all breaking changes in https://github.com/PyLops/pylops/blob/dev/MIGRATION_V1_V2.md so it will be easy for users to migrate. We could double up and also add what you suggest, keep tol at the end and give a warning/error if used, let's see what @cako think?

prisae:
I forgot to import warnings, hence the failed tests.

OK, I move the `atol` back, added `versionadded` and `deprecated` version numbers. Let's see what @cako thinks.

cako:
So, I think we were talking about two different situations. One is branch 1.x, which should not break anything, and merely add features and fix bugs. My comments on removing `tol` and not changing the raised error were with this in mind. Another situation is for version v2, where we can "do whatever we want" in a sense.

For branch 1.x, we cannot change the order of arguments. So the changes introduced in the PR have to go in v2, which I believe everyone is fine with, and indeed is how these changes are documented.

So if everyone is in agreement, the final touches I would make is follow to @prisae's suggestion and change ValueError to AttributeError and document all changes in [MIGRATION_V1_V2.md](https://github.com/PyLops/pylops/blob/dev/MIGRATION_V1_V2.md). Something like this:

* In `dottest`, use `rtol` instead of `tol`. When using `raiseerror=True` it now emits an `AttributeError` instead of a `ValueError`. In addition, `dottest` now also supports `atol`.

mrava87:
@cako, this looks perfect to me wrt v2.

Personally I think the game of keeping v1 alive is a good one, but getting us to progress both is a dangerous game as eventually we will get fewer people to switch. 

My idea for v1 was to really only make changes if we find real bugs. This one I would not call it a bug as it has worked until now without any problem ...someone can argue, as @prisae did, that the 1e-15 water level could be dangerous... is this a big enough change to also modify the v1 branch, and in that case yes we need to follow what you suggest, keep `tol` and add `atol` into a new PR targeting v1.8.x. Shall we?

cako:
Yup, that seems fine with me!

prisae:
Just heading for lunch/dinner now. I will separate the two things later, now many things make more sense!

prisae:
So I will change this one for v2. - The change for v1, shall I open the PR against `master` or against `dev`? I assume then, following your comments, that only the tolerance aspect goes into v1, nothing else (hence not #335 either), correct?

mrava87:
Yes, let’s finish this one for dev. 

No refractoring is ok also for v1 as you are going to make a PR anyways and that one doesn’t really introduce breaking changes. And you can add atol as last optional input :) that PR should go against v1.18.x which we are going to use to make bug fixes on v1 :)

prisae:
Another question: would you prefer the:
- `atol=1e-21`
- `atol=0` (like np.testing.assert_allclose)?
for v2?

prisae:
Have a look. For me it is good to go, if the tests pass. I kept `atol=1e-21`. It is probably a good sensible default for 95% of the problems, and others can adjust it...

cako (approved):
Looks great!

prisae:
Oh - of course. Now I have to rewrite all the tests to  use `rtol` - will do

cako:
Oops, I think some of the tests need to be rewritten. Not sure what is happening!

prisae:
I am on it @cako - just changing `tol` to `rtol`, as it is now backwards incompatible. Should be ready in a few minutes.

prisae:
I think I caught all... :crossed_fingers:

cako (approved):

PR #337, opened 2022-02-19
Revamped __init__ files

mrava87:
This PR revamps the __init__ files with 2 major changes:

- Docstrings are added such that `help(module)` will print some description
- `__all__` lists are used to define what methods in a module will be imported when `from <module> import *`

### DIFF
--- a/pylops/basicoperators/__init__.py
+++ b/pylops/basicoperators/__init__.py
@@ -1,37 +1,111 @@
-# isort: skip_file
-from .FunctionOperator import FunctionOperator
-from .MemoizeOperator import MemoizeOperator
-from .Regression import Regression
-from .LinearRegression import LinearRegression
-from .MatrixMult import MatrixMult
-from .Diagonal import Diagonal
-from .Zero import Zero
-from .Identity import Identity
-from .Restriction import Restriction
-from .Flip import Flip
-from .Symmetrize import Symmetrize
-from .Spread import Spread
-from .Transpose import Transpose
-from .Roll import Roll
-from .Pad import Pad
-from .Sum import Sum
-from .Real import Real
-from .Imag import Imag
-from .Conj import Conj
+"""
+Basic Linear Operators
+======================
 
-from .VStack import VStack
-from .HStack import HStack
-from .Block import Block
-from .BlockDiag import BlockDiag
-from .Kronecker import Kronecker
+The subpackage basicoperators extends some of the basic linear algebra
+operations provided by numpy providing forward and adjoint functionalities.
 
-from .Smoothing1D import Smoothing1D
-from .Smoothing2D import Smoothing2D
+A list of operators present in pylops.basicoperators :
 
-from .CausalIntegration import CausalIntegration
-from .FirstDerivative import FirstDerivative
-from .SecondDerivative import SecondDerivative
-from .Laplacian import Laplacian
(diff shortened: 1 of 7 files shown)


mrava87:
@cako, have a look and let me know what you think?

PS: I still want to do some checks on the documentation to make sure this new inits don’t affect the way examples and tutorial appear below each operator.

pylops/signalprocessing/__init__.py
```diff
@@ -1,22 +1,76 @@
...
+    Bilinear                    	Bilinear interpolation operator.
+    FFT                             One dimensional Fast-Fourier Transform.
+    FFT2D                       	Two dimensional Fast-Fourier Transform.
+    FFTND                           N-dimensional Fast-Fourier Transform.
+    Shift                           Fractional Shift operator.
+    DWT                             One dimensional Wavelet operator.
+    DWT2D                           Two dimensional Wavelet operator.
+    Seislet                         Two dimensional Seislet operator.
+    Radon2D	                        Two dimensional Radon transform.
+    Radon3D	                        Three dimensional Radon transform.
```
cako, line 23: Whitespace issue

pylops/basicoperators/__init__.py
```diff
@@ -1,37 +1,116 @@
...
+operations provided by numpy providing forward and adjoint functionalities.
+
+A list of operators present in pylops.basicoperators :
+
+    MatrixMult	                    Matrix multiplication.
+    Identity                        Identity operator.
+    Zero         	                Zero operator.
+    Diagonal     	                Diagonal operator.
+    Transpose   	                Transpose operator.
+    Flip        	                Flip along an axis.
```
cako, line 15: Some more whitespace issues

pylops/signalprocessing/__init__.py
cako, line 23: some more whitespace

cako:
Looks good to me, just some minor details. First, there are some issues with tab characters. I flagged some, but many don't appear in the GitHub interface, only on my local code editor. For example, there is a tab right next to "AVOLinearModelling". My recommendation would be to automatically substitute them on vscode or your editor of choice.

I think it is also a good opportunity to remove the isort comments in favor of a global option in .pre-commit-config.yaml:
```python
  - repo: https://github.com/pycqa/isort
    rev: 5.10.1
    hooks:
      - id: isort
        name: isort (python)
        args:
          [
            "--profile",
            "black",
            "--skip",
            "__init__.py",
            "--filter-files",
            "--line-length=88",
          ]
```

Other than that, everything looks good. I get the help on the terminal and it looks great!

mrava87:
Thanks! 

I checked all trailing spaces, now it should be fixed. And modified the `.pre-commit-config.yaml` as you suggested, better than acting on each file. Locally my documentation seems the same as the one on `latest` so once we agree this is ready I will give it a shot on readthedocs and hopefully there is no problem there either :)

cako (approved):

cako:
Sounds good. I'll let you merge after the check on readthedocs?

mrava87:
I can't really run readthedocs without merging... this is one more reason why having latest on dev is a good idea. Shall I merge? And if something arise I will need to figure out what to change a make a new PR...

cako:
Done!

PR #340, opened 2022-02-20
#335 and #336; but backwards compatible

prisae:
#335 and #336; but backwards compatible - just as discussed in #336

### DIFF
--- a/pylops/utils/dottest.py
+++ b/pylops/utils/dottest.py
@@ -1,3 +1,4 @@
+import warnings
 import numpy as np
 
 from pylops.utils.backend import get_module, to_numpy
@@ -7,11 +8,13 @@ def dottest(
     Op,
     nr=None,
     nc=None,
-    tol=1e-6,
+    rtol=1e-6,
     complexflag=0,
     raiseerror=True,
     verb=False,
     backend="numpy",
+    atol=1e-21,
+    tol=None,  # deprecated, use rtol
 ):
     r"""Dot test.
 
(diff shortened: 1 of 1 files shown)


mrava87:
Mmh, I worry this isn't 100% backward compatible, or at least not for someone choosing to use positional arguments. Can we keep `tol` for `rtol` and just add `atol` as last input parameter :)

prisae:
This is backwards compatible for both positional and keyword arguments @mrava87 - otherwise many tests would fail!

mrava87:
Oh I see now, sorry my bad!

Can you please just change the versionadded thing I commented on and then it’s good to go :)

prisae:
What do you mean with "the versionadded thing"?

I did add `versionadded` for `atol`, and `deprecated` for `tol`; I did not do anything for `rtol`, because I thought it is not a change in functionality (nor position), just the name changed. Do you want me to add a `versionadded` by `rtol`?
```
    atol : :obj:`float`, optional
        Absolute dottest tolerance
        .. versionadded:: 2.0.0
    tol : :obj:`float`, optional
        Dottest tolerance
        .. deprecated:: 2.0.0
            Use ``rtol`` instead.
```

prisae:
Oh, I realize that I wrote `versionadded:: 2.0.0`, but it should be `1.18.????`?

mrava87:
Yep, this is what I meant as this will go into a v1.8.1 :)

mrava87:
Alright, I let the tests finish and merge.

Thanks a lot 🚀

PR #346, opened 2022-03-07
Patch swapaxes

cako:
This PR reworks the use of `swapaxes`. First, it always applies `swapaxes` even when not required, since the overhead cost is almost nil. Second, it always swaps to axis=-1.

This PR should be applied after the PR #310.


mrava87:
No comments here, I merge :)

Thanks!

PR #347, opened 2022-03-08
Implement `dimsd` for all operators with `dims`

cako:
## Motivation
The are many operators whose output is of a different shape as that of input. In these cases, it can be confusing to the user that does not know exactly how this output is shaped.

## Changes
* This PR introduces a standardized `dimsd` for all operators which take `dims` except `MatrixMul`, `Patch2D`, `Sliding2D` and `Interp`
* `dims_fft` in FFT operators and `dims_d` in `Sum` are replaced by `dimsd`
* `dims` and `dimsd` are now tuples always

### DIFF
--- a/pylops/signalprocessing/DWT.py
+++ b/pylops/signalprocessing/DWT.py
@@ -6,6 +6,7 @@ import numpy as np
 
 from pylops import LinearOperator
 from pylops.basicoperators import Pad
+from pylops.utils._internal import _value_or_list_like_to_tuple
 
 try:
     import pywt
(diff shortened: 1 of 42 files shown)


mrava87:
This is very useful. I am almost done reviewing this, just a question:

the few operators that you did not do are because they are mostly def based operators, right? And what about Laplacian, I guess that’s the same?

cako:
Honestly some of them I just didn't know how to do it right. These were `MatrixMul`, `Patch2D`, `Sliding2D`. `Laplacian` I forgot, so I will patch it up.

Note that `Laplacian` is a summation of many `LinearOperators`, all of them with the same `dims/dimsd`. So, ideally `dims`/`dimsd` could propagate through additions/multiplications. For example let's say we have two operators, `A` with `dims = (3, 2, 101)` and `dimsd = (3, 2, 51)`, and `B` with `dims = (3, 2, 51)` and `dimsd = (3, 2, 1)`. then `B @ A` would have `dims = (3, 2, 101)` and `dimsd = (3, 2, 1)`. Of course, if `C` has no `dims`/`dimsd`, then `C @ A` would have `dims` but no `dimsd`. Unfortunately it is still not clear for me what is the best way to implement this, so if you have any suggestions, let me know!

mrava87:
I totally agree with all your points :) 

For the laplacian i was thinking the same as what you did. Just add those to the aslinearoperator after passing it back :)

Let me take a look at the other 3 whilst I review this. For MatrixMult I think we may need to make a breaking change as dims has a different meaning to all other dims here, it’s more like extradims.. if we do this then we can easily make dims and dimsd inside the unit, what do you think?

pylops/signalprocessing/DWT.py
```diff
@@ -104,15 +105,16 @@ def __init__(self, dims, axis=-1, wavelet="haar", level=1, dtype="float64"):
         if isinstance(dims, int):
```
mrava87, line 105: is this still needed given that you apply `_value_or_list_like_to_tuple` after?

mrava87:
A few more comments:

- following `Gradient` I also added all the other input parameters to `Laplacian`
- shall we add to `Attributes` dims and dimsd like you did for the FFT (especially now that all operators have them?) - if yes, as its just docstrings, we can leave it to another PR
- I modified `MatrixMult` as I suggested in the previous comment, will push this and few other small things tomorrow morning :)

cako:
Great, I think it is ready now!

mrava87:
Agree :)

I think we are left with `Patch*D` and `Sliding*D` which are actually among the ones I struggle myself sometimes to reshape properly for printing, ahahah.... but I think I can add `dims` and `dimsd`. I'll do that and then merge this PR!

PR #353, opened 2022-03-26
Support for non-flattened array multiplication

cako:
Follow-up to #348.

This draft PR implements a basic support for non-flat vector, in hopes of spurring further discussion on design choices for this feature.

The key part of the implementation relies in [`LinearOperator.dot`](https://github.com/cako/pylops/blob/9ecd38010c7a2c7a9b8a0a230f563ed225451e4d/pylops/LinearOperator.py#L233) identifying whether `dims` and `dimsd` are present. If this is true, it flattens the array and passes it onto `LinearOperator.matvec`. Upon the return of a (flattened) vector, it then reshapes to `dimsd` any array that came in as a `dims`-shaped operator. In this way, matrix-matrix multiplications remain unchanged and do no support ndarrays.

An example of expected behavior using the Symmetrize operator can be found in [FlattenDims.ipynb](https://github.com/cako/pylops_notebooks/blob/master/developement/FlattenDims.ipynb), summarized below:

|     dims   |    dimsd   | expected dimsd | correct? |
--------------|--------------|----------------------|-------------|
|   (20, 20) |   (30, 20) |     (30, 20)   | ✓|
|     (400,) |    error   |       error    | ✓|
|(2, 10, 20) |    error   |       error    | ✓|
|    (2, 10) |    (3, 10) |      (3, 10)   | ✓|
| (2, 10, 1) |    error   |       error    | ✓|
|      (20,) |      (30,) |        (30,)   | ✓|
|    (20, 1) |    (30, 1) |      (30, 1)   | ✓|

There is an argument for using matmat for `(2, 10, X)` instead of erroring, but for the sake of simplicity this was not implemented.

In addition to the changed to `LinearOperator`, I am also providing a decorator to simplify the reshape/flatten pattern, exemplified in this PR by Symmetrize:
 https://github.com/PyLops/pylops/blob/9ecd38010c7a2c7a9b8a0a230f563ed225451e4d/pylops/basicoperators/Symmetrize.py#L75-L93

### Next steps
If we want to proceed with this, there are a few things to sort out. First, we need to decide which shapes are accepted, and which are not. The above table is only a suggestion.

Once that is defined, we need to make dims/dimsd "first-class" attributes which get passed onto transpose/adjoint, multiplications, etc.

### DIFF
--- a/pylops/LinearOperator.py
+++ b/pylops/LinearOperator.py
@@ -253,12 +253,32 @@ class LinearOperator(spLinearOperator):
         elif np.isscalar(x):
             return _ScaledLinearOperator(self, x)
         else:
-            if x.ndim == 1:  # or x.ndim == 2 and x.shape[1] == 1:
-                return self.matvec(x)
+            dims = getattr(self, "dims", None)
+            dimsd = getattr(self, "dimsd", None)
+
+            reshape_out = False
+            if x.shape == dims and dimsd is not None:
+                reshape_out = True
+                x = x.ravel()
+
+            if x.ndim == 1:
+                y = self.matvec(x)
+                if reshape_out:
+                    y = y.reshape(dimsd)
+                return y
             elif x.ndim == 2:
                 return self.matmat(x)
             else:
-                raise ValueError("expected 1-d or 2-d array or matrix, got %r" % x)
+                raise ValueError(
+                    (
+                        "Wrong shape.\nFor vector multiplication, expects either a 1d "
+                        "array or, an ndarray of size `dims` when `dims` and `dimsd` "
+                        "both are available.\n"
+                        "For matrix multiplication, expects a 2d array with its first "
+                        f"dimension is equal to {self.shape[1]}.\n"
+                        f"Instead, received an array of shape {x.shape}."
+                    )
+                )
 
     def div(self, y, niter=100):
         r"""Solve the linear problem :math:`\mathbf{y}=\mathbf{A}\mathbf{x}`.
--- a/pylops/avo/avo.py
+++ b/pylops/avo/avo.py
@@ -627,7 +627,7 @@ class AVOLinearModelling(LinearOperator):
                 + self.spatdims
             )
         y = self.ncp.sum(self.G * x[:, :, self.ncp.newaxis], axis=1)
-        return y
+        return y.ravel()
 
     def _rmatvec(self, x):
         if self.spatdims is None:
@@ -641,4 +641,4 @@ class AVOLinearModelling(LinearOperator):
                 + self.spatdims
             )
         y = self.ncp.sum(self.G * x[:, self.ncp.newaxis], axis=2)
-        return y
+        return y.ravel()
--- a/pylops/utils/_internal.py
+++ b/pylops/utils/_internal.py
@@ -1,4 +1,47 @@
+from functools import wraps
+
 import numpy as np
+
+
+def reshape_flatten(forward=True):
+    """Decorator for the common reshape/flatten pattern used in many operators.
+
+    Example
+    =======
+    A ``_matvec`` (forward) function can be simplified from
+
+    .. code-block:: python
+
+        def _matvec(self, x):
+            x = np.reshape(x, self.dims)
+            y = do_things_to_reshaped(y)
+            y = y.ravel()
+            return y
+
+    to
+
+    .. code-block:: python
+        @reshape_flatten()
+        def _matvec(self, x):
+            y = do_things_to_reshaped(y)
+            return y
+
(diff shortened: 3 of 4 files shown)


pylops/LinearOperator.py
```diff
@@ -253,12 +253,32 @@ def dot(self, x):
         elif np.isscalar(x):
             return _ScaledLinearOperator(self, x)
         else:
-            if x.ndim == 1:  # or x.ndim == 2 and x.shape[1] == 1:
-                return self.matvec(x)
+            dims = getattr(self, "dims", None)
+            dimsd = getattr(self, "dimsd", None)
+
+            reshape_out = False
```
mrava87, line 259: Can you please add a comment here, explaining what this piece of code does... that it will reshape the output back into a ndarray of size dimsd when the input had shape dims

pylops/LinearOperator.py
```diff
@@ -253,12 +253,32 @@ def dot(self, x):
         elif np.isscalar(x):
             return _ScaledLinearOperator(self, x)
         else:
-            if x.ndim == 1:  # or x.ndim == 2 and x.shape[1] == 1:
-                return self.matvec(x)
+            dims = getattr(self, "dims", None)
```
mrava87, line 256: ultimately we should avoid this check by making sure we always have dims and dimsd (same as shape and dtype)

mrava87: I checked

```
In [7]: %timeit getattr(Mop, "dims", None)
60.5 ns ± 0.315 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [8]: %timeit Mop.dims
37.2 ns ± 1.3 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

though I am not sure if we care as it is an irrelevant time difference...

pylops/avo/avo.py
```diff
@@ -627,7 +627,7 @@ def _matvec(self, x):
                 + self.spatdims
             )
         y = self.ncp.sum(self.G * x[:, :, self.ncp.newaxis], axis=1)
-        return y
+        return y.ravel()
```
mrava87, line 630: We should consider making this operator consistent with others, using `dims` and `dimsd` :)

Also, why is this needed now. In the current dev:

```
(pylops.avo.AVOLinearModelling(np.ones(5),nt0=100) * np.ones(300)).shape
Out[21]: (500,)
```
and
```
(pylops.avo.AVOLinearModelling(np.ones(5),nt0=100, spatdims=20) * np.ones(300*20)).shape
Out[23]: (10000,)
```

pylops/utils/_internal.py
```diff
@@ -1,6 +1,49 @@
...
+    def decorator(func):
+        @wraps(func)
+        def wrapper(self, x):
+            x = x.reshape(getattr(self, inp_dims))
+            y = func(self, x)
+            return y.ravel()
+
+        return wrapper
+
+    return decorator
```
mrava87, line 44: this looks very good, probably we should add this routine to the documentation in the PyLops utilities (it is true it is only for developers but it is something a bit more `exposed` than things like `_value_or_list_like_to_array` and it may start to be present in most operators.

What do you think?

cako:
I reorganized this entirely based on the idea of using properties in LinearOperators introduced in #365. The current version works like this:

1. Every LOp has `dims` and `dimsd`. They are arbitrary tuples which must be compatible with shape.
2. Suppose that LOp.dims = (5, 10) and LOp.dimsd = (9, 10). The following rules shapes are obtained on LOp @ x

   |Reference|     x.shape   |    (LOp @ x).shape   | Note | 
   ---|--------------|--------------|-----|
   V0|      (50,) |      (90,) | Standard vector multiplication
   V1|    (5, 10) |    (9, 10) | "Vector" of size (5, 10)
   M0|    (50, 1) |    (90, 1) | Standard one-column matrix multiplication
   M1| (5, 10, 1) |    (9, 10, 1)   | "One-column matrix" of "vector" (5, 10)
   M2|   (50, 20) |   (90, 20) | Standard matrix multiplication
   M3|(5, 10, 20) |    (9, 10, 20)   | "Matrix" of 20 x (5, 10)
   ||     (1000,) |    error   | Can be reshaped to (50, 20) but size is ambigous
   || any other kind of shape | error |
3. SciPy solvers will struggle with V1, M1 and M3. These cases can be coerced to error if LOp.flattened is set to True. Any operator with flattened = True will have legacy behavior. I'm happy to hear about other ways on how we can do this.

In addition, I have moved and refactored the ``reshape`` decorator. It now relies primarily on the name of the function being decorated to find out which dimensions to reshape by, and also includes an optional swapaxis. See the new Symmetrize operator for examples.

cako:
Moved to #367

Issue #352, opened 2022-03-24
Improve linting


marcusvaltonen:
I installed flake8 to check for code quality and ran it on pylops. It quite quickly finds a lot of flaws in the code quality. Some of the more common once are:

```
F401 '<module-name>' imported but unused
E203 whitespace before ':'
```

What is the line length? What I can find in the pre-commit hook it is 88 characters, but at multiple places this is exceeded.
Moreover flake8 is not an aggressive linter compared to e.g. pylint, so I believer this is a minimal requirement.

## Suggestion
Enforce PEP8 using flake8 and e.g. flake8-import-order. Also consider enabling some of pylint's checkers for code quality (at least initially). This can be enforced as a pre-commit hook or in the CI-pipeline.

cako:
Hi @marcusvaltonen,

Thank you for the issue. Indeed, the code is not perfectly PEP8 compliant, not least due to some of the choices regarding line lengths.

* Many strings are longer than the 88 chosen for black. The reason for this is that black does not format strings by default. [This is an experimental feature](https://github.com/psf/black/issues/1802) I though prudent to not enable until it is stable. That's not to say that we would be very happy to receive some commits fixing those issues! It is just that solving them automatically may mess things up now and in the future.
* The whitespace issue is also a well known point of contention between black and flake8. [Black's argument is that E203 is not PEP8 compliant](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#slices), so there isn't even an option to format it differently.
* Regarding unused imports, this should definitely not be happening. However, there are some slightly annoying situations that flake8 will flag as "unused import" in `__init__.py` files. For example here is a simplified version of `__init__.py` under `avo/` which complains of unused imports:
  ```python
  from .poststack import *
  __all__ = ["AVOLinearModelling"]
  ```
  Here is one that does not complain:
  ```python
  from .poststack import AVOLinearModelling
  __all__ = ["AVOLinearModelling"]
  ```
  While PEP8 discourages (but does not ban) wildcard imports (even allowing an exception), this is a bit of a gray area. However, flake8 complains, mostly because of the wildcard import. But not using the wildcard is a bit annoying since it requires a lot of repetition. But maybe that's for the best? I might be a good idea to remove the wildcard imports, as it will allow flake8 and other linters more introspection...
[...]

PR #354, opened 2022-03-27
fix: flake8 warnings and errors

cako:
Following #352, but not clo sing it, this PR removes all errors pointed out by the following linting config:
```
[flake8]
ignore = E203, E501, W503, E402
per-file-ignores =
    __init__.py: F401, F403, F405
max-line-length = 88
```

A good follow-up to this PR would be to cl ose #352 by incorporating a linter into the development process somehow. Since that is still under debate, this PR will does note attempt that.


mrava87:
Looks good, thanks for doing this :)

mrava87 (approved):

Issue #351, opened 2022-03-23
scipy.linalg.solve slower than numpy.linalg.solver for dense matrices.


mrava87:
## Problem
See discussion in https://github.com/PyLops/pyproximal/issues/47

## Solution
Include the possibility to choose numpy solver in `LinearOperator.div`

PR #356, opened 2022-03-29
feature: add option to choose dense solver to div

mrava87:
Following https://github.com/PyLops/pylops/issues/351, this commit includes the possible to choose whether to use scipy (default, for backward compatibility) or numpy solve when dealing with explicit operators.

### DIFF
--- a/pylops/LinearOperator.py
+++ b/pylops/LinearOperator.py
@@ -4,7 +4,9 @@ import logging
 
 import numpy as np
 import scipy as sp
-from scipy.linalg import eigvals, lstsq, solve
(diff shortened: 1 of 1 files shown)


mrava87:
@cako @marcusvaltonen, if you agree, I can make a second PR targeting the v1.18.x branch and make a release so that the problem of using a slower solver (scipy) as default can be at least avoided by explicitly choosing numpy - all of this applies only to explicit matrices, of course

marcusvaltonen:
I think it is a good compromise.

cako (approved):

Issue #358, opened 2022-03-31
Bug: Convolve1D when dims is scalar and method='direct'


mrava87:
# Problem
The new version of Convolve1D without the parameter `N` and only `dims` is bugged when working with 1d signals (`dims` is a scalar) and choosing `method=direct`. This is because the `_choose_convfunc` internal method checks for `dims` being `None` instead of being a scalar (previous behaviour in the presence of `N`

# Solution
A simple change can be introduced in the `_choose_convfunc` to check the type of dims instead of checking for None. In case this is int (or float to be on the safe side) we need to enter the first block of if statement which selects the correct method for the 1d case. This allows the `direct` one.

cako:
Closed by #359

PR #359, opened 2022-03-31
fix: handle case with dims scalar in Convolve1D

mrava87:
See https://github.com/PyLops/pylops/issues/358 for details.

### DIFF
--- a/pylops/signalprocessing/Convolve1D.py
+++ b/pylops/signalprocessing/Convolve1D.py
@@ -15,7 +15,7 @@ def _choose_convfunc(x, method, dims):
 
     Choose and return the function handle to be used for convolution
     """
-    if dims is None:
+    if isinstance(dims, (int, float)):
         if method is None:
             method = "direct"
         if method not in ("direct", "fft"):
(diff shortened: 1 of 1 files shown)


cako (changes requested):
I didn't investigate this problem in convolve very deeply when I was coding the dims/dimsd, but it seems like I introduced the bug. However, I think we should refrain in general from using `isinstance` like this. The reason is the following: the code below produces `True, False, False`:

```python
one = 1
one_np32 = np.ones(1, dtype=np.int32)[0]
one_np64 = np.ones(1, dtype=np.int64)[0]

print(one, isinstance(one, int))
print(one_np32, isinstance(one_np32, int))
print(one_np64, isinstance(one_np64, int))
```

I created a function which you can use when you need to check if int/float, etc: https://github.com/PyLops/pylops/blob/30a7af4054bd7847b2fa7e8b90e1b1afdafdb103/pylops/utils/_internal.py#L51

With that said, in this case, I think you can use a different solution. `self.dims` will *always* be a tuple. So if you just want to check that the input is one dimensional, you need to check if `len(self.dims) == 1`. Then, you call `_choose_convfunc` with `self.dims` instead of `self.dimsorig`. I haven't tested this (it might screw up something else), but that's the approach I'd try first.

mrava87:
Alright, thanks! 

I need to remember the methods you created and avoid isinstance ;) let me check the option you suggested, if it works I think it’s better than checking types

cako (approved):

cako:
closes #358

PR #361, opened 2022-04-05
feature: added scalings to Patch2d

mrava87:
Added option to use patch-wise scalings.


mrava87:
I had this sitting there for long time, probably not super useful but you never know :)

cako (approved):

Issue #90, opened 2019-06-16
Class based solvers


mrava87:
## Change

Move from function based to class based solvers. This would require creating a base class with the following:

- `__init__`
- `step`: run a single step 
- `callback`: run after any step (a user can just overwrite it - currently is passed to the solver as function)
- `print`: print update
- `solve`: run entire solver

## Motivation

- This allows running multiple inversion with different solver parametrization (niter, tol, etc) without having to rerun the preparation part, which could be time consuming and independent on the solver parameters.
- This allows users to run both the solver as is or run it in a more fine-grained fashion (using `step`). 

## Countermotivation

- This is a breaking change, is it worth?
- Scipy has function based solvers, if we move away from them (for also `cg`, `cgls`, `lsqr`) we will need to call them differently inside our apps depending on whether we use numpy or cupy arrays - right now we just call a different solver but with a very similar function signature.

marcusvaltonen:
This is really the way to go. I just want to add code reusability as part of the motivation.

PR #363, opened 2022-04-09
Class-based solvers

mrava87:
## Motivation
This PR is aimed at tackling https://github.com/PyLops/pylops/issues/90, namely create class-based solvers in PyLops and provide thin-wrapper for the equivalent function based solvers (to ensure backward compatibility and minimum modification to current user-code).

## Rationale
The general idea is to create a template solver called `Solver` in `basesolver.py` and sub-class all solvers from it.
I will not provide here a detailed description of how the solver has been designed here, as I attempted to do so in a new page in the documentation (file called `addingsolver.py`. 

This PS is currently meant to create some discussion among developers and mature the overall solution prior to implementing it on every solver currently available in PyLops.

@cako @marcusvaltonen, I look forward to your comments prior to moving this forward


marcusvaltonen:
I need to go through it in more detail, but at a first glance it looks good. One question though, that I have seen in other places in pylops: why not ABCMeta declare the base classes and put the abstractmethod decorator on the methods that are mandatory? It is standard practice in my opinion, and with good reason.

mrava87:
Great, have a look in more details and let us know what you think. There is no rush :)

Regarding the abstract method, you mean something like this https://stackoverflow.com/questions/7196376/python-abstractmethod-decorator. I was not aware of it, but yes it seems like a good idea of making sure if some methods are not implemented to raise an immediate error to the user.

marcusvaltonen:
Okay, I thought it was so common that you made an active decision of not using it. But if not, it is a best practice of defining templates in python. As you point out it ensures that all mandatory methods are implemented at the time of instantiation, so you don't get errors raised later in the pipeline. It also makes sure you are not instantiating the base classes itself. I would highly recommend using it. (And the stackexchange thread is for Python 2.7, nowadays you would write it differently).

```
from abs import ABCMeta, abstractmethod


class Base(metaclass=ABCMeta):
    @abstractmethod
    def method(self, param):
        pass


class Derived(Base):
    def method(self, param):
        return param
```

The `abstractmethod` works together with the `property` decorator too, so you can enforce abstract properties too.

pylops/optimization/solver.py
```diff
@@ -2,10 +2,202 @@
...
+            self.r = self.y - self.Op.matvec(self.x)
+        self.c = self.r.copy()
+        self.kold = self.ncp.abs(self.r.dot(self.r.conj()))
+
+        # create variables to track the residual norm and iterations
+        self.cost = []
+        self.cost.append(np.sqrt(self.kold))
+        self.iiter = 0
+
+    def step(self, show=False):
```
marcusvaltonen, line 112: I would not rely on properties here. A proper input and output is the clear way to go; it communicates to the user what is persistent information and what is used locally. Essentially I mean keep the same interface as the `solve()` method.

pylops/optimization/solver.py
```diff
@@ -2,10 +2,202 @@
...
+            msg = f"{self.iiter:6g}    {np.real(self.x[0]):4.1e} + {np.imag(self.x[0]):4.1e}j    {self.cost[self.iiter]:11.4e}"
+        print(msg)
+
+    def _print_finalize(self):
+        print(
+            f"\nIterations = {self.iiter}        Total time (s) = {time.time() - self.tstart:.2f}"
+        )
+        print("-----------------------------------------------------------------\n")
+
+    def setup(self, x0=None, niter=None, tol=1e-4, show=False):
```
marcusvaltonen, line 76: Part of this can also be in the base class, and then just call super() for the implementation of CG

pylops/optimization/solver.py
```diff
@@ -2,10 +2,202 @@
...
+        <10x10 Identity with dtype=float64>
+        >>> cgsolve = CG(I, np.ones(10))
+        >>> cgsolve.callback = callback
+
+        >>> cgsolve.callback()
+        Running callback
+        """
+        pass
+
+    def solve(self, x0=None, niter=10, tol=1e-4, show=False):
```
marcusvaltonen, line 173: Also move to baseclass? Most of the solver specific stuff is in `self.step()`. Again I suggest having a different interface,
```
x = self._step(x, *args)
self.callback(x)
```

pylops/optimization/solver.py
```diff
@@ -2,10 +2,202 @@
...
+
+        This routine must be passed by the user as follows (when using the `solve`
+        method it will be automatically invoked after each step of the solve)
+
+        Examples
+        --------
+        >>> import numpy as np
+        >>> from pylops.basicoperators import Identity
+        >>> from pylops.optimization.solver import CG
+        >>> def callback():
```
marcusvaltonen, line 159: What is the meaning of this callback? Why doesn't it take `x` input as before? It doesn't seem as useful to be as before.

pylops/optimization/solver.py
marcusvaltonen: I understand that you want to do `cg = CG(*args)` followed by `cg.setup()` and `cg.step()` to step-by-step debug your algorithm. I see that use-case, although it might not be a very common one (accept in development of the solver itself).

I see, this needs more careful though. I would probably design the steps a bit different. Perhaps

```
cg = CG(Op, x)
out = cg.solve(y)
```

I don't like that you are relying so much on properties. I have seen projects that rely on `self.xxx` being set, but then you want to add an extra feature down the road and it is not set anymore or incorrectly possibly messing up something else. As far as I am concerned the intermediate steps don't need to be saved as a property. I think it is a better design pattern. But, again, these are just some initial thoughts being blurted out late at night... I need to revisit them.

marcusvaltonen (commented):
Some very hasty comments. I would like to re-iterate this and have a continued discussion.

pylops/optimization/solver.py
marcusvaltonen: What if I want to use the callback for testing my solver against the ground truth, e.g.
```
history = []
def my_callback(x)
    history.append(np.linalg.norm(x - x_gt))
```
This is a good use-case, and perhaps you want to plot the error per iteration `plt.plot(history)`. You can't do this anymore, with this design.

pylops/optimization/solver.py
marcusvaltonen: I would like this method to return as before. I don't see why you would call solve and then manually pick out the parts yourself is a better design choice.

pylops/optimization/solver.py
```diff
@@ -2,10 +2,202 @@
...
+       vol. 49. 1952.
+
+    """
+
+    def __init__(self, Op, y):
+        self.Op = Op
+        self.y = y
+        self.ncp = get_array_module(y)
+
+    def _print_setup(self):
```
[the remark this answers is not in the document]
mrava87: mmh apart from `_print_finalize (which I moved to the base class) I am not so convinced about the benefit of doing this... there is quite some difference between solvers. 

Ok, they all instantiate start and print some similar lines but eventually they diverge quite a lot. I worry we won't gain much. Let's get all solvers in place and then we can look at

pylops/optimization/solver.py
```diff
@@ -2,10 +2,202 @@
...
+        self.tstart = time.time()
+        if self.niter is None:
+            strpar = "tol = {self.tol:10e}\tniter = {self.niter}"
+        else:
+            strpar = "tol = {self.tol:10e}"
+        print(
+            "CG\n"
```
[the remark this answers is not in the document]
mrava87: Good idea. See comment above.

pylops/optimization/solver.py
mrava87: Alright, I am a bit confused then. I wasn't so sure about class-based solvers (even though I myself opened a GitHub issue) but then when you suggested this I thought perhaps it is a good idea and we should give it a try.

But, if we simply design a class based solver to be 

```
cg = CG(Op, x)
out = cg.solve(y)
```

then I am not so sure I see the point. Can you explain where you see the benefit of using these 2 lines over the current `out=cg(Op,y,x)`?

My initial design was based on the idea that if we have a modular solver composed of setup-step-finalize a user could run one step at the time in a for loop and do anything else in between (similar to torch). On the other hand, a developer/interested user that wants to dig deeper in the solver working would likely need to focus only on the step method.

The point about relying on properties is a good one. Again, my initial idea was that to uniquely define a problem you must define:

- operator
- data
- solver

You could then run the solve method many times with different starting guesses and parameters. I definitely do not think it is good to initialize the solver with `x`, `y` makes much more sense (unless we want to only pass `Op` to begin with and pass `y` to `setup`. But then again we will rely on properties. However if we don't do that we will need to pass it to step and that sounds weird to me...

@cako, if you have time it would be good to have your opinion here too :)

pylops/optimization/solver.py
mrava87: _print_finalize is now also in the base call.

I am not sure I understand why we need to call super() though

pylops/optimization/solver.py
mrava87: Sorry, this was a mistake. `callback` should be removed from CG and just put in base class.

mrava87: Mmmh I see your point but again I am not convinced I agree. Now we have an `x` that we keep passing in and out of the step...

I personally prefer the idea of relying on self.x, @cako we need your opinion here ;)

pylops/optimization/solver.py
mrava87: _print_finalize is now also in the base call.

I am not sure I understand why we need to call super() though

pylops/optimization/solver.py
mrava87: Sorry, this was a mistake. callback should be removed from CG and just put in base class.

pylops/optimization/solver.py
mrava87: `Why doesn't it take x input as before?` Sorry this is a mistake, we should pass `x`

`What if I want to use the callback...?` This is the main use case of callback. You can simply use a lambda function 

```
def callback(xx, xhist):
    xhist.append(xx.copy())
...
cgsolve.callback = lambda x: callback(x, xhist)
```

This is how I would suggest to treat the callback in the current function based implementation and I don't think this can change for class based simply because inside the `solve` method we don't know what callback would look like and the best way is to enforce it to have one single input. Do you have a better suggestion?

mrava87:
If you want to give it a go yourself and get a better feeling for this, have a look at this notebook https://github.com/PyLops/pylops_notebooks/blob/master/developement/SolverClass.ipynb

pylops/optimization/solver.py
mrava87: Regarding solve returning values, this makes sense. I can do that in the next commit :)

cako:
Sorry guys, I haven't had time to look at this yet, but I will!!

pylops/optimization/solver.py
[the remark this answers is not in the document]
marcusvaltonen: Okay but I'm referring to the *formatting*, not the content. Like the linewidth, how many decimal places and spacings between columns and stuff like

```
print("-----------------------------------------------------------")
```

pylops/optimization/solver.py
marcusvaltonen: Well, the way it is defined now that won't be necessary.

pylops/optimization/solver.py
marcusvaltonen: I am often faces with a different usecase: I want to solve the same problem for different kinds of data over and over again. So, in my view the data is not really "part of the problem". I could do something like

```
cg = CG(Op, x)
for i, y in enumerate(lots_of_y):
    out[i] = cg.solve(y)
```

Secondly, the main benefit is the encapsulation, code re-usability and templating (making it clearer for users what to define for new solvers). It allows for standardization of printing outputs and most likely makes things like switching to GPUs easier since all solvers are streamlined.

pylops/optimization/solver.py
marcusvaltonen: I don't love the self.x. Just make sure to set it to None before you return or something. So that the value is not left and someone forgets to call `setup()` and end up in weird situations. Of course these are user error, but a well-designed framework should be built to capture these things. Such issues could not arise if you keep `x` local.

pylops/optimization/solver.py
marcusvaltonen: As long as `x` is passed there is no issue. I thought it was removed for some other reason. I don't want to change the interface.

pylops/optimization/solver.py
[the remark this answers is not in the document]
mrava87: Then I don’t understand… you mean we could have

print(‘-‘*N) 

but then what do we do we make a _print_line() in the base class and call it when we want a line… is this what you have in mind?

mrava87:
Thanks @marcusvaltonen for the additional a comments, I’m going to try to adapt this to satisfy as many needs as possible:

- running the solver with multiple data 
- running the solver with multiple initial guesses 

In the meantime can you please mark as resolved those conversations you think have been addressed, so I’m sure I know where more work is needed :)

marcusvaltonen:
> Thanks @marcusvaltonen for the additional a comments, I’m going to try to adapt this to satisfy as many needs as possible:
> 
> * running the solver with multiple data
> * running the solver with multiple initial guesses
> 
> In the meantime can you please mark as resolved those conversations you think have been addressed, so I’m sure I know where more work is needed :)

I would also like initial guess as optional

```
out=cg.solve(y, x0) 
```

pylops/optimization/solver.py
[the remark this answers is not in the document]
marcusvaltonen: I guess it is hard to say when only a single solver is in the PR. Just leave it out from this PR and when all Solvers are class based it can be refactored. Just an idea on my end, to separate the formatting from the content. Essentially, the specific solver should never call print() but be handled in the base class. I could probably have a go at it later.

pylops/optimization/solver.py
mrava87: Mmh ok, this depends on who you are and what you do... the other use case is also quite common, you want to solve the same problem with different starting guesses.

I would say we should try to make sure we accommodate for as many 'useful' use cases as possible. My new suggested flow is as follows.


In words:

- Create the solver object
- Setup the solver with data, solver hyperparameters (and, optionally, the initial guess). This will return the initial guess to be passed to step
- Step the solver Niter times passing x (at iter i-1) and obtaining x (at iter i)
- Optionally interleave step with callback
- Finalize solver

In code:

```
cgsolve = CG(Aop)

xhist = []
cgsolve.callback = lambda x: callback(x, xhist)

x = cgsolve.setup(y=y, x0=np.zeros(nx), tol=0, show=True)
for i in range(ny):
    x = cgsolve.step(x, show=True if i % 10 == 0 else False)
    cgsolve.callback(x)
cgsolve.finalize(show=True)
```

mrava87: Note that all our solvers are interchangeable between numpy (CPU) and copy (GPU), so the switching to GPU thing is already well taken into account and things won't change for the class based solvers

pylops/optimization/solver.py
mrava87: ok, then you need to explain better as I don't understand... please provide a short example of what you mean

pylops/optimization/solver.py
mrava87: Ok, for this I think I would like a second opinion (@cako). 

If the problem is only `x`, I agree we can keep this out of the properties of the solver (see my last push if that makes sense, where I never even set self.x so I don't need to do the ugly thing of `self.x=None`... but if the problem is that you would like not to rely on properties at all, for Op, y, residual, hyperparams, etc. (which I hope is not what you mean), then I worry we are doing the Java mistake here, where we want to do something functional programming is good at but we do it with classes...

pylops/optimization/solver.py
marcusvaltonen: Your example snippet is very long,I don't see a normal user writing that many lines. Everything that is not possible in two steps:

```
cg = CG(Op)
our = cg.solve(y, *args, **kwargs) 
```

More than that is not very user-friendly. It works for both multiple data and starting positions as well.

marcusvaltonen: > Also, note that all our solvers are interchangeable between numpy (CPU) and copy (GPU), so the switching to GPU thing is already well taken into account and things won't change for the class based solvers

Sure but the base class could call `get_array_module`, then it would be a bit more clear for a new user how to use it. But, agreed, not a mind-blowing change.

pylops/optimization/solver.py
marcusvaltonen: I only have an issue with x, not the rest.

pylops/optimization/solver.py
marcusvaltonen: Sorry, I was confused that you mentioned `_print_finalize`, but the original comment was regarding `setup()`.
My idea, as I recall it now, was to move the common parts for solvers, e.g. initialization of `x0`, `self.cost` and the starting value, `self.iiter` to the base class, then call `super()` ad add the solver specific parameters.

pylops/optimization/solver.py
```diff
@@ -2,10 +2,186 @@
...
+            Number of executed iterations
+        cost : :obj:`numpy.ndarray`, optional
+            History of the L2 norm of the residual
+
+        """
+        self.setup(x0=x0, niter=niter, tol=tol, show=show)
+        while self.iiter < niter and self.kold > self.tol:
+            show = (
+                True
+                if self.iiter < 10 or niter - self.iiter < 10 or self.iiter % 10 == 0
```
marcusvaltonen, line 175: Consider not hard-coding 10.

mrava87: > I would also like initial guess as optional
> 
> ```
> out=cg.solve(y, x0) 
> ```
> 

It is in now :)

mrava87: And good point about not hard-coding 10... this is spicy behavior, reason why I followed it in the first place but we can make it more flexible for sure

mrava87: Have a look at the new parameter ``itershow``

pylops/optimization/solver.py
mrava87: Oook good, then it makes sense to me :)

pylops/optimization/solver.py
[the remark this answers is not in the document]
mrava87: Alright, I am going to change all solvers once we agree on the format. I won't merge this PR until that is done. 

So for now I will keep this comment unsolved and we can pick it up again and see what can be done

pylops/optimization/solver.py
mrava87: Exactly, a common user will never do more that the two lines you write (or if they prefer and have to solve a single instance of the problem even use the associated function call). This was more meant to be what we do under the hood in `solve` or what an advanced user can do directly if they want to be flexible (eg do some plotting) and not do it within callback.

Regarding `get_array_module`, let me think about... maybe a better way is to create within the base class:

- setup
- solve
- finalize

which all have do something that will be in common for most solvers, and in each solver:

- _setup
- _solve
- _finalize

that we ask dev to implement when making a new solver and does the specific thing of that solver. An example

```
def setup(self, y=None, x0=None, show=False, **kwargs):
       self.y = y
        self.ncp = get_array_module(y)
        x = self._setup(x0, **kwargs)

        # print setup
        if show:
            self._print_setup(np.iscomplexobj(x))
        return x
```

and 

```
    def _setup(self, y=None, x0=None, niter=None, tol=1e-4, show=False):
 
        self.tol = tol
        self.niter = niter

        # initialize solver
        if x0 is None:
            x = self.ncp.zeros(self.Op.shape[1], dtype=self.y.dtype)
            self.r = self.y.copy()
        else:
            x = x0.copy()
            self.r = self.y - self.Op.matvec(x)
        self.c = self.r.copy()
        self.kold = self.ncp.abs(self.r.dot(self.r.conj()))

        # create variables to track the residual norm and iterations
        self.cost = []
        self.cost.append(np.sqrt(self.kold))
        self.iiter = 0
        
        return x
```

If this is more along the lines of what you had in mind (note this also makes the _print being called by the base class), I can do the same for the other methods?

pylops/optimization/solver.py
mrava87: Oh right, I see now. 

The idea of setup is to avoid the __init__ doing too much. For example for a solver like ISTA I foresee the computation of the max eigenvalue to be more suited for setup than for the init method (as a matter of preference I would like init to be almost instantaneous). So we could define some 'super'-hyperparameters that are always present (eg niter) and move them to the init, would this be better for you?

pylops/optimization/solver.py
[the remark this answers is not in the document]
cako: Sorry for being late to the party, but I think TensorFlow is a good inspiration for this.  Their `model.fit` method takes a verbose option, which is pretty much the same across all solvers. In their case they show things like iterations/epochs. 

In our case, solvers are completely free to process the data in any way they like. In this case, I don't see a benefit in having a base class method handling it. For more complex printing, TensorFlow uses callbacks which can be register for any kind of event, including printing (methods called "on_step", "on_solve", "on_setup", etc.). I think we could probably benefit from this approach here.

pylops/optimization/solver.py
cako: I agree with this. The init should only register the class attributes. Any real computation should be elsewhere (i.e., setup).

cako:
Sorry for taking so long to get to this. In general my perception is very positive about this feature. I think the only thing that is missing for me is a more flexible way of add callbacks for each step (on_step, on_setup, etc.). We could also provide some basic callbacks like iteration tracker and so on. In my vision the callbacks would be implemented as classes, but in a way that you can easily create them from functions. Just my 2¢!

pylops/optimization/solver.py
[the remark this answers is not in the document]
mrava87: Alright, I do not have much experience with Tensorflow but I had a look at it (https://www.tensorflow.org/api_docs/python/tf/keras/Model#fit) and I think I get your point. Let me try to see if we agree:

- their `model.fit` is going to be a bit like our `solve`. I think this is Keras (ok now it is all in Tensorflow) so I compare it to our `solve` method... we actually allow now two levels of 'complexity': high level, Keras like, or low level with `step` and letting user do the for..loop and anything else, Torch or pure Tensorflow like. Makes sense?
- so within `solve` we have a `show` (we always had it and I would not change it into verbose to be consistent with spicy which is the API we are the most aligned with) and now`itershow=[N1, N2, N3]` which allows choosing at which iterations to show the progress (e.g. first N1, last N2, and every N3 in between). **Do you feel we should have more degrees of verbosity/showing like in Tensorflow (even though they also say people should use 0 or 2..)?**
- for the callbacks, we currently allow only one, are you suggesting we should allow a list? In tensorflow, it looks to me you have a Callback class and you can register multiple callbacks. The full list is here, https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/Callback#expandable-1 (I don't see the ones you mention unless I am looking in the wrong place) and then inside fit those are called at specific places... are you suggesting we do something along these lines? This sounds interesting but I find it a bit different from the  _print_xxx() we are discussing here, unless you think **we should also turn also all prints into callbacks?** We could then have predefined callbacklists for all solvers, and allow user to instantiate them and modify any of the callbacks. To me we may be up to making something more complicated that it should be as we provide pretty much custom printing for every solver and I don't see why someone would want to change them (but maybe I am wrong ;) ). So we currently have:

     * _print_setup
     * _print_step
     * _print_finalize
     * callback

**Do you see the need for more, any suggestion?**

pylops/optimization/solver.py
mrava87: Oook, then I think I am going to resolve this one. Having now worked on refractoring 3 solvers, I think the only 3 things that all solvers have in common are operator, data, and initial model... as we want the latter two to be passed at setup and solve stage...

I think also for solvers like FISTA, setup will trigger the computation of the Lipschitz constant and if we call again setup/solve (with for example a new y) we can simply check that if it is already being computed and stored we just don't do it again.

pylops/optimization/solver.py
mrava87: I tried this but didn't seem to add much value and a lot more verbosity. I think for now I'll stick with what I have

mrava87:
Thanks @cako!

As I replied to my comment above, I think the idea of having more callbacks sounds interesting. Let me think about it a bit, I want to find something useful for example at setup stage so to make a working example that makes sense for you to take a look :)

mrava87:
One more thing I would like to get some feedback: defining class solvers as 'XX' and their equivalent function solvers as 'xx' **in the same file** leads to a problem in the documentation. Only one page is made as I think sphinx cannot handle pages with same name and just difference in small/capital. 

One simple way around this is to make a new file, I was thinking to call it `wrappers.py` that contains only the function based solvers. What do you think? Any other alternative suggestion?

cako:
I dug into this a big and apparently it's an issue in case-insensitive filesystems (like Mac by default). So they added a fix for this which goes in conf and you basically create a dictionary with the ones you dont want to "lose". Maybe give it a go (see https://github.com/sphinx-doc/sphinx/pull/7927)? Not sure it's very sustainable though...

pylops/optimization/solver.py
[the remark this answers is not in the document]
cako: * Yup, I think it is a good approach.
* I'm OK with the current verbosity levels, but this can be changed by callbacks, which lead me to:
* Yes, TensorFlow implements a callback class, which could be overhead for us, but they get an extremely flexible "engine" with it. Every TF callback has methods like "on_train_end", "on_epoch_begin" and [things like that](https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/Callback#methods). This gives the user fine-gained introspection in case they want to print/store/manipulate any part of the fitting procedure. Where I think we can benefit from this: I was envisioning that each solver can have an "on_(setup|step|solve)_(begin|end)" list of callbacks. Now, the reason these callbacks should probably be a class and not a function is because with a class you can store information from any and all steps/iterations/solves.
   Here is an example: I would like to lsqr solve and print the percentage change in the norm of the data residual. This requires that I keep information from the previous steps. Another example: I want a "patience"-like stopper related to this gradient norm: if the data residual is increasing at a slow pace, stop or change something. The nice thing about this is that verbosity can be changed very easily by creating a LoggerCB(verb=0,1,2,3,...) that just implements prints.
   I know that many of the callbacks they implement are related to the nonlinear nature of the solvers, but not all of them. If we think we can benefit from some of them (see [some default ones](https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/)), I think it would be worth the effort (and overhead). None of this would change the fact that the simple function interface should still be available (and be very easy to use!).

mrava87:
> I dug into this a big and apparently it's an issue in case-insensitive filesystems (like Mac by default). So they added a fix for this which goes in conf and you basically create a dictionary with the ones you dont want to "lose". Maybe give it a go (see [sphinx-doc/sphinx#7927](https://github.com/sphinx-doc/sphinx/pull/7927))? Not sure it's very sustainable though...

Mmh interesting, thanks for finding it :) At the moment I think I landed on a solution where I split class solvers from their equivalent function solvers in two files (all in a single file was starting to become way too long), so this problem is not a problem anymore ;)

mrava87:
@cako, I think I am done (or at least all solvers have been now converted to class-based 💥 

I have also made a proposal for a class called `Callbacks` and shown how I envision its usage in a new tutorial `classsolvers.py`. Have a look and revert if you think this makes sense and it is along the lines of what you had in mind. So far it is implemented only for 3 solvers (IRLS, ISTA, FISTA) but I can easily do that for all the others once we like the design.

I had to make a lot of changes to tutorials/examples as I want all solvers now to have signature `Op, y,..., otherargs, x0=, other kwargs` which was not the case before. Do you agree we should make sure all solvers are as homogenous as possible even if that introduces a number of breaking changes?

cako (commented):
Very nice PR! I think it's shaping up really well! I have a few suggestions before we move forward with it. (Let me know if you'd like me to tackle them myself).

1. I'm starting to get a bit worried with the user experience of importing solvers. Important functions/classes need to be readily available to the user. For example, to import CG (class-based) we need to do `from pylops.optimization.basic import CG`. It is not clear either what "basic" means, nor that CG is a class-based solvers. I have a suggestion to improve this: move the importing of all solvers to a `pylops.optimization.solvers`. The `pylops/optimization/solvers.py` file could look like this:
   ```python
   from ._basic, ._basicc, ... import *
   ```
    I understand the original intention of separating sparse/other solvers, but I would try to favor a more unified approach and take advantage of the fact that we can break things with 2.0. If not, we could at least rename basic to something else, e.g., systemofequations, (and move lsqr to leastsquares.py where it should be more at home).
2. This leads me to the naming conventions. Playing around with the code for a couple of hours I misspelled basicc for basic more than once. I'm predicting that this naming convention is going to become messy and lead to many annoying bugs soon hahaha. Here is a suggestion: rename all solver classes from Xyz to XyzSolver, and place them in the same file as the function-based one (which is often just a simple wrapper). So when the user does something like `from pylops.optimize.solvers import CGSolver` it is clear that CG is a class-based solver.
3. Provided the above suggestions are accepted, the documentation could look something like this:
   ```rst
   Solvers
   -------
   Template
   ~~~~~~~~
   .. currentmodule:: pylops.optimization.basesolver
  
   .. autosummary::
     :toctree: generated/
  
       Solver
  
   Conjugate gradients
   ~~~~~~~~~~~~~~~~~~~
  
   .. currentmodule:: pylops.optimization.solvers
  
   .. autosummary::
      :toctree: generated/
      cg
      CGSolver
   ```
   Importantly, for this to look nice, the classes and functions have to have a descriptive first line in the documentation, e.g., r"""Conjugate gradient class-based solver""".
4. The callbacks are pretty much spot on with what I was thinking. And I think we can take them even further by beefing up out base `Solver` class. I wrote a proof of concept to register the callbacks automatically upon `__init__`. That way, when implementing a new solver, the developer does not need to call `callback.on_(setup|step|run)_(begin|end)` manually within the implementation of the class. The parent `__init__` can take care of that. Have a look [at this gist](https://gist.github.com/cako/f9117ffba970fb2c3c2823fd1e318ea1) for an idea of how that could work (the annotations is just to improve readability).
5. I found some typos here and there, I wrote them down locally and I'll take care of them further down the road (I didn't want to change any files right now).

cako:
> I had to make a lot of changes to tutorials/examples as I want all solvers now to have signature `Op, y,..., otherargs, x0=, other kwargs` which was not the case before. Do you agree we should make sure all solvers are as homogenous as possible even if that introduces a number of breaking changes?

Absolutely!

mrava87:
Thanks, I also like the way it is coming along. Let me reply to your points, I can then handle them and you can do one final round adding your changes before merge. However, since this has been a long ride, my suggestion is that once happy, we close this PR and I make a new one with a number of targeted commits (eg one for base, one for each submodule, and one for tests and examples) to have a clean history. And I write a summary in the PR of what has changed based on the discussions we had here... we can always look back at this closed PR for more details. Do you agree?

`1. I'm starting to get a bit worried with the user experience of importing solvers...`

I see this but I am not sure about your suggestion. Then why not **simply exposing solvers at the `pylops.optimization` level.** By working on the `__init__` file we can allow people just doing `from pylops.optimization import CG`? And we can **make all other files "private" (eg `_basicc`, `_basic`)**? 

`1. I understand the original intention of separating sparse/other solvers, but I would try to favor a more unified approach and take advantage of the fact that we can break things with 2.0. If not, we could at least rename basic to something else, e.g., systemofequations, (and move lsqr to leastsquares.py where it should be more at home).`

Alright, if you agree with the above, this becomes more of a developer-side thing. At the start of PyLops, I wanted to create handy wrappers over spicy solvers which could add regularization/precond. Here the initial name `leastsquares`. With time, mostly due to the cupy backend, I had to create solvers like cg,cgls,lsqr as they were not yet supported in cupy with LinearOperators (also this turned out useful as in some cases scipy solvers don't allow for example callbacks). The name `solver` was mostly to differentiate them from something more high level like what we have in the `leastsquares` module. I see the point you make about lsqr, but then also cgls falls in that category.. which will leave us with only cg in the `basic`. I am not sure what is better name than `basic` or `leastsquares` but **I want to keep a distinction between a low-level solver like `lsqr` and a high-level one like `regularized_inversion` so I am not in favor of moving them to the same file**. If you see better names than `basic` and `leastsquares` I am happy to consider them, but they need to be more explicative... if it is about the same, then I would favor a conservative approach where we don't change them.

`2. This leads me to the naming conventions. `

For developers I am not worried about the fact that `_basicc`, `_basic` are similar (but different) and should not lead to bugs if people are careful about what they do (I think)... For users, the fact of having capital/CamelCase for classes and small for functions should make it easy for users to understand what is what, I don't particularly like adding prefix/suffix to things unless they really add something. The fact our solvers used to be CamelCase but functions is what I tried to fix now. In short, **I prefer we don't add `Solver` but use python conventions to hint what is class and what is function.**

`3. Provided the above suggestions are accepted, the documentation`
Mmmh not sure, sorry! I really don't like `XyzSolver`, so what about:


Solvers
-------
Template
~~~~~~~
.. currentmodule:: pylops.optimization

.. autosummary::
   :toctree: generated/

    basesolver.Solver

Basic
~~~~

.. currentmodule:: pylops.optimization._basicc

.. autosummary::
   :toctree: generated/

   CG
   CGLS
   LSQR

.. currentmodule:: pylops.optimization._basic

.. autosummary::
   :toctree: generated/

   cg
   cgls
   lsqr

etc
```

so they will just appear together. Again, if anything better than Basic I am happy but I would not split each pair of solvers into a category...

`4. The callbacks are pretty much spot on with wh`
Great. I like the idea you suggest. I wanted to find a way to avoid having to change all solvers and I was thinking about decorators but I just decided I'd wait until I get your feedback on the general setup. Let me try this based on your gist and let you know how it goes :)

`5.I found some typos here and there, I wrote `
As I said above, let's tackle these last few things, then you can have a final pass making those minor changes :)

**Comment by marcusvaltonen** [comment]
Hi, just wanted to say that I think the changes are shaping up nicely. I like the snake_case=function, CamelCase=class convention and the callbacks.

File: docs/source/addingsolver.rst [hunk 8]
```diff
@@ -0,0 +1,248 @@
...
+At this point, we need to implement the core of the solver, its `step`. Here, we take the input at previous iterate,
+update it following the rule of the solver of choice, and return it. The other input parameter required by this method
+is ``show`` to choose whether we want to print a report of the step on screen or not. However, if appropriate, a user
+can add additional input parameters. For CG, the step is:
+
+.. code-block:: python
+
+    def step(self, x, show=False):
+        Opc = self.Op.matvec(self.c)
+        cOpc = self.ncp.abs(self.c.dot(Opc.conj()))
```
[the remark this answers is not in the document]
mrava87: I deliberately use matvec/rmatvec when possible inside the solver because they avoid the various checks that dot does (which is triggered by @). And you can see this in the tutorial Linear Operators. Do you think this is a wise strategy or an overkill? 

Maybe I should explain this in the documentation instead of switching to @?

docs/source/addingsolver.rst
```diff
@@ -0,0 +1,248 @@
...
+        a = self.kold / cOpc
+        x += a * self.c
+        self.r -= a * Opc
+        k = self.ncp.abs(self.r.dot(self.r.conj()))
+        b = k / self.kold
+        self.c = self.r + b * self.c
+        self.kold = k
+        self.iiter += 1
```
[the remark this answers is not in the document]
mrava87: Good point, done :)

docs/source/addingsolver.rst
```diff
@@ -0,0 +1,248 @@
...
+calling the ``step`` method. It is also usually convenient to implement a finalize method; this method can do any required post-processing that should
+not be applied at the end of each step, rather at the end of the entire optimization process. For CG, this is as simple
+as converting the ``cost`` variable from a list to a numpy array. For more details, see our implementations for CG.
+
+Last but not least, we can wrap it all up in the ``solve`` method. This method takes as input the data, initial
+model and the same hyperparameters of setup and runs the entire optimization process. For CG:
+
+.. code-block:: python
+
+    def step(self, y, x0=None, niter=10, tol=1e-4, show=False, itershow=[10, 10, 10]):
```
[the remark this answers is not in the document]
mrava87: Yep :)

docs/source/addingsolver.rst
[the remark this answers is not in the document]
cako: Ok, let's keep it matvec for now. I was mostly thinking about when we need to extend it for ndarray support.

cako:
Sure, let's get this one wrapped up after this comment.

1. Ok, how about this. Under PyLops API Menu, we change "Solver" to Optimization. "Basic" becomes "Base Solvers", "Least Squares" becomes "Least Squares Inversion" and "Sparsity" becomes "Sparse Inversion". Then we just expose everything through `pylops.optimization` and, privately, we can have files like `_base_solvers_fun.py`, `_base_solvers_cls.py`, `_leastsquares_fun.py`, `_leastsquares_cls.py`, etc. What do you think?
  My reasoning behind this is the following: I don't want people to be looking at docs all the time. Things should be easy to find. If they know pylops has an optimization namespace, the first try is to grab it from there. I mainly want to avoid the situation with scipy: to grab a simple function, you need to dig through like 3 namespaces. On the docs, however, we want to be explicit, so we can separate into logical units. In fact I got the idea from Scipy, e.g., `cg` and `lsqr` are in the same namespace ([scipy.sparse.linalg](https://docs.scipy.org/doc/scipy/reference/sparse.linalg.html#module-scipy.sparse.linalg)) but they are logically separated in the docs.
2. If you agree with the above, nothing to do on this point.
3. Or on this point
4. Cool, let me know if I can help!
5. Agreed!

mrava87:
1. Alright, I suspected this as I remember seeing something along these lines long time ago. If we try to import all solvers under namespace `pylops.optimization` we get the following
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [1], in <cell line: 12>()
      9 import matplotlib.pyplot as plt
     10 import scipy as sp
---> 12 import pylops
     13 from scipy.sparse.linalg import cg as sp_cg
     14 from pylops import CG, CGLS, LSQR

File ~/Desktop/KAUST/OpenSource/pylops/pylops/__init__.py:49, in <module>
      1 """
      2 PyLops
      3 ======
   (...)
     45 
     46 """
     48 from .config import *
---> 49 from .LinearOperator import LinearOperator
     50 from .basicoperators import *
     51 from . import (
     52     avo,
     53     basicoperators,
   (...)
     57     waveeqprocessing,
     58 )

File ~/Desktop/KAUST/OpenSource/pylops/pylops/LinearOperator.py:27, in <module>
     24     from scipy.sparse.linalg._interface import _ProductLinearOperator
     26 from pylops import get_ndarray_multiplication
---> 27 from pylops.optimization._basicsolvers_fun import cgls
     28 from pylops.utils.backend import get_array_module, get_module, get_sparse_eye
     29 from pylops.utils.estimators import trace_hutchinson, trace_hutchpp, trace_nahutchpp

File ~/Desktop/KAUST/OpenSource/pylops/pylops/optimization/__init__.py:38, in <module>
     35 from ._basicsolvers_cls import CG, CGLS, LSQR
     36 from ._basicsolvers_fun import cg, cgls, lsqr
---> 38 from ._leastsquares_cls import NormalEquationsInversion, RegularizedInversion, PreconditionedInversion
     39 from ._leastsquares_fun import normal_equations_inversion, regularized_inversion, preconditioned_inversion

File ~/Desktop/KAUST/OpenSource/pylops/pylops/optimization/_leastsquares_cls.py:7, in <module>
      4 from scipy.sparse.linalg import cg as sp_cg
      5 from scipy.sparse.linalg import lsqr
----> 7 from pylops.basicoperators import Diagonal, VStack
      8 from pylops.optimization.basesolver import Solver
      9 from pylops.optimization._basicsolvers_fun import cg, cgls

File ~/Desktop/KAUST/OpenSource/pylops/pylops/basicoperators/__init__.py:43, in <module>
      1 """
      2 Basic Linear Operators
      3 ======================
   (...)
     40     SecondDirectionalDerivative     Second Directional derivative.
     41 """
---> 43 from .FunctionOperator import *
     44 from .MemoizeOperator import *
     45 from .Regression import *

File ~/Desktop/KAUST/OpenSource/pylops/pylops/basicoperators/FunctionOperator.py:6, in <module>
      1 from numbers import Integral
      3 from pylops import LinearOperator
----> 6 class FunctionOperator(LinearOperator):
      7     r"""Function Operator.
      8 
      9     Simple wrapper to functions for forward `f` and adjoint `f_c`
   (...)
     64     array([2.,  3.])
     65     """
     67     def __init__(self, f, *args, **kwargs):

TypeError: module() takes at most 2 arguments (3 given)
```

I have played a bit and dont seem to be able to find a solution that is nice (i.e., doesn't mess up codes or forces to merge contents of files like `basic.py` and `leastsquares.py` solvers, which I don't want to do.

The best solution I could come up with is to just merge class and function based solvers in one file (as the function ones are actually very small) and allow users to import things in two ways.

They could import:
```
from pylops.optimization.basicsolvers import CG, CGLS, LSQR
from pylops.optimization.leastsquares import NormalEquationsInversion, RegularizedInversion, PreconditionedInversion
from pylops.optimization.sparsity import IRLS, OMP, ISTA, FISTA, SPGL1, SplitBregman
from pylops.optimization.basicsolvers import cg, cgls, lsqr
from pylops.optimization.leastsquares import normal_equations_inversion, regularized_inversion, preconditioned_inversion
from pylops.optimization.sparsity import irls, omp, ista, fista, spgl1, splitbregman
```
or
```
from pylops import CG, CGLS, LSQR
from pylops import NormalEquationsInversion, RegularizedInversion, PreconditionedInversion
from pylops import IRLS, OMP, ISTA, FISTA, SPGL1, SplitBregman
from pylops import cg, cgls, lsqr
from pylops import normal_equations_inversion, regularized_inversion, preconditioned_inversion
from pylops import irls, omp, ista, fista, spgl1, splitbregman
```

I know this is not exactly what you wants (having `pylops.optimization`) but at the moment I can't really find a way for this to happen. Any suggestion?

mrava87:
4. I think there may be a problem here. I envision a callback to have access to every property of the solver plus the model `x`. Based on previous discussions we decided x should NOT be an attribute of the class, rather only passed in and out of methods. I do see an easy way to modify your `_registercallbacks` such that it can accommodates for this situation. Any suggestion?

mrava87:
4. Carlos, what do you think? I made a few changes to allow `x0` in setup and `x` in step and run ;)

cako (approved):

mrava87:
Mmh, I wasn’t done with this… I’ll make another small PR where I merge all class based and function based files so we can import them as I suggested above. Or shall we keep it as is with the xxxc for class based and xxx for function based?

PR #364, opened 2022-04-09
CI: added flake8 github action

mrava87:
Handles https://github.com/PyLops/pylops/issues/352

### DIFF
--- a/.github/workflows/flake8.yaml
+++ b/.github/workflows/flake8.yaml
@@ -18,5 +18,5 @@ jobs:
       - name: flake8 Lint
         uses: py-actions/flake8@v2
         with:
-          ignore: "E203,E501,W503,E402"
+          ignore: "E203,E501,W503,E402,F401,F403,F405"
           max-line-length: "88"
--- a/.github/workflows/flake8.yaml
+++ b/.github/workflows/flake8.yaml
@@ -18,5 +18,5 @@ jobs:
       - name: flake8 Lint
         uses: py-actions/flake8@v2
         with:
-          ignore: "E203, E501, W503, E402"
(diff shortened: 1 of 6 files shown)


cako (changes requested):
Could you also add
```
[flake8]
ignore = E203, E501, W503, E402
per-file-ignores =
    __init__.py: F401, F403, F405
max-line-length = 88
```
to `setup.cfg` and 
```makefile
lint:
	flake8 docs/
	flake8 examples/
	flake8 pylops/
	flake8 pytests/
	flake8 tutorials/
```
to `Makefile`?

marcusvaltonen:
I think it looks good, including cakos change request. One thing regarding the makefile target, though. Defined this way, if an error occurs in "docs" will the other lines get executed? Isn't it better to run it as a one-liner, i.e. "flake8 path1 path2", then you minimize the overhead from running flake and also make sure all packages are checked. Otherwise, there is always the risk that someone running "make lint" corrects the flake issues only on partial parts of the package.

cako:
> One thing regarding the makefile target, though. Defined this way, if an error occurs in "docs" will the other lines get executed? Isn't it better to run it as a one-liner, i.e. "flake8 path1 path2", then you minimize the overhead from running flake and also make sure all packages are checked.

You are absolutely right. I was having this issue and didn't realize that was the problem. Thanks!

mrava87:
Done. I followed @marcusvaltonen suggestion to call flake8 once on all targets. Also I added it as req in the requirement-dev and environment-dev files, and I added some description in the list of steps to contribute code.

Have a look and merge if you agree :)

cako (approved):

PR #365, opened 2022-04-11
Refactor LinearOperator to allow propagation of attributes

cako:
## Description
While working on #353, it became clear that `dims` and `dimsd` needed to become "first-class" attributes of `LinearOperator`. To do so in a clean way, I decided to refactor some parts of the base class. In this refactor, `shape` and `dims`/`dimsd` are now properties that complement each other, with the user only needing to supply one of them. I also changed `clinear` and `explicit` for properties which default to `True` and `False`, respectively.

The refactor is completely compatible with current code. However, the recommended way of implementing new operators from now on should be by calling `super().__init__` with the attributes as arguments. Of course, the old way of setting attributes individually also works, but will generally result in longer and more cluttered code.

Finally, with `dims` and `dimsd` as "first-class" attributes, they can now be propagated to operators derived from them through mathematical operations such as addition and multiplication. For example, if operator `B` has `dims = (10, 10)` and `dimsd = (20, 10)` and operator `A` has `dims = (100, 3)` (and any `dimsd` compatible with `B.shape[1] == 100`), then `C = B @ A` will have `dimsd = (20, 10)` and `dims = (100, 3)`.


pylops/LinearOperator.py
```diff
@@ -47,32 +48,211 @@ class LinearOperator(spLinearOperator):
...
+                getattr(self.Op, "explicit", False) if explicit is None else explicit
+            )
+            if explicit and hasattr(Op, "A"):
+                self.A = Op.A
+            name = getattr(Op, "name", None) if name is None else name
+
+        if dtype is not None:
+            self.dtype = dtype
+        if shape is not None:
+            self.shape = shape
```
mrava87, line 106: since it is a property now should it not be `self._shape`? Same for all others...

pylops/LinearOperator.py
```diff
@@ -47,32 +48,211 @@ class LinearOperator(spLinearOperator):
...
+            dimsd = getattr(self, "_dimsd", None)
+            if dims is None or dimsd is None:  # Cannot find both dims and dimsd, error
+                raise AttributeError(
+                    (
+                        f"'{self.__class__.__name__}' object has no attribute 'shape' "
+                        "nor both fallback attributes ('dims', 'dimsd')"
+                    )
+                )
+            _shape = (np.prod(dimsd), np.prod(dims))
+            self._shape = _shape  # Update to not redo everything above on next call
```
mrava87, line 131: why is this needed here but not outside of the if clause?

pylops/LinearOperator.py
```diff
@@ -47,32 +48,211 @@ class LinearOperator(spLinearOperator):
...
+        return _shape
+
+    @shape.setter
+    def shape(self, new_shape):
+        new_shape = tuple(new_shape)
+        if not isshape(new_shape):
+            raise ValueError(f"invalid shape %{new_shape:r} (must be 2-d)")
+        dims = getattr(self, "_dims", None)
+        dimsd = getattr(self, "_dimsd", None)
+        if dims is not None and dimsd is not None:  # Found dims and dimsd
```
mrava87, line 141: is there a case where setting a new shape will not make it incompatible with dims/dimsd?

If not, it sounds like changing shape is never a good idea and should be forbidden maybe?

mrava87 (commented):
Looks really nice. I just have a few minor questions here and there but I think this should definitely be merged soon :)

mrava87:
Just one question, shall we after this make a new PR where all operators use the new `super` instead of directly defining dims/dimsd?

pylops/LinearOperator.py
cako: I thought about this, but I opted to keep using the setter, since has checks/coercions to ensure the values are "legal". For example, setting `self._shape = [3, 2]` would be acceptable, but `self.shape = [3, 2]` is nicer since it would coerce into a tuple (`self._shape == (3, 2)`).

pylops/LinearOperator.py
cako: A simple getter with `default_value` could look like this:
```python
@property
def shape():
    if getattr(self, "_shape", False):
        return default_value
    return self._shape
```
In the actual getter, all I something similar: I check that `self._shape` has not been initialized yet (for example, the `__init__` function in a new operator does not provide shape, but provides `dims`/`dimsd`), and if it hasn't, then it can try to fall back on `dims`/`dimsd`. If those values do exist, we can just store them the first time they are computer. Moreover, they are "legal" (a 2-tuple of ints which agree with `dims`/`dimsd`), so they can bypass the checks. It would be perfectly OK to do `self.shape = _shape` in line 131, but it would just execute more code for no added benefit.

pylops/LinearOperator.py
cako: This is also a good point. Currently you can change a shape as long as you don't have dims/dimsd. Maybe this doesn't make sense.

cako:
> Just one question, shall we after this make a new PR where all operators use the new `super` instead of directly defining dims/dimsd?

Yes, I can take care of that when the PR is ready!

pylops/LinearOperator.py
mrava87: Right, I see.

pylops/LinearOperator.py
mrava87: Ok, I see. Once we have all operators with dims/dimsd in place then probably this will never be the case but for now we can allow it :)

mrava87:
Alright, for me this is good to go

Issue #264, opened 2021-11-17
Can FISTA run on matrices of data?


eurunuela:
Hello,

I am not sure this is the right place to ask this. I would love to add `pylops` to my algorithms, but I mainly work with multivariate formulations, i.e., with the entire matrix of data at once.

Does `pylops.optimization.sparsity.FISTA` accept an `np.ndarray` of size `200,10` for instance? I have tried it but cannot make it work.

If I can use FISTA like that, I'll definitely want to contribute to add a new operator.

Thank you!

mrava87:
Hi @eurunuela,
currently it will not work as internally we call `matvec` and `rmatvec` but it should be an easy change to make it work for multiple right-hand sides at the same time :) 

I'll take a look and let you know how it goes soon

eurunuela:
Thank you @mrava87 !

I thought it would be straightforward too, just making sure that the matrix dimensions make sense for `matcvec`.

mrava87:
Ok, this was easier than I expected.

I just pushed updated ISTA/FISTA on the master. Try:

```
nt = 61
nx = 5
dt = 0.004
t = np.arange(nt)*dt
x = np.zeros(nt)
x[10] = -.4
x[int(nt/2)] = 1
x[nt-20] = 0.5
x = np.outer(x, np.ones(nx))

h, th, hcenter = pylops.utils.wavelets.ricker(t[:101], f0=20)

Cop = pylops.signalprocessing.Convolve1D(nt, h=h, offset=hcenter,
                                         dtype='float32')
y = Cop*x
yn = y + np.random.normal(0, 0.1, y.shape)

xista, niteri, costi = \
    pylops.optimization.sparsity.ISTA(Cop, y, niter=400, eps=5e-1,
                                      tol=1e-8, returninfo=True)

xfista, niteri, costi = \
    pylops.optimization.sparsity.FISTA(Cop, y, niter=400, eps=5e-1,
                                       tol=1e-8, returninfo=True)

fig, axs = plt.subplots(1, 4, figsize=(14, 5))
axs[0].imshow(x)
axs[0].axis('tight')
axs[1].imshow(y)
axs[1].axis('tight')
axs[2].imshow(xista)
axs[2].axis('tight')
axs[3].imshow(xfista)
axs[3].axis('tight')
```

This is a simple example of deconvolution using multiple 1d signals as multiple RHS. Have a look and see if this is what you need for your problem :)

eurunuela:
[...]

PR #367, opened 2022-04-16
Support for NDArray multiplication

cako:
This PR takes over from #353 in introducing valid multiplication by ND-arrays without the need for reshaping
The current version works like this:

1. Every LOp has `dims` and `dimsd`. They are arbitrary tuples which must be compatible with LOp.shape.
2. Suppose that LOp.dims = (5, 10) and LOp.dimsd = (9, 10). The following shapes are obtained on LOp @ x

   |    | Reference   | x.shape                 | (LOp @ x).shape                           | Note                                             |
   |----|-------------|-------------------------|-------------------------------------------|--------------------------------------------------|
   | V0 | (50,)       | (90,)                   | Standard vector multiplication            |                                                  |
   | V1 | (5, 10)     | (9, 10)                 | "Vector" of size (5, 10)                  |                                                  |
   | M0 | (50, 1)     | (90, 1)                 | Standard one-column matrix multiplication |                                                  |
   | M1 | (5, 10, 1)  | (9, 10, 1)              | "One-column matrix" of "vector" (5, 10)   |                                                  |
   | M2 | (50, 20)    | (90, 20)                | Standard matrix multiplication            |                                                  |
   | M3 | (5, 10, 20) | (9, 10, 20)             | "Matrix" of 20 x (5, 10)                  |                                                  |
   |    |             | (1000,)                 | error                                     | Could be reshaped to (50, 20) but is ambigous |
   |    |             | any other kind of shape | error                                     |                                                  |
3. SciPy solvers will struggle with V1, M1 and M3. These cases can be coerced to error if LOp.flattened is set to True. Any operator with flattened = True will have legacy behavior. I'm happy to hear about other ways on how we can do this.

This PR also provides a `reshape`` decorator. It relies primarily on the name of the function being decorated to find out which dimensions to reshape by, and also includes an optional swapaxis call. See the new Symmetrize operator for examples.

### DIFF
--- a/pylops/LinearOperator.py
+++ b/pylops/LinearOperator.py
@@ -66,6 +66,17 @@ class LinearOperator(spLinearOperator):
         .. versionadded:: 1.17.0
 
         Operator is complex-linear. Defaults to ``True``.
+    flattened : :obj:`bool`, optional
+        .. versionadded:: 2.0.0
+
+        Defaults to ``False``. If ``False``, multiplications by N-dimensional arrays
+        are accepted if they have shape ``dims`` or ``(*dims, P)``. For an ``(M, N)``
+        operator, the respective outputs will have shaped ``dimsd`` and
+        ``(*dimsd, P)``.
+
+        When ``True``, the operator only multiplies flattened inputs, that is,
+        arrays with shapes ``(N,)`` or ``(N, P)``. The respective outputs will have
+        shapes ``(M,)`` and ``(M, P)``. This behavior is equivalent to PyLops v1.x.
     name : :obj:`str`, optional
         .. versionadded:: 2.0.0
 
--- a/pylops/LinearOperator.py
+++ b/pylops/LinearOperator.py
@@ -82,6 +82,7 @@ class LinearOperator(spLinearOperator):
         dimsd=None,
         clinear=None,
         explicit=None,
+        flattened=None,
         name=None,
     ):
         if Op is not None:
@@ -98,6 +99,9 @@ class LinearOperator(spLinearOperator):
             )
             if explicit and hasattr(Op, "A"):
                 self.A = Op.A
+            flattened = (
+                getattr(Op, "flattened", False) if flattened is None else flattened
+            )
             name = getattr(Op, "name", None) if name is None else name
 
         if dtype is not None:
@@ -112,6 +116,8 @@ class LinearOperator(spLinearOperator):
             self.clinear = clinear
         if explicit is not None:
             self.explicit = explicit
+        if flattened is not None:
+            self.flattened = flattened
         self.name = name
 
     @property
@@ -232,6 +238,18 @@ class LinearOperator(spLinearOperator):
         del self._explicit
 
     @property
+    def flattened(self):
+        return getattr(self, "_flattened", False)
+
+    @flattened.setter
+    def flattened(self, new_flattened):
+        self._flattened = bool(new_flattened)
+
+    @flattened.deleter
+    def flattened(self, new_flattened):
+        del self._flattened
+
+    @property
     def name(self):
         return getattr(self, "_name", None)
 
@@ -245,7 +263,7 @@ class LinearOperator(spLinearOperator):
 
     def _copy_attributes(self, dest, exclude=["name"]):
         """Copy attributes from one LinearOperator to another"""
-        attrs = ["dims", "dimsd", "clinear", "explicit", "name"]
+        attrs = ["dims", "dimsd", "clinear", "explicit", "flattened", "name"]
         if exclude is not None:
             for item in exclude:
                 attrs.remove(item)
@@ -301,6 +319,7 @@ class LinearOperator(spLinearOperator):
             Opx = aslinearoperator(x)
             self._copy_attributes(Op)
             Op.clinear = Op.clinear and Opx.clinear
+            Op.flattened = Op.flattened or Opx.flattened
             # Replace if shape-like
             if len(self.dims) == 1:
                 Op.dims = Opx.dims
@@ -467,6 +486,7 @@ class LinearOperator(spLinearOperator):
             Op = aslinearoperator(_ProductLinearOperator(self, x))
             self._copy_attributes(Op, exclude=["dims", "name"])
             Op.clinear = Op.clinear and x.clinear
+            Op.flattened = Op.flattened or x.flattened
             Op.dims = x.dims
             return Op
         elif np.isscalar(x):
@@ -484,12 +504,12 @@ class LinearOperator(spLinearOperator):
                 x = x.reshape((-1, x.shape[-1]))
             if x.ndim == 1:
                 y = self.matvec(x)
-                if is_dims_shaped:
+                if is_dims_shaped and not self.flattened:
                     y = y.reshape(self.dimsd)
                 return y
             elif x.ndim == 2:
                 y = self.matmat(x)
-                if is_dims_shaped_matrix:
+                if is_dims_shaped_matrix and not self.flattened:
                     y = y.reshape((*self.dimsd, -1))
                 return y
             else:
--- a/pylops/LinearOperator.py
+++ b/pylops/LinearOperator.py
@@ -474,12 +474,32 @@ class LinearOperator(spLinearOperator):
             self._copy_attributes(Op)
             return Op
         else:
-            if x.ndim == 1:  # or x.ndim == 2 and x.shape[1] == 1:
-                return self.matvec(x)
+            dims = getattr(self, "dims", None)
+            dimsd = getattr(self, "dimsd", None)
+
+            reshape_out = False
+            if x.shape == dims and dimsd is not None:
+                reshape_out = True
+                x = x.ravel()
+
+            if x.ndim == 1:
+                y = self.matvec(x)
+                if reshape_out:
+                    y = y.reshape(dimsd)
+                return y
             elif x.ndim == 2:
                 return self.matmat(x)
             else:
-                raise ValueError("expected 1-d or 2-d array or matrix, got %r" % x)
+                raise ValueError(
+                    (
+                        "Wrong shape.\nFor vector multiplication, expects either a 1d "
+                        "array or, an ndarray of size `dims` when `dims` and `dimsd` "
+                        "both are available.\n"
+                        "For matrix multiplication, expects a 2d array with its first "
+                        f"dimension is equal to {self.shape[1]}.\n"
+                        f"Instead, received an array of shape {x.shape}."
+                    )
+                )
 
     def div(self, y, niter=100, densesolver="scipy"):
         r"""Solve the linear problem :math:`\mathbf{y}=\mathbf{A}\mathbf{x}`.
--- a/pylops/LinearOperator.py
+++ b/pylops/LinearOperator.py
@@ -475,15 +475,23 @@ class LinearOperator(spLinearOperator):
             return Op
         else:
             is_dims_shaped = x.shape == self.dims
+            is_dims_shaped_matrix = len(x.shape) > 1 and x.shape[:-1] == self.dims
             if is_dims_shaped:
+                # (dims1, ..., dimsK) => (dims1 * ... * dimsK,) == self.shape
                 x = x.ravel()
+            if is_dims_shaped_matrix:
+                # (dims1, ..., dimsK, P) => (dims1 * ... * dimsK, P)
+                x = x.reshape((-1, x.shape[-1]))
             if x.ndim == 1:
                 y = self.matvec(x)
                 if is_dims_shaped:
                     y = y.reshape(self.dimsd)
                 return y
             elif x.ndim == 2:
-                return self.matmat(x)
+                y = self.matmat(x)
+                if is_dims_shaped_matrix:
+                    y = y.reshape((*self.dimsd, -1))
+                return y
             else:
                 raise ValueError(
                     (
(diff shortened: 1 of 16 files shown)


mrava87:
Great, this is a nice addition! As I previously said, if this can be done is great as long as it doesn't make things confusing on the user side (for those users that don't care about this), and it seems like this is the case :)

I have a number of questions/comments:

1. You say `SciPy solvers will struggle with V1, M1 and M3.`. To the best of my knowledge scipy solvers can only handle single RHS (ie. 1d `b` vectors), and the documentation of eg [cg](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.cg.html) says that. I tried just to be sure:2. 

```
from scipy.sparse.linalg import cg as sp_cg

ny, nx, nrhs = 100, 50, 4
x = np.outer(np.ones(nx), np.arange(nrhs)+1)
A = np.random.normal(0, 1, (ny, nx))
A = A.T @ A
Aop = pylops.MatrixMult(A)
y = A @ x

xest = sp_cg(Aop, y, maxiter=4)

```
and as expected this breaks. So also M2 won't work. Only V0 (and apparently M0 for scipy cg but not for scipy lsqr). 

2. The same applies to our own solvers with the 'lucky' exception of ISTA/FISTA (see https://github.com/PyLops/pylops/issues/264). This is more by chance than by design, because ISTA/FISTA never requires computing a vector norm. Instead for any of the solvers in `pylons.optimization.solver` submodule, multiple RHS are not (yet) allowed because of this:

```
self.kold = self.ncp.abs(self.r.dot(self.r.conj()))
```

Then we would need to do:

```
self.kold = self.ncp.sum(self.ncp.abs(self.r * self.r.conj()), axis=0)
```

which is not out of scope (we can maybe do it after we have reorganized all the solvers in classes) but not possible today

2.

pylops/LinearOperator.py
```diff
@@ -82,6 +82,7 @@ def __init__(
         dimsd=None,
         clinear=None,
         explicit=None,
+        flattened=None,
```
mrava87, line 85: Seems like you have not added flattened to the doctoring?

pylops/optimization/leastsquares.py
```diff
@@ -108,6 +108,18 @@ def NormalEquationsInversion(
     """
     ncp = get_array_module(data)
 
+    # Ensure compatibility with SciPy functions
```
mrava87, line 111: Mmh I am not sure I understand this. If I understand correctly you simply check if the operator is not flattened you apply flattened, so that the solver will simply go from a multiple RHS problem to a flattened problem?

I think this is reasonable but comes with some caveats. Imagine you want to do a N-traces seismic inversion and write it as a single ``PoststackLinearModelling`` operator. You can then include a ``Laplacian`` regularizers to couple those traces. Now let's take a (perhaps stupid) case where you have 2 datasets and want to put them side by side into a N*2 matrix ``y`` and use the new option M2 to invert for a N*2 matrix  ``x``. This would flatten the operators and couple the laplacian on the first and second column of x?

I am just not sure, I am asking as you may have thought about ;)

pylops/utils/decorators.py
```diff
@@ -0,0 +1,76 @@
+from functools import wraps
+from typing import Callable, Optional
+
+
+def reshaped(
+    func: Optional[Callable] = None,
```
mrava87, line 6: mmh have we started introducing typing annotations?

I wonder if it is best to wait and then do this in one unique 'boring' PR so we make sure everything is done consistently across the library?

pylops/LinearOperator.py
cako: Done

pylops/optimization/leastsquares.py
cako: > Mmh I am not sure I understand this. If I understand correctly you simply check if the operator is not flattened you apply flattened, so that the solver will simply go from a multiple RHS problem to a flattened problem?

Not exactly. [The logic in the flattened part of the LinearOperator](https://github.com/PyLops/pylops/blob/c4abb6a54d5efdcf279a749b6d4acdbf184913fc/pylops/LinearOperator.py#L508-L513) does not flatten N-d array inputs. If the arrays are not shaped (N,) or (N, P), the multiplication will raise an error. This was the same behavior as the v1.x, so it is basically an easy way to ensure that your code doesn't break going from v1.x to v2. I'm tempted to make the default rely on a module-level variable but I haven't decided if that is a good idea or not. Something like `pylops.set_ndarray_multiplication(True/False)`.

In this particular function (leastsquares), I ensure all input operators are flattened = True, and then set them back. It's a bit hacky. The module-level implementation would make the code considerably easier, but globals are usually not a great idea.

pylops/utils/decorators.py
cako: Fair enough! Removed them.

pylops/optimization/leastsquares.py
cako: Just an addition to this. I introduced the flattened logic because of the annoying case when dims is flat i.e., dims = (N,) but dimsd is not e.g. dimsd = (M1, ..., MK). This case appears in interpolation. I previously (erroneously) assumed that flat inputs -> flat outputs. But that is clearly not the case, which means we can't rely on the shape of dims to decide whether to flatten the output or not. 

I figured the best solution was to separate the v2 behavior from v1. In v1 you always get an error when you try n-d array multiplication (unless its a properly shaped matrix). In v2, if we have a dims-shaped input, we will output a dimsd-shaped array. In the case above (dims is flat but dimsd is not), it *will* reshape the output. Taking the interpolation example, this makes sense.

cako:
1. Fair enough, I didn't realize this. But it's ok, we don't use it like that and we don't plan on using it like that.
2. Yea, we at some point need to decide if we want to support the matrix `b`. Regarding the n-d shaped inputs, I didn't notice any issues but maybe to be on the safe side we can also "protect" the operators (set them to flattened)? (Starting to think `pylops.set_nd_multiplication` is not such a bad idea after all 😅

pylops/optimization/leastsquares.py
mrava87: Mmh I’m confused. If you refer to Interp/Bilinear, I would say the input has dims.. if that is 1D (dims=(N,), how can you get an output that is ND?

mrava87: Regarding the first comment, so you put all operators to flattened=True so that their outputs are flat (which is what solvers need at the moment)?

cako: > Mmh I’m confused. If you refer to Interp/Bilinear, I would say the input has dims.. if that is 1D (dims=(N,), how can you get an output that is ND?

The issue I was having was with ``Bop.H`` in plot_bilinear.py. It takes a flat array and outputs a 2D array. Maybe I'm doing something wrong?

cako: > Regarding the first comment, so you put all operators to flattened=True so that their outputs are flat (which is what solvers need at the moment)?

Exactly. They will also error if the input is not flat (v1.x behavior).

mrava87: > > Mmh I’m confused. If you refer to Interp/Bilinear, I would say the input has dims.. if that is 1D (dims=(N,), how can you get an output that is ND?
> 
> The issue I was having was with `Bop.H` in plot_bilinear.py. It takes a flat array and outputs a 2D array. Maybe I'm doing something wrong?

That's correct. I was referring to the forward and that one goes from grid to random points, so the adjoint goes from random points to a grid :)

mrava87:
> * Yea, we at some point need to decide if we want to support the matrix `b`. Regarding the n-d shaped inputs, I didn't notice any issues but maybe to be on the safe side we can also "protect" the operators (set them to flattened)? (Starting to think `pylops.set_nd_multiplication` is not such a bad idea after all 😅

I think ultimately it is a good idea, as you can leverage as much as possible vectorization to apply the operators... it just needs some changes in the solver, which we can plan to do. What would `pylops.set_nd_multiplication` do?

cako:
My idea was to use this set_nd_multiplication method as a global way of turning the ND-array multiplication on/off. I don't think it actually makes sense to keep it on a per-operator basis as I have implemented it.

mrava87:
Alright I see. So that if it is turned off the matvec/rmatvec will always flattens the input arrays before passing them to the internal methods (_matvec/_rmatvec) or they will throw an error if an nd-array is passed?

cako:
It should throw an error (v1.x behavior)

pylops/optimization/leastsquares.py
cako: I just committed the changes for setting the ndarray support on a module level. This also creates a nice pattern to alter behavior within functions for backward compatibility like this:

```python
from pylops import disabled_ndarray_multiplication

def old_style_func(Op, y, ...):
    x = x.ravel() # Optional, otherwise will error on non-1d arrays
    with disabled_ndarray_multiplication():
        # do stuff in v1.x way
```

Let me know if that makes sense. If you do, just resolve the conversation :)

cako:
Hey @mrava87, I'm pretty happy with how it works now. I'm tempted to merge with dev before Transform 2022 if you are happy as well!

pylops/optimization/leastsquares.py
```diff
@@ -106,65 +107,66 @@ def NormalEquationsInversion(
     implicitly assumed to be zero.
 
     """
-    ncp = get_array_module(data)
+    with disabled_ndarray_multiplication():
```
mrava87, line 110: Why is this needed here? And not in other solvers in this module?

mrava87:
This looks good :) 

Just to be sure, I have two minor questions:

- you have so far created a framework for handling NDArray multiplications and changed only one operator to use the reshaped decorator (Symmetrize). I guess you will make a new PR following this for all the other operators where this code pattern make sense?

- you have flagged this as breaking. Is it really or by default this would not change any behaviors of v1. I ask because you have not added anything to the MIGRATION_V1_V2.md file. Have a think about if anything is needed there.

I think you can merge once you are happy with it.

pylops/optimization/leastsquares.py
cako: I was having an issue with this solver before because it calls the SciPy cg which does not accept NDarray multiplication. However, I looked into it a bit deeper and it seems that there are several instances in the codebase where methods that only work for 1d arrays are called (e.g. scipy lsqr and cg). They are mostly internal, and since out tests are all built for v1.x behavior, I didn't catch this.

One solution is to create a wrapper for these scipy functions so that users can understand how to interact with functions that might be reliant on v1.x behavior (but that they cannot change/are external). It would work something like this:

```python
# v1.x code: only works with 1d arrays
from scipy.sparse.linalg import lsqr
xlsqr, *_ = lsqr(Sop, y.ravel())

# proposed new code
def add_ndarray_support_scipy(func): # this function would be in pylops.utils.decorators
    """
    Decorator which converts a SciPy-solver-type 
    function into one that supports ndarray multiplication.
    """
    def wrapper(A, b, *args, **kwargs): # this is the signature of sparse solvers in scipy
        with disabled_ndarray_multiplication():
            res = func(A, b.ravel(), *args, **kwargs)
            res[0] = res[0].reshape(A.dims)
            return res

from scipy.sparse.linalg import lsqr
lsqr = add_ndarray_support_scipy(lsqr) # "monkey-patch"
xlsqr, *_ = lsqr(Sop, y) # works with ndarrays

```

cako:
1. Yes, I think the best course of action is to refactor the operators not only to use the best practices from this PR, but also from some previous ones. The main changes that would be done are: do not register shape or dims/dimsd directly, instead call `super().__init__(dims=dims, dimsd=dimsd)`.
2. I flagged this as breaking mostly because of some issues I had while running the tests after I implemented the this functionality. Yes, code that worked before *should* keep working, but there were a couple of places that had bugs that needed to be fixed before the change was actually non-breaking. Not sure if this is really "breaking", but better safe than sorry :) With regard to the migration, I will fill that out!

pylops/optimization/leastsquares.py
mrava87: Mmh, I see. Yes we have quite a few functions that call a solver, and if the input is a numpy array by default they would call the appropriate scipy solver  (already if it is cupy they will call our equivalent as at that time cupy did not have scipy solvers in place for linear operators.. I need to check if they do now). 

We could to begin with do this, and in the longer run when all our solvers are refactored change them so that they can handle multiple rhs, then check if this is the case and use them instead of scipy ones… or even switch all together to our solvers and just have some tutorial that shows you can use scipy solvers but we now reccomend using ours?

mrava87:
Btw, not sure if you want to do it now with Symmetrize or later with another operator, but I think having an example that shows how to use these new functionalities would be valuable, otherwise they may end up being ‘undiscovered’ ;)

pylops/optimization/leastsquares.py
cako: I implemented the above method (with some small alterations) and in general it works. I also added in the migration readme how to use operators with the new LinearOperator (either disable nd-array mult or decorate them).

The only thing missing are the sparse solvers, but I need to look into those a bit better. For now the solution to those is to disable nd-array support when calling them.

cako:
I added the NMO example, but I want to go through all the examples/tutorials and sort everything out. I'll merge before T22 so we can get more eyes on this!

mrava87:
Sounds good. For the NormalEquationInversion it is ok, but please do not spend time on sparsity as I plan to have all operators in class based mode (or at least a first draft) in the next week or so. So it’s better we wait and then see based on that what problems we get with the new Ndarray options and decide which changes to make. I NormalEquationInversion is actually ready for me to push. I will look at what you changed and include it there as well as for other solvers

cako:
Awesome, let's wrap up the solvers first. W.r.t the NormalEquationInversion, very few changes so it should be easy to make it work :)

Issue #368, opened 2022-04-16
Black failing in pre-commit


mrava87:
Based on https://github.com/psf/black/issues/2964 we should change this line in `.pre-commit-config.yaml` file:

```
-    rev: 21.12b0
+    rev: 22.3.0
```

marcusvaltonen:
I'm not @cako but I think it sounds OK:

cako:
Closed by #370

PR #370, opened 2022-04-16
bug: fix black version in pre-commit

mrava87:
See https://github.com/PyLops/pylops/issues/368.


mrava87:
@cako  I just changed laptop and pre-commit wasn't working (got an error for black).

Looking around I found the Github issue I refer to in https://github.com/PyLops/pylops/issues/368. Can you check if you agree with this change?

cako (approved):

PR #376, opened 2022-04-30
fix: bring documentation back alive after https://github.com/PyLops/pylops/pull/367/files#diff-ed7f0264f2d893898309d8acd9ec1e0a2c2ae0854131e2045c74638da5947040

mrava87:
This PR fixes problems in the documentation arising after https://github.com/PyLops/pylops/pull/367. It also highlights a possible bug, a possible inconsistent behavior, and a suggested change to https://github.com/PyLops/pylops/pull/367

## Possible bug

In `pylops/optimization/leastsquares.py`, https://github.com/PyLops/pylops/pull/367 introduced the following change:

```
dataregs = [ncp.zeros(int(Reg.shape[0]), dtype=Reg.dtype) for Reg in Regs] -> dataregs = [ncp.zeros(Reg.dimsd, dtype=Reg.dtype) for Reg in Regs]
```

I believe that this is incorrect and should be reverted. The reason for this being incorrect is that it would make a NDarray, and later lead to an dimension error at this line:

```
datatot = datatot - RegOp * x0
```

When calling the solver within `with pylops.disabled_ndarray_multiplication():`, which at the moment is required because the lsqr/cgls solvers called within  RegularizedInversion do not handle NDarrays and `add_ndarray_support_to_solver` cannot be used due to the function signature being `RegularizedInversion(Op, Regs, data, **kwargs)` and not `wrapper(A, b, x0=None, **kwargs)`

## Possible inconsistency

Currently ISTA and FISTA work also with multiple RHS (see https://github.com/PyLops/pylops/issues/264). Now if we wrap it in `@add_ndarray_support_to_solver`, we lose this property. On the other hand if we don’t and we have an operator that takes NDarrays and returns NDarrays and the solver is not called protected by the with `pylops.disabled_ndarray_multiplication()` (or the global one), we will also get an unwanted behaviour. Any idea how to get past this dilemma (without asking the user to know all these little details)?

## Suggested change 

For `add_ndarray_support_to_solver` to work across the board, all solvers    should have the following function signature `func(A, b, x0=None, **kwargs)`. This is not the case (eg RegularizedInversion, SplitBregman). My suggestion is to change this function slightly:

```
def wrapper(A, b, *args, **kwargs):  # SciPy-type signature
        if "x0" in kwargs:
            kwargs["x0"] = kwargs["x0"].ravel()
         with disabled_ndarray_multiplication():
            res = list(func(A, b.ravel(), *args, **kwargs))
```

and adapt all solvers to have the following `template` signature `solver(A, b, *args, **kwargs) where x0 is always the first of the kwargs. I have already implemented the rearrangement of parameters, if you agree I will make sure to propagate this to all solvers in https://github.com/PyLops/pylops/pull/363.


pylops/utils/decorators.py
```diff
@@ -4,22 +4,22 @@
...
-        Solver type function. Its signature must be ``func(A, b, x0=None, **kwargs)``.
+        Solver type function. Its signature must be ``func(A, b, *args, **kwargs)``.
         Its output must be a result-type tuple: ``(xinv, ...)``.
     """
 
     @wraps(func)
-    def wrapper(A, b, x0=None, **kwargs):  # SciPy-type signature
-        if x0 is not None:
-            x0 = x0.ravel()
+    def wrapper(A, b, *args, **kwargs):  # SciPy-type signature
```
cako, line 18: I initially though about using this signature. What happens then the scipy solver is called like this `cg_wrapped(A, b, x0)` (x0 is args[0])? I haven't tried it but I think it wouldnt work.

mrava87: Mmh maybe, but the one we have so far can definitely not be used for a couple of solvers which require more mandatory parameters than A and y…

In theory if x0 is a kwarg and someone passes it as arg, it is not surprising that this will raise an error… but then we are basically saying we need x0 as 3rd argument all the time no matter it is passed as x0=x0 or x0? I don’t think we can do that for all solvers, then we may need multiple of these or we need to find a way to write solvers that don’t rely on this?

mrava87: Also for example scipy lsqr does not have x0 as 3rd element so I think calling A,y,x0 is not really something we should accommodate for… what do you think?

mrava87: One other option could be to make a decorator `disabled_ndarray_multiplication` which we can add to solvers like `RegularizedInversion` for the moment as they are not compatible with ndarray multiplications. So that we leave `add_ndarray_support_to_solver ` to pure solvers. What do you think?

mrava87:
Let me give you an example:

```
Rop = pylops.Restriction((100, 20), iava=np.array((5, 10, 12)), axis=1)
Dop = pylops.FirstDerivative((100, 20))

x = np.ones(100*20)
y = Rop @ x

xinv = pylops.optimization.leastsquares.NormalEquationsInversion(
    Rop, [Dop], y, epsRs=[np.sqrt(0.1)], returninfo=False, **dict(maxiter=100)
)

xinv = pylops.optimization.leastsquares.RegularizedInversion(
    Rop, [Dop], y, epsRs=[np.sqrt(0.1)], returninfo=False, **dict(maxiter=100)
)
```

With the current code (prior to this patch), neither of these two will work (because the dimensions of dimds of the data term and regularization term don't agree). With the suggested changes, if you add with `with pylops.disabled_ndarray_multiplication():`, the solvers will work. However, I don't like this as it seems very fragile and most users will have to try without and then eventually add it. 

Until we have a good solution (hopefully before releasing v2, but I am not even sure if we want to have solvers that can handle vectors (single rhs), matrices (multiple rhs), as well as non-reshaped vectors (single rhs with ndarray behavior)), I suggest we wrap the leastsquares solvers with `@disable_ndarray_multiplication` as proposed in my last commit.

What do you think?

mrava87:
I am also removing this test `test_NormalEquationsInversion_ndarray` as I think it is testing a sort of lucky edge case rather than an expected behavior of this solver.

cako:
Hi Matteo, thanks for the patch, I do agree with your final conclusions, but let me just go over my thought process when I was designing this.

The first problem I encountered was in NormalEquationsInversion which relies on cg/lsqr solvers from scipy which cannot (and will not) support NDarrays. So we have two options: reimplement everything ourselves (bad) or just wrap each call with a ravel/reshape. And instead of wrapping each of them individually, I [looked at their signatures](https://docs.scipy.org/doc/scipy/reference/sparse.linalg.html), and (erroneously) concluded all of them were `solver(A, y, x0=None, ...)`. I didn't realize that `lsqr` (and `lsmr`) do not have this signature. So yes, my initial idea was to use these to only wrap scipy solver. Indeed I considered naming the decorator `add_ndarray_support_to_scipy_solver` but this would be too long hahaha.

The problem that caused was that in NormalEquationsInversion, it was creating a flat zero-vector (of size shape[0]) which generated the bug. I then tried to make it inherit from Op.dims, but Op.dims is still n-dimensional even under disable_ndarray... the only difference is that under disable_ndarray the multiplications dont get reshaped. But you don't lose the dims property. So, I considered solving this by shaping the zeros the same shape as y, but this didn't feel like a very stable solution.

Anyways, so we don't break anything, for now I agree with your solution: disable everywhere we don't explicitly know how to do it yet.

pylops/utils/decorators.py
cako: I would personally keep `add_ndarray_support_to_solver` only for pure solvers. This way works.

cako (approved):

mrava87:
Thanks for the explanation. I think it makes sense, and to the long name, I don't mind long names when they help understanding what a method does.

I would say, for now let me go ahead. After I am done with the revamping of the solvers, we could create 2 methods, `add_ndarray_support_to_scipy_solver` and add_ndarray_support_to_leastsquares_solver` (as I will make all of them have the same function signature) and wrap those around with it :)

PR #378, opened 2022-05-01
fix: Bug: Deghosing tutorial broken after v1.17.0 #377

### LINKED ISSUE #377: Bug: Deghosing tutorial broken after v1.17.0
Unfortunately this went unnoticed for a while, luckily I was checking all tutorials after the major refactoring of our solvers and I discovered a pretty weird behavior. See:

Working: https://pylops.readthedocs.io/en/v1.16.0/tutorials/deghosting.html

Broken: https://pylops.readthedocs.io/en/v1.17.0/tutorials/deghosting.html

I think I was able to find when this problem was introduced https://github.com/PyLops/pylops/commit/7e596d4dad3793d6430204b7a9b214a9dc39616c, specifically the switch `fftshift->fftshift_after` which should have been instead `fftshift -> ifftshift_before` as I think also the documentation suggests.

@cako can you confirm you agree this is a bug. If you agree, I am going to fix it first on v1.18 and then on our dev branch :)

cako:
Closes #377

### DIFF
--- a/pylops/waveeqprocessing/oneway.py
+++ b/pylops/waveeqprocessing/oneway.py
@@ -165,9 +165,9 @@ def PhaseShift(vel, dz, nt, freq, kx, ky=None, dtype="float64"):
     else:
         dims = (nt, kx.size, ky.size)
         dimsfft = (freq.size, kx.size, ky.size)
-    Fop = FFT(dims, dir=0, nfft=nt, real=True, dtype=dtype)
+    Fop = FFT(dims, dir=0, nfft=nt, real=True, ifftshift_before=True, dtype=dtype)
     Kxop = FFT(
-        dimsfft, dir=1, nfft=kx.size, real=False, fftshift_after=True, dtype=dtypefft
+        dimsfft, dir=1, nfft=kx.size, real=False, ifftshift_before=True, dtype=dtypefft
     )
     if ky is not None:
         Kyop = FFT(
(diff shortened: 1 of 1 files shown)


mrava87:
Do we really care about fixing v1.17? I guess we could release v1.17.1 if you want?

Also, in my test I didn't seem to need the one on `Fop = FFT(dims, dir=0, nfft=nt, real=True, ifftshift_before=True, dtype=dtype)`, are you sure it is needed?

cako:
> Do we really care about fixing v1.17? I guess we could release v1.17.1 if you want?

I just pulled that branch to make sure it wasn't introduced later. The PR should go directly to v1.18

> Also, in my test I didn't seem to need the one on `Fop = FFT(dims, dir=0, nfft=nt, real=True, ifftshift_before=True, dtype=dtype)`, are you sure it is needed?

That's right, it was a mistake! Indeed the associated frequency is not shifted: https://github.com/PyLops/pylops/blob/225627a72efb02bc0e6f0496978e4ff33f0bff36/pylops/waveeqprocessing/oneway.py#L317

mrava87:
Great :) 

I made a new PR as I think we are getting issues with python 3.8 due to the change in SciPy (I think SciPy has stopped supporting 3.6 and 3.7 from that version, reason why our tests don't break for those versions)

If I make the change directly on top of the current v1.18.x, the problem does not arise as we fixed it :)

PR #384, opened 2022-05-15
Minor: Finalize restyling of optimization module

mrava87:
This PR finalizes the remaining points of discussion in https://github.com/PyLops/pylops/pull/363:

- Added args and kwargs to all methods in Solver.
- Merged class-based and function-based solvers in unique file
- Cleanup the documentation of adding solver.


cako (approved):
Beautiful

PR #398, opened 2022-06-13
Added matvec/rmatvec count

mrava87:
This PR introduces a new decorator `count` which is used to count the number of matvec/rmatvec/matmat/rmatmat performed by a single operator. These counts are stored in the following variables `matvec_count`,  `rmatvec_count`, `matmat_count`, `rmatmat_count` and can be restarted using the method `reset_count`

### DIFF
--- a/pylops/utils/decorators.py
+++ b/pylops/utils/decorators.py
@@ -123,24 +123,40 @@ def reshaped(func=None, forward=None, swapaxis=False):
     return decorator
 
 
-def count(f=None, forward=True, mat=False):
+def count(func=None, forward=None, matmat=None):
     """Decorator used to count the number of forward and adjoint performed by an operator.
 
     Parameters
     ----------
-    f : :obj:`callable`, optional
+    func : :obj:`callable`, optional
         Function to be decorated when no arguments are provided
     forward : :obj:`bool`, optional
-        Whether to count the forward (``True``) or adjoint (``False``)
-    mat : :obj:`bool`, optional
-        Whether to count the matvec (``False``) or matmat (``True``)
+        Whether to count the forward (``True``) or adjoint (``False``). If not provided, the decorated
+        function's name will be inspected to infer the mode. Any operator having a name
+        with 'rmat' as substring will be defaulted to False
+    matmat : :obj:`bool`, optional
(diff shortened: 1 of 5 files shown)


pylops/utils/decorators.py
```diff
@@ -121,3 +121,42 @@ def wrapper(self, x):
     if func is not None:
         return decorator(func)
     return decorator
+
+
+def count(f=None, forward=True, mat=False):
```
cako, line 126: Was it a design choice to not derive forward/mat based on the `f.__name__` as was done in the `reshaped` decorator? If it wasn't and you think it's a good idea, let me know and I can implement it.

pylops/utils/decorators.py
```diff
@@ -121,3 +121,42 @@ def wrapper(self, x):
...
+                    self.matvec_count -= x.shape[-1]
+                else:
+                    self.matvec_count += 1
+            else:
+                if mat:
+                    self.rmatmat_count += 1
+                    self.rmatvec_count -= x.shape[-1]
+                else:
+                    self.rmatvec_count += 1
+            y = f(self, x)
```
cako, line 155: I think it's better to put this line in the beginning of the wrapper. If for some reason `f` errors, the linear operator will have the wrong count.

cako (changes requested):

pylops/utils/decorators.py
mrava87: Agree :)

pylops/utils/decorators.py
mrava87: Mmh not really. I think it makes sense to try to be as consistent as possible between different routines in the library, so I fixed it. Have a look and let me know if you agree.

Question: shall I still be verbose in `LinearOperator` or not (I mean shall I just use @count or specify the parameters? What do you suggest? For now I don't change it, but I tried and the new code would work as well by doing the internal evaluation based on `f.__name__`

cako: Awesome, the change looks good. Regarding your question, personally I would remove all the arguments, but that's my personal preference.

cako (approved):
Fee free to merge

PR #399, opened 2022-06-13
feature: added variable shift to Shift

mrava87:
This PR modified the `Shift` operator to allow user to pass shifts that vary across all axes apart from the one the shift
is applied. An example of how to use this new feature is also provided inn the `plot_shift` example

### DIFF
--- a/pylops/signalprocessing/Shift.py
+++ b/pylops/signalprocessing/Shift.py
@@ -1,4 +1,5 @@
 import numpy as np
+from numpy.core.multiarray import normalize_axis_index
 
 from pylops.basicoperators import Diagonal
 from pylops.signalprocessing import FFT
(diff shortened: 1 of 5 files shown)


pylops/signalprocessing/Shift.py
```diff
@@ -95,8 +98,24 @@ def Shift(
...
+
+    shift = _value_or_list_like_to_array(shift)
+
+    if shift.size == 1:
+        shift = np.exp(-1j * 2 * np.pi * Fop.f * shift)
+        Sop = Diagonal(shift, dims=dimsdiag, axis=axis, dtype=Fop.cdtype)
+    else:
+        # add dimensions to shift to match dimensions of model and data
+        if axis < 0:
+            axis = len(dims) + axis
```
cako, line 110: Consider using `normalize_axis_index` found in `numpy.core.multiarray`

cako (commented):
Looks good!

pylops/signalprocessing/Shift.py
mrava87: Good point, added :)

mrava87:
Thanks for the suggestion, I always forget about `normalize_axis_index`. Feel free to merge if you don't have any further comment

cako (approved):

Issue #390, opened 2022-05-21
Solvers file organization


mrava87:
## Motivation
In https://github.com/PyLops/pylops/pull/363 and https://github.com/PyLops/pylops/pull/384, the `optimization` module has been modified to accommodate for class-based solvers. However, the last solution has both class- and function-based solvers in the same file. This creates a nasty problem in the documentation as those solvers have same name (the only difference is capital versus lower case).

## Solution
I suggest to reintroduce what used to be an intermediate solution (that was later abandoned), with a small change. Class based solvers are moved to `cXX` files whilst function-based solvers stay in `XX`. Having `c` as prefix is much easier to remember for devs than having `c` as suffix (like it was done in the intermediate solution), and makes it easy to have a nice documentation

cako:
Could we do `fun_` and `cls_` instead of `c` prefix?

mrava87:
I don’t like it, especially because having nothing for the function will avoid a breaking change that I don’t see being very beneficial. We could have cls_basic etc, if you think that helps

cako:
That sounds good!

PR #400, opened 2022-06-18
feature: split class and function solvers

mrava87:
Handles https://github.com/PyLops/pylops/issues/390.


docs/source/api/index.rst
```diff
@@ -153,29 +153,42 @@ Template
 Basic
 ~~~~~
```
cako, line 155: Would be interesting to add a "Class-based interface" here

docs/source/api/index.rst
```diff
@@ -153,29 +153,42 @@ Template
...
 
-.. currentmodule:: pylops.optimization.basic
+.. currentmodule:: pylops.optimization.cls_basic
```
cako, line 157: Would be interesting to add a "Functional interface" here, etc

cako (approved):

docs/source/api/index.rst
mrava87: Mmh not sure I understand?

PR #405, opened 2022-06-26
feat: creates dip_estimate and reorganizes examples

cako:
## Description

This PR creates a `dip_estimate` function which accompanies `slope_estimate`. The dips are estimated directly from the structural tensor in a more stable manner than would be obtained from applied atan to the slopes.

In addition, it also improves on the examples which use these functions.


cako:
Sure! Dips = angles measured in radians. Slopes = tangent of said dips. Local dip filters can be parametrized in terms of slope or dips. The issue with slopes is that they become infinite when the line becomes vertical. Depending on the filter formulation (e.g. Fomel's PWD), they can become unstable if using slope. Similarly, estimating dips is generally more stable because it uses atan2, so it can handle x = 0 (vertical line). Outside of those cases, tan of `dip_estimate` should give the same as `slope_estimate`, but of course you can't obtain a dip directly from slope (I mean, you can using `atan`, but not `atan2`).

mrava87:
Right, I see. I had done some comparisons in the past between the structure tensor and PWd and I remember getting results that were not consistent… maybe I was missing this ;)

PS: see this https://github.com/PyLops/pylops_notebooks/blob/master/developement/SeismicInterpolation-Gradients.ipynb towards the end where I do all sort of slope estimations and try to use it to guide inversion, maybe on the light of what you have added here we could get it work better

mrava87:
Anyways, I am happy for you to merge if you think this is ready :)

PR #410, opened 2022-07-20
feature: metrics

mrava87:
# Motivation
Users are usually in the position of creating small routines to compute metrics (mse, snr, psnr, ...) and wrap them into a callback. It makes sense to create a small utility submodule `pylops.utils.metric` where we can add common metrics when 
needed. 

Moreover, leveraging the new Callbacks class, a new callback is created that computes all metrics during iterations


cako (approved):
Looks awesome!

Issue #278, opened 2021-12-28
Add typing annotations


cako:
Typing annotations are quickly becoming ubiquitous in Python. They improve readability, documentation, and when paired with a modern IDE, can improve autocomplete suggestions and supply advanced warnings.

I propose that we incorporate type hinting as much as possible in a future release, at the very least for function/class signatures in user-facing code core library code.

### Definition of done

1. Type hinting in all user-facing function and class signatures. Anything which is exported should be type hinted.
2. Type hinting should use either standard library hinting (native in Python 3.8 and later, provided by the `typing-extensions` package in earlier versions) or NumPy typing when appropriate (provided under `numpy.typing` in NumPy versions 1.20 and higher). For example, a function which takes a `numpy.ndarray` should be annotated with `numpy.typing.NDArray`.

cako:
@mrava87 I put this as a To do for v2.0.0, I hope that's ok!

mrava87:
Sounds good! It is a lot of work but worth for a big release :) we can aim for it in coming months also to remove the various deprecated parameters here and there

cako:
@mrava87 could we upgrade to NumPy 1.20?

mrava87:
I guess you mean as minimal allowed numpy version? 

For v2, sounds like a good idea :)

PR #412, opened 2022-07-21
feature: typing annotation (avo, basicoperators, and utils)

mrava87:
This PR handles https://github.com/PyLops/pylops/issues/278 and provides typing annotation to all functions and classes of pylops modules avo, basicoperators, and utils.


mrava87:
@cako I noticed when adding typing annotations that the documentation becomes very ugly. Just asking in case you know a workaround before I start looking for it.

Screen Shot 2022-08-08 at 11 50 17 AM

mrava87:
Alright, I found it (`autodoc_typehints = "none"`)

cako:
Let me know if you need any further help with this

mrava87:
It should be easy for me to finish this. 

One thing I was struggling a bit is to annotate inputs/outputs that are of pylops.LinearOperator type. I found some suggestions online that you can simply put the name of my class, but mypy doesn’t seem very happy about this. Any suggestion?

mrava87:
@cako can you let me know if you have any comment/suggestion here? Otherwise I would like to merge this and proceed with the remaining modules :)

pylops/avo/avo.py
```diff
@@ -290,7 +325,11 @@ def approx_zoeppritz_pp(vp1, vs1, rho1, vp0, vs0, rho0, theta1):
     return rpp
 
 
-def akirichards(theta, vsvp, n=1):
+def akirichards(
+    theta: npt.ArrayLike,
+    vsvp: Union[float, npt.ArrayLike],
+    n: int = 1,
+) -> Union[npt.ArrayLike, npt.ArrayLike, npt.ArrayLike]:
```
cako, line 332: Should be Tuple[npt.Arraylike, ...]

pylops/avo/avo.py
```diff
@@ -360,7 +399,11 @@ def akirichards(theta, vsvp, n=1):
     return G1, G2, G3
 
 
-def fatti(theta, vsvp, n=1):
+def fatti(
+    theta: npt.ArrayLike,
+    vsvp: Union[float, npt.ArrayLike],
+    n: int = 1,
+) -> Union[npt.ArrayLike, npt.ArrayLike, npt.ArrayLike]:
```
cako, line 406: Same here

pylops/avo/avo.py
```diff
@@ -432,7 +475,11 @@ def fatti(theta, vsvp, n=1):
     return G1, G2, G3
 
 
-def ps(theta, vsvp, n=1):
+def ps(
+    theta: npt.ArrayLike,
+    vsvp: Union[float, npt.ArrayLike],
+    n: int = 1,
+) -> Union[npt.ArrayLike, npt.ArrayLike, npt.ArrayLike]:
```
cako, line 482: Here as well

pylops/avo/poststack.py
```diff
@@ -196,17 +204,17 @@ def PoststackLinearModelling(
...
+    m0: npt.ArrayLike = None,
+    explicit: bool = False,
+    simultaneous: bool = False,
+    epsI: float = None,
+    epsR: float = None,
+    dottest: bool = False,
+    epsRL1: float = None,
     **kwargs_solver
-):
+) -> Union[npt.ArrayLike, npt.ArrayLike]:
```
cako, line 217: Another Union -> Tuple

pylops/basicoperators/FirstDerivative.py
```diff
@@ -97,7 +100,11 @@ def __init__(
         self.order = order
         self._register_multiplications(self.kind, self.order)
 
-    def _register_multiplications(self, kind, order):
+    def _register_multiplications(
+        self,
+        kind: str,
+        order: int,
+    ) -> Callable:
```
cako, line 107: The return here should be None

pylops/basicoperators/SecondDerivative.py
```diff
@@ -87,7 +90,10 @@ def __init__(
         self.edge = edge
         self._register_multiplications(self.kind)
 
-    def _register_multiplications(self, kind):
+    def _register_multiplications(
+        self,
+        kind: str,
+    ) -> Callable:
```
cako, line 96: None here too

pylops/utils/decorators.py
```diff
@@ -44,7 +57,11 @@ def wrapper(A, b, *args, **kwargs):  # SciPy-type signature
     return wrapper
 
 
-def reshaped(func=None, forward=None, swapaxis=False):
+def reshaped(
+    func: Callable = None,
```
cako, line 61: Should be Optional[Callable]

pylops/utils/decorators.py
```diff
@@ -44,7 +57,11 @@ def wrapper(A, b, *args, **kwargs):  # SciPy-type signature
...
+    forward: bool = None,
```
cako, line 62: Optional[bool] etc

pylops/utils/decorators.py
```diff
@@ -123,7 +140,11 @@ def wrapper(self, x):
     return decorator
 
 
-def count(func=None, forward=None, matmat=None):
+def count(
+    func: Callable = None,
+    forward: bool = None,
+    matmat: bool = False,
```
cako, line 146: Same here

pylops/utils/estimators.py
```diff
@@ -1,22 +1,28 @@
...
 
 import numpy
+import numpy.typing as npt
 
 from pylops.utils.backend import get_module
 
 
-def _sampler_gaussian(m, batch_size, backend_module=numpy):
+def _sampler_gaussian(
+    m: float, batch_size: int, backend_module=numpy
```
cako, line 11: backend_module: ModuleType

pylops/utils/estimators.py
```diff
@@ -1,22 +1,28 @@
...
+) -> Union[float, npt.ArrayLike]:
     return backend_module.random.randn(m, batch_size)
 
 
-def _sampler_rayleigh(m, batch_size, backend_module=numpy):
+def _sampler_rayleigh(m: float, batch_size: int, backend_module=numpy) -> npt.ArrayLike:
```
cako, line 16: and here

pylops/utils/estimators.py
```diff
@@ -1,22 +1,28 @@
...
     z = backend_module.random.randn(m, batch_size)
     for i in range(batch_size):
         z[:, i] *= m / backend_module.dot(z[:, i].T, z[:, i])
     return z
 
 
-def _sampler_rademacher(m, batch_size, backend_module=numpy):
+def _sampler_rademacher(
+    m: float, batch_size: int, backend_module=numpy
```
cako, line 24: also ModuleType

pylops/utils/multiproc.py
```diff
@@ -1,7 +1,15 @@
...
+from typing import List, Tuple
 
+import numpy.typing as npt
 
-def scalability_test(Op, x, workers=[1, 2, 4], forward=True):
+
+def scalability_test(
+    Op,
+    x: npt.ArrayLike,
+    workers: List = [1, 2, 4],
```
cako, line 10: maybe List[int]?

pylops/utils/signalprocessing.py
```diff
@@ -76,113 +84,14 @@ def nonstationary_convmtx(H, n, hc=0, pad=(0, 0)):
...
-
-def slope_estimate(d, dz=1.0, dx=1.0, smooth=5, eps=0.0):
+def slope_estimate(
+    d: npt.ArrayLike,
+    dz: float = 1.0,
+    dx: float = 1.0,
+    smooth: int = 5,
+    eps: float = 0.0,
+    dips: bool = False,
+) -> Union[npt.ArrayLike, npt.ArrayLike]:
```
cako, line 94: Tuple

cako (commented):
@mrava87 thanks for the hard work, this probably took a good while!

I left a few changes through the files, but I think some general comments are in order.

* Whenever you `return a, b`, the type annotation should be `Tuple[typeof(a), typeof(b)]` as opposed to `Union[...]`. The case for union would be `return a if something else b`.
* Whenever you have a function argument `myint = None`, you should annotate it with `myint: Optional[int] = None`. You skip the `Optional` only when you supply a default non-`None` value like `myint: int = 1`.
* Many times you used `Union[Tuple, List, npt.Arraylike]`. I wonder if it's not better to use a standard library base class like `collections.abc.Sized` or `collections.abc.Collection`
* Why are there no annotations for `LinearOperator`?
* Would it be possible to run the tests through mypy? I'm not sure if this is a thing. Basically I just want to make sure that the typing is respecting how we normally call these functions.

pylops/avo/avo.py
mrava87: I kept making this mistake, I checked now for `-> U` and fixed all of them :)

mrava87:
Thanks!

> @mrava87 thanks for the hard work, this probably took a good while!
> 
> I left a few changes through the files, but I think some general comments are in order.
> 
> * Whenever you `return a, b`, the type annotation should be `Tuple[typeof(a), typeof(b)]` as opposed to `Union[...]`. The case for union would be `return a if something else b`.

Thanks, I kept messing this up. Should be fixed everywhere now

> * Whenever you have a function argument `myint = None`, you should annotate it with `myint: Optional[int] = None`. You skip the `Optional` only when you supply a default non-`None` value like `myint: int = 1`.

Oh, this one I didn't know. I tried to fix it everywhere (but I may have missed one or two ;) )

> * Many times you used `Union[Tuple, List, npt.Arraylike]`. I wonder if it's not better to use a standard library base class like `collections.abc.Sized` or `collections.abc.Collection`

Mmmh you mean `vp1: Union[List, Tuple, npt.ArrayLike],` -> `vp1: Collection` (with `from collections.abc import Collection`?
> * Why are there no annotations for `LinearOperator`?

I was struggling with this. Somewhere I read I should do `Type[LinearOperator]`: https://stackoverflow.com/questions/44664040/type-hints-with-user-defined-classes but mypy wasn't happy about it.  Let's for example take `Block.py`. If I do `ops: List[LinearOperator]` or `ops: List[Type[LinearOperator]]`, in both cases mypy gives: 
```
pylops/basicoperators/Block.py:19: error: Module "pylops.LinearOperator" is not valid as a type
Found 1 error in 1 file (checked 1 source file)
``` 
Any suggestion?
> * Would it be possible to run the tests through mypy? I'm not sure if this is a thing. Basically I just want to make sure that the typing is respecting how we normally call these functions.

Technically I think it is possible, but so far I have so many errors that I am not sure it makes sense.

pylops/utils/multiproc.py
mrava87: So far I didnt add the [] to List and Tuple. I did it now... phew, so much work ;)

pylops/utils/estimators.py
mrava87: Nice, I didnt know about this

cako:
Give me a little longer to try to come up with a reasonable solution to the LinearOperator issue. The issue stems from the clash between the file name (a "module" in the eyes of mypy) and the class. For example, LinearOperator.py within Python doubles as the "module" LinearOperator (`from . import LinearOperator` for example) but also the class (`from .LinearOperator import LinearOperator`). This happens not only with LinearOperator but with almost every operator class. The only way I've found to get rid of this problem for good is to rename the file LinearOperator.py to something else, e.g., linear_operator.py. This is actually [what is suggested by PEP8](https://softwareengineering.stackexchange.com/questions/308972/python-file-naming-convention) so I don't think it's bad design choice. For the user, this change should be transparent, so I think we should take advantage of the major version and just do it. I have started doing this for all files, but haven't finished yet.

Another issue I have been facing is the clash between multiple imports. This happens because of wildcard imports used in the __init__s. For example, if we do
```python
__all__ =  ["Convolve1D", "Smoothing1D"]
from .Convolve1D import *
from .Smoothing1D import *
```
but then within Smoothing1D.py we have a `from .Convolve1D import Convolve1D`, we now have a double import. If one enables errors in MyPy, this generates a *lot* of errors. The solution for this is to add `__all__ = ["Smoothing1D"]` in Smoothing.py (and respectively for all files). This should also have the benefit of not polluting the namespaces with spurious things, so I am also working on getting these changes implemented.

The last issue the use of `npt.ArrayLike`. I've actually found that this is a pretty poor type for what we want to do. It doesn't have indexing for example. Basically `npt.ArrayLike` only guarantees that `np.asarray(x)` won't fail. So whenever we use it, we should probably have an `ncp.asarray` converting the input. For outputs, we should avoid it, favoring something like `Union[np.ndarray, cp.ndarray]`. I've started working on this, but again, not done yet.

So give me a couple days and I'll circle back!

mrava87:
Alright I see. Sounds like a good plan :)

Let me suggest something. I am done with typing annotation of signalprocessing, let me make a PR so that we don't work on both at the same time. We could merge those two PRs and then you can work on file renaming before we go ahead with the remaining modules. Sounds good?

cako:
Sure. Let's merge these two and then I'll work on the renaming. Otherwise it's going to get very complicated!

cako (approved):

PR #414, opened 2022-07-24
feature: patchhelper

mrava87:
## Motivation
In their current implementation, `Patch` and `Sliding` operators require users to provide the number of windows to be used for the model vector. This may not be easy to define upfront and therefore a `design` parameter was introduced that infers such parameters from all the other input parameters. This is however suboptimal as it requires running the code interactively, finding the nwins and then using it.

## Solution 
This PR introduces a new method associated to each `Patch` and `Sliding` operator that can be run upfront to design the patching/slicing strategy and provides the correct nwins. The `design` parameter is also removed from the operators


cako (commented):
So as to not confuse the users, I think the design functions should probably be changed to snake_case instead of CamelCase.

mrava87:
Good point, I will do that :)

mrava87:
@cako, check you agree with this

cako (approved):

Issue #407, opened 2022-07-19
Nasty bug in FFT2D when working with cupy arrays


mrava87:
# Problem
The `pylops.signalprocessing.FFT2D` operator crashes when trying to deal with cupy arrays. This seems due to the fact that `nfft` is stored as tuple - note that this was not the case before the major refractoring of the FFT routines, reason why things used to work (too bad we can't have CI for cupy operators :( )

Basic example:
```
dt, dx = 0.005, 5
nt,nx  = 100, 201
t      = cp.arange(nt)*dt
x      = cp.arange(nx)*dx
f0     = 10
nfft   = 2**10
d      = cp.outer(cp.sin(2*np.pi*f0*t), cp.arange(nx)+1)

FFTop = FFT2D(dims=[nt,nx], nffts=[nfft,nfft], sampling=[dt,dx])
dottest(FFTop, nfft*nfft, nt*nx, complexflag=2, verb=True, backend='cupy') # crashes

FFTop.nffts = list(FFTop.nffts)
dottest(FFTop, nfft*nfft, nt*nx, complexflag=2, verb=True, backend='cupy') # works

```

Error:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-42-d1129b05654c> in <module>
      8 
      9 FFTop = FFT2D(dims=[nt,nx], nffts=[nfft,nfft], sampling=[dt,dx])
---> 10 dottest(FFTop, nfft*nfft, nt*nx, complexflag=2, verb=True, backend='cupy')
     11 FFTop.nffts = list(FFTop.nffts)
     12 

~/Documents/OpenSource/pylops/pylops/utils/dottest.py in dottest(Op, nr, nc, rtol, complexflag, raiseerror, verb, backend, atol, tol)
    108         v = v + 1j * ncp.random.randn(nr).astype(rdtype)
    109 
--> 110     y = Op.matvec(u)  # Op * u
    111     x = Op.rmatvec(v)  # Op'* v
    112 

~/Documents/OpenSource/pylops/pylops/LinearOperator.py in matvec(self, x)
    141             raise ValueError("dimension mismatch")
    142 
--> 143         y = self._matvec(x)
    144 
    145         if x.ndim == 1:

~/Documents/OpenSource/pylops/pylops/signalprocessing/FFT2D.py in _matvec(self, x)
     72         else:
     73             print(type(x))
[...]

PR #419, opened 2022-08-16
fix: forced nffts in FFT*D to be tuple for cupy

mrava87:
Handles https://github.com/PyLops/pylops/issues/407.


mrava87:
@cako  see if you agree with this small fix :)

cako (approved):

PR #426, opened 2022-08-23
Improve type annotation to minimize mypy errors

cako:
This PR is far from perfect, but I attempted to bring some sanity into mypy. Previously I was getting hundreds of thousands of errors when running mypy on the files within the `pylops` folder. The vast majority of errors stemmed from the use of `npt.ArrayLike` instead of `npt.NDArray`. While `npt.ArrayLike` is more general, it doesn't have a myriad of methods we normally required, like `.shape` and `len`.

Another issue was not being able to annotate LinearOperators, which can be circumvented by renaming some files. The general changes fall under the following categories:

* Rename files so filenames and classes do not conflict
* Add `__all__` guards to all files to reduce namespace conflict and multiple imports
* Only use `npt.ArrayLike` when it does not generate an error (no method which isn't provided is used). Otherwise use `npt.NDArray`
* Create some aliases like `IntNDArray` for `npt.NDArray[np.int_]` and `InputDimsLike` for `Union[Sequence[int], IntNDArray]`. These are all placed in `pylops.utils.typing`. This choice allows for quick refactoring in case we want to include new supported types.
* `_value_or_list_like_to_array/tuple` have been refactored and simplified to become `_value_or_sized_to_array/tuple`

Work which remains to be done:
* Fix/annotate in examples/, tutorials/, pytests/, `pylops/utils`, `pylops/waveeqprocessing`, `pylops/linearoperator.py`
* Rerun mypy on all folders to ensure that errors are minimized as much as possible. This can be done through editor automatic linting, or from the command-line with e.g., `mypy  examples/`


mrava87:
Off to SEG today, I'll try to get to this asap

cako:
> Off to SEG today, I'll try to get to this asap

Have fun!!

pylops/basicoperators/pad.py
```diff
@@ -68,25 +70,25 @@ class Pad(LinearOperator):
...
+        dtype: DTypeLike = "float64",
         name: str = "P",
     ) -> None:
         if np.any(np.array(pad) < 0):
             raise ValueError("Padding must be positive or zero")
-        dims = _value_or_list_like_to_tuple(dims)
+        dims = _value_or_sized_to_tuple(dims)
         # Accept (padbeg, padend) and [(padbeg, padend)]
-        self.pad = [pad] if len(dims) == 1 and len(pad) == 2 else pad
+        self.pad: Sequence = [pad] if len(dims) == 1 and len(pad) == 2 else pad
```
mrava87, line 82: Mmh what is the reasoning for annotating this. I think we create a lot of internal variables in init methods, but I am not sure we should go all the way to annotating all of them?

mrava87 (commented):
Great job!

I think this is good to go for me, we can follow these coding patterns to annotate the remaining modules.

I am not so sure we need to do the same for tutorials and examples, I think annotations are just for the package... we should simply exclude these files from mypy checks, what do you think?

pylops/basicoperators/__init__.py
```diff
@@ -40,38 +40,38 @@
...
+from .smoothing1d import *
+from .smoothing2d import *
+from .causalintegration import *
+from .firstderivative import *
+from .secondderivative import *
+from .laplacian import *
+from .gradient import *
+from .directionalderivative import *
 
 __all__ = [
```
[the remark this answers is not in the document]
cako: It is not! But it prevents us from inadvertently exporting functions if we add a file that doesn't have __all__. At the end of the day it's matter of choice, and since it's already there I opted to not remove it.

pylops/basicoperators/pad.py
cako: Sometimes when annotating, the parser cannot tell exactly which type the variable will be. For example in this case, `self.pad` could be a `list` but if the user passes a `tuple`, it will be that. Mypy in this case will complain that it cannot figure out which type `self.pad` is. To get rid of the complaint, I annotated it.

My general approach to this PR was: if mypy gave me an error that I could fix without changing the code, I went ahead and did it.

cako:
Awesome, I'll merge then. With respect to examples/ tutorials/, I don't mean annotating variables/functions in them, but ensuring that the types we annotated in the operators are actually respecting the function calls. 

For example, if we annotated `dims` as `tuple`, then running mypy on examples would likely flag many errors, because our operators can take dims as tuples, lists, ndarrays, etc. So instead of annotating as tuple, we should probably annotate with something else, like an `Sequence[int]` or something else (I think in the end I went with something a bit more complex). At the same time, if we are calling our operators with `dims = [100.0, 10.0]`, this is bad practice since dimensions should be integers and not floats, so it can inform us on how to improve our code quality.

I will say that we should absolutely keep mypy as a suggestion/linter and not precommit hook or anything required. At the end of the day we want users who also use annotations in their code to be able to interact with our library in a predictable manner, so that's the main motivation IMO.

pylops/basicoperators/__init__.py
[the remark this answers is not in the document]
mrava87: Alright, makes sense. Let's keep it :)

pylops/basicoperators/pad.py
mrava87: Ok, good point. Let's use this strategy. If we get an error and it is clear this is due to the parser having issues telling the type we annotate, otherwise we dont :)

mrava87:
> Awesome, I'll merge then. With respect to examples/ tutorials/, I don't mean annotating variables/functions in them, but ensuring that the types we annotated in the operators are actually respecting the function calls.
> 
> For example, if we annotated `dims` as `tuple`, then running mypy on examples would likely flag many errors, because our operators can take dims as tuples, lists, ndarrays, etc. So instead of annotating as tuple, we should probably annotate with something else, like an `Sequence[int]` or something else (I think in the end I went with something a bit more complex). At the same time, if we are calling our operators with `dims = [100.0, 10.0]`, this is bad practice since dimensions should be integers and not floats, so it can inform us on how to improve our code quality.
> 
> I will say that we should absolutely keep mypy as a suggestion/linter and not precommit hook or anything required. At the end of the day we want users who also use annotations in their code to be able to interact with our library in a predictable manner, so that's the main motivation IMO.

Alright, I agree. Mypy can help spot little inconsistencies like the float vs. int for dims. I can look into the tutorial and examples once we have annotated all remaining files :)

And yes, we dont want to have a precommit hook. I can try to add a GitHub action like for flake8 for mypy to we can ignore (but is there to remind us if we introduce errors)

PR #428, opened 2022-08-25
fix: enforce float to scalars in optimizers for cupy compatibility

mrava87:
A number of problems arose when running these solvers with cupy. They can all be reconducted to 1element arrays staying on the GPU versus scalars being automatically moved back to the CPU. 

This is now enforced throughout for values that we want to store in cost function history lists and similar.

### DIFF
--- a/pylops/optimization/cls_basic.py
+++ b/pylops/optimization/cls_basic.py
@@ -316,9 +316,11 @@ class CGLS(Solver):
         # create variables to track the residual norm and iterations
         self.cost = []
         self.cost1 = []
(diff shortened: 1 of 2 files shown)


pylops/optimization/cls_basic.py
```diff
@@ -668,7 +673,10 @@ def setup(
             self.alfa = self.ncp.linalg.norm(self.v)
             if self.alfa > 0:
                 self.v = self.v / self.alfa
-                self.w = self.v.copy()
+        else:
```
mrava87, line 676: I also catched this bug, before when the condition wasnt met

cako (approved):

PR #429, opened 2022-08-27
Feature: TorchOperator

mrava87:
# Motivation

This PR introduced `TorchOperator`, an operator designed to wrap PyLops operators into `torch.autograd.Function` to allow integration of PyLops operators into PyTorch AD graphs.

The design of the operator is directly borrowed from `pylops-gpu` (https://github.com/PyLops/pylops-gpu/blob/master/pylops_gpu/TorchOperator.py) with minor changes to allow the use of only native PyLops operators.

# Milestones
- [x] Add operator
- [x] Add tests
- [x] Add tutorial - requires creating a new `requirements-doc` file like in https://github.com/PyLops/pylops-gpu/blob/master/requirements-doc.txt


mrava87:
I found this https://github.com/patrick-kidger/torchtyping but I find it an overkill for us given the amount of torch code we have. I would say until pytorch has native support for typing annot we should stick to the way I have done it. Do you agree?

pylops/torchoperator.py
```diff
@@ -0,0 +1,128 @@
+import numpy as np
+
+from pylops import LinearOperator
+from pylops.utils import deps
+
```
cako, line 5: Could we have an `__all__` here?

cako (changes requested):
This is great! My only suggestion is to have a warning when an array is moved from CPU to GPU. It will just ensure that PyLops is not surreptitiously killing performance.

pylops/torchoperator.py
mrava87: Sure, forgot about that ;)

mrava87:
> This is great! My only suggestion is to have a warning when an array is moved from CPU to GPU. It will just ensure that PyLops is not surreptitiously killing performance.

Good point. So far I am quite sure this is never going to happen. I don't have a GPU at hand right now, but from the old TorchOperator in pylops-gpu I remember that this was not allowed. What I mean is that in our current implementation we have 2  options:
- the input is a torch tensor on CPU and the TorchOperator will move this to numpy (`x=x.cpu().detach().numpy()`) and dispatch computations to one of our operators on the CPU. 
- the input is a torch tensor on GPU and the TorchOperator will move this to cupy (` x = cp.fromDlpack(to_dlpack(x))`) and dispatch computations to one of our operators on the GPU.

In both cases TorchOperator will also do the reverse with the output of the matvec/rmatvec.

Now I see the point that some of our operators (though a small portion) do not have a GPU backend. In this case it may make sense to allow a force trigger that even if torch is operating on GPU, the array will be moved to the CPU, and in that case I agree we could raise a warning. Is this what you had in mind?

Let me try adding this, and once I am back in KAUST I will test it before we go ahead and merge this.

pylops/torchoperator.py
mrava87: Added

mrava87:
@cako what do you think about these changes?

Please do not merge yet as I want to give it a go with my work machine where I can really make sure I have a mixed gpu-cpu case working (and giving the warning) :)

cako (approved):

mrava87:
Alright, I checked this and now all 3 behaviors are allowed: CPU-CPU, GPU-GPU, CPU-GPU (pylops-torch).

See: https://github.com/PyLops/pylops_notebooks/blob/master/developement/TorchOperator.ipynb

PR #432, opened 2022-08-28
Typing annotations for LinearOperator, optimization, and waveeqprocessing

mrava87:
Continues https://github.com/PyLops/pylops/pull/412


mrava87:
@cako, I am having some issues importing `LinearOperator` in any of the cls_* optimization files. It seems to me this is still the circular import problem that I thought it may have been fixed by your changes. Do you mind taking a look?

Also, I get a bunch of mypy errors because the methods of many solvers are not consistent with the signature of their parent class `Solver`. I thought adding args and kwargs will help but it doesn't really seem to, any suggestion?

cako:
> @cako, I am having some issues importing `LinearOperator` in any of the cls_* optimization files. It seems to me this is still the circular import problem that I thought it may have been fixed by your changes. Do you mind taking a look?

Not sure if this is the best way, but one quick fix is to use the `TYPE_CHECKING` constant:

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from pylops.linearoperator import LinearOperator
...
class Solver(metaclass=ABCMeta):
    def __init__(
        self,
        Op: "LinearOperator",
        callbacks: Callbacks = None,
    ) -> None:
```
This way the import will only be used for analysis, it will never run during runtime.

> Also, I get a bunch of mypy errors because the methods of many solvers are not consistent with the signature of their parent class `Solver`. I thought adding args and kwargs will help but it doesn't really seem to, any suggestion?

As per [this answer on SO](https://stackoverflow.com/questions/51003146/python-3-6-signature-of-method-incompatible-with-super-type-class) it seems like a tough problem to solve. I tried [one of the answers](https://stackoverflow.com/a/70231676) and it did reduce the error, but it seems kind of hacky.

mrava87:
Alright, I see :)

Let me try the TYPE_CHECKING route.

For the second issue, I am not sure whether is better to keep some errors for now, or go for the hacky way. What do you suggest?

mrava87:
@cako have a look at these changes. Personally, I think the hack to solve the second problem isn't really worth the effort. To me there should be a way to have a barebone method `meth(args: Any, **kwargs: Dict[Any, Any])` that one can then subclass with anything... perhaps things will change in future, let's leave it as is for now and accept we get some errors.

pylops/linearoperator.py
```diff
@@ -83,15 +88,15 @@ class LinearOperator(spLinearOperator):
...
-        name=None,
-    ):
+        Op: spLinearOperator = None,
+        dtype: DTypeLike = None,
+        shape: ShapeLike = None,
+        dims: ShapeLike = None,
+        dimsd: ShapeLike = None,
+        clinear: bool = None,
+        explicit: bool = None,
+        name: str = None,
```
cako, line 98: These should be `Optional`

pylops/waveeqprocessing/oneway.py
```diff
@@ -14,51 +15,11 @@
...
 from pylops.utils.backend import to_cupy_conditional
 from pylops.utils.tapers import taper2d, taper3d
+from pylops.utils.typing import DTypeLike, NDArray
 
 logging.basicConfig(format="%(levelname)s: %(message)s", level=logging.WARNING)
 
 
-def _phase_shift(phiin, freq, kx, vel, dz, ky=0, adj=False):
-    """Phase shift extrapolation for single depth level in 2d/3d constant
-    velocity medium
```
cako, line 23: I didnt't fully understand where this went

pylops/waveeqprocessing/wavedecomposition.py
```diff
@@ -355,20 +365,20 @@ def PressureToVelocity(
...
-    dtype="complex128",
-    name="U",
-):
+    nt: int,
+    nr: int,
+    dt: float,
+    dr: float,
+    rho: float,
+    vel: float,
+    nffts: Optional[InputDimsLike] = (None, None),
```
cako, line 374: No need for Optional. The type here should be `Union[InputDimsLike, Tuple[None, None])`. Same for some other functions here.

pylops/waveeqprocessing/oneway.py
mrava87: I just realized that this was never used, in fact the very same code was repeated in the `_PhaseShift` init method. I don't recall how this happened, likely because at some point I was trying to write a one-way migration operator using this function but I eventually never pushed it and this function stayed there... ;)

mrava87:
Do you agree this is good to go?

We can later try to further improve the annotations but at this point I think all modules are covered and most errors in mypy are either hard to fix or they require some restructuring of our codes pattern design

cako:
Almost! Give me a couple hours, I'm working on a few bugs I found. I will push them soon!

cako (approved):

cako:
Done! If you agree, go ahead and merge!

mrava87:
Thanks, this looks great :)

PR #433, opened 2022-09-05
fix: fixed Diagonal to deal with NDarrays

mrava87:
The Diagonal operator is currently unable to deal with inputs that are not flattened. This change allows any ND array to be fed as input to the matvec/rmatvec operations.


cako (approved):

cako:
Could you add a test as well before we merge this?

mrava87:
Good point, test added :)

Issue #430, opened 2022-08-27
Homogenize imports of optional libraries


mrava87:
# Motivation
In PyLops, we have an increasing number of optional dependencies. For these dependencies 2 import patters have started to emerge:

- cupy-cusignal: via the `backend.py` module with availability check performed in `deps.py`
- others: direct import in files with try-except pattern

# Proposed solution
- Add `*_enabled` variables for all other optional dependences in `deps.py`
- Add a function `*_import` for each optional dependences in `deps.py` that parses error messages during imports of those optional dependencies. This function looks like this (eg for devito):
```
def devito_import(module):
    if devito_enabled:
        try:
            import devito
        except Exception as e:
            devito_message = f"Failed to import devito (error:{e})."
    else:
        devito_message = (
            f"Devito package not installed. In order to be able to use"
            f'the {module} module run "pip install devito".'
        )
    return devito_message
```
and use it in combination with `*_enabled` when import this library, eg:
```
if deps.devito_enabled:
    import devito

    from examples.seismic import AcquisitionGeometry, Model
    from examples.seismic.acoustic import AcousticWaveSolver
else:
    devito = None
    devito_message = deps.devito_import("twoway")
```

instead of the old way
```
try:
    import devito

    from examples.seismic import AcquisitionGeometry, Model
    from examples.seismic.acoustic import AcousticWaveSolver
except ModuleNotFoundError:
    devito = None
    devito_message = (
        "Devito package not installed. In order to be able to use"
        'the twoway module run "pip install devito".'
    )
except Exception as e:
    devito = None
    devito_message = f"Failed to import devito (error:{e})."
```

So far the only downside I see for the new version is that a library gets imported twice instead ones. Is that a big deal?

Opinions?

cako:
Nice, I think this is definitely something we should tackle. With respect to the "double import", nothing to worry about since Python does not actually import it twice. It just skips it. To actually reimport a module one has to `importlib.reload(themodule)`.

One thing, this pattern of try, import devito, otherwise devito = None triggers errors with mypy. One technique is [described here](https://adamj.eu/tech/2021/12/29/python-type-hints-optional-imports/), basically using global bools to keep track. So all in all I do agree with the design, but I would suggest skipping the `devito = None` altogether and just checking `if deps.devito_enabled` whever you want to check for it. Then one never accidentally tries to call a function of the optional module outside of a `if deps.devito_enabled` block. What do you think?

mrava87:
[...]

PR #440, opened 2022-09-24
Feature : homogenize imports of optional deps

mrava87:
See https://github.com/PyLops/pylops/issues/430 for details.


mrava87:
@cako I am yet to do it for Kirchhoff and lsm as I would like the other PR merged first.

cako (approved):

PR #442, opened 2022-10-09
fix: miscellanous of bug fixes for cupy

mrava87:
This commit fixes a number of bugs that were discovered when re-running all development-cupy notebooks with the pylopsv2 codebase.


mrava87:
@cako I leave it for you to take a look even though I acknowledge you may not  have easy access to a gpu to test this. 

In most cases I could confidently see why the codes stopped running and connect it with one of the changes that we introduced in the new codebase :)

PR #450, opened 2022-10-29
Feature: TorchOperator with Nd-array inputs

mrava87:
This PR enables TorchOperator to work directly with Nd-tensor inputs both in single and batch mode. This is achieved by using the ``Op.dims`` parameter and changing the definitions of `self.matvec` and `self.rmatvec` to be able to handle Nd-arrays

### DIFF
--- a/pylops/torchoperator.py
+++ b/pylops/torchoperator.py
@@ -2,6 +2,8 @@ __all__ = [
     "TorchOperator",
 ]
 
+from typing import Optional, Union
+
 import numpy as np
 
 from pylops import LinearOperator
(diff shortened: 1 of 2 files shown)


cako:
Why isnt `dims` obtained directly from the input operator `Op`? I might be missing something...

mrava87:
Mmh good point, I didn’t think about. Until now we just did .T and .T when defining the matmat/rmatmat with batches as we explicitly asked for the input to be 2d. I thought we should allow NDarrays with batch size as first like usual in torch and then internally move at last so I needed dims… should be just a matter of moving the super call further up… let me try!

mrava87:
@cako see now what you think

cako (approved):
Really good stuff! Thanks for making this work :)

PR #459, opened 2022-12-13
Feature: Blending

mrava87:
This PR introduces a new operator, namely `pylops.waveeqprocessing.blending.Blending` which implements the blending of seismic data (or any acoustic data acquired in similar fashion) for simultaneous shooting applications. 

This is based on @NickLuiken work and refractored to be consistent with PyLopsv2 operators' layout. It comprises of 3 private functions/classes implementing the following blending strategies:

- Continouos
- Group
- Half

and a public class which exposes all of them in the same manner (same inputs and shapes of model/data).

Moreover, this PR modifies the Deblending tutorial to use now this operator instead of an operator defined within the example.

### DIFF
--- a/docs/source/api/index.rst
+++ b/docs/source/api/index.rst
@@ -121,6 +121,7 @@ Wave-Equation processing
     PressureToVelocity
     UpDownComposition2D
     UpDownComposition3D
+    Blending
     MDC
     PhaseShift
     Kirchhoff
--- a/docs/source/api/index.rst
+++ b/docs/source/api/index.rst
@@ -121,7 +121,9 @@ Wave-Equation processing
     PressureToVelocity
     UpDownComposition2D
     UpDownComposition3D
-    Blending
+    BlendingContinuous
+    BlendingHalf
+    BlendingGroup
     MDC
     PhaseShift
     Kirchhoff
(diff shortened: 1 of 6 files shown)


mrava87:
FYI @NickLuiken, I modified the group blending to use PyLops operators like the half one instead of having matvec/rmatvec natively implemented. By simply using `BlockDiag` I get the same result and about the same speed so I think this is a better solution, easier to read and leading to less code to maintain.

For the continuous, I left the implementation as you did as I think there is no way with pure PyLops operators to strike a balance between simplicity of code and performance - this is much more performant that making a huge padding like we used to do :)

mrava87:
@cako do you have time to take a look at this?

mrava87:
I just realized I forgot to add typing annotations, let me do that before accepting the PR ;)

cako:
Hi guys,

Really great stuff. Everything looks good, but I do have one suggestion. Since the different blending modes take different arguments, from a usability perspective it may make sense to make them separate entities. You already have the classes implemented separately, the only required changes would be to scrap the main public class `Blending` and expose the other three as `BlendingContinuous`, `BlendingHalf`, `BlendingGroup` or something like that. Even though it is documented nicely in this case, I try to avoid functions/classes that take optional parameters depending on other parameters. It gets confusing and is not very scalable.

mrava87:
@cako I agree, given that there are some differences it is probably not so beneficial to have a single overarching operator.. I always try to avoid having too many operators but I see in this case it doesn't really bring any benefit.

Have a look if you agree on the changes and feel free to merge :)

cako (approved):
Awesome! Looks great

PR #461, opened 2022-12-24
Create servedoc in Makefile

cako:
This PR makes it easy for developers to view docs locally by adding `make servedoc` as an option in the `Makefile`.


mrava87:
Nice, this is very useful :)

PR #462, opened 2022-12-26
Typing annotation to lsm and kirchhoff

mrava87:
This PR adds typing annotations to the `lsm` and `kirchhoff` files. These should be the only two files where typing annotations were missing

### DIFF
--- a/pylops/waveeqprocessing/kirchhoff.py
+++ b/pylops/waveeqprocessing/kirchhoff.py
@@ -250,7 +250,7 @@ class Kirchhoff(LinearOperator):
         t: NDArray,
         srcs: NDArray,
         recs: NDArray,
-        vel: Union[float, NDArray],
+        vel: NDArray,
         wav: NDArray,
         wavcenter: int,
         y: Optional[NDArray] = None,
@@ -259,10 +259,10 @@ class Kirchhoff(LinearOperator):
         dynamic: bool = False,
         trav: Optional[NDArray] = None,
         amp: Optional[NDArray] = None,
-        aperture: Optional[Union[float, Tuple]] = None,
-        angleaperture: Union[float, Tuple] = 90,
+        aperture: Optional[Tuple[float, float]] = None,
+        angleaperture: Union[float, Tuple[float, float]] = 90.0,
         anglerefl: Optional[NDArray] = None,
-        snell: Optional[Union[float, Tuple]] = None,
+        snell: Optional[Tuple[float, float]] = None,
         engine: str = "numpy",
         dtype: DTypeLike = "float64",
         name: str = "K",
@@ -415,7 +415,7 @@ class Kirchhoff(LinearOperator):
         srcs: NDArray,
         recs: NDArray,
         y: Optional[NDArray] = None,
-    ) -> Union[
+    ) -> Tuple[
         int,
         int,
         NDArray,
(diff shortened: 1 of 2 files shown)


pylops/waveeqprocessing/kirchhoff.py
```diff
@@ -242,28 +245,28 @@ class Kirchhoff(LinearOperator):
...
+        y: Optional[NDArray] = None,
+        mode: str = "eikonal",
+        wavfilter: bool = False,
+        dynamic: bool = False,
+        trav: Optional[NDArray] = None,
+        amp: Optional[NDArray] = None,
+        aperture: Optional[Union[float, Tuple]] = None,
+        angleaperture: Union[float, Tuple] = 90,
+        anglerefl: Optional[NDArray] = None,
+        snell: Optional[Union[float, Tuple]] = None,
```
cako, line 265: If possible, identify what the `Tuple` is of. For example, a `Tuple` of two ints: `Tuple[int, int]`. Or a `Tuple` of an unknown number of floats: `Tuple[float, ...]`.

pylops/waveeqprocessing/kirchhoff.py
```diff
@@ -406,7 +409,27 @@ def __init__(
...
 
     @staticmethod
-    def _identify_geometry(z, x, srcs, recs, y=None):
+    def _identify_geometry(
+        z: NDArray,
+        x: NDArray,
+        srcs: NDArray,
+        recs: NDArray,
+        y: Optional[NDArray] = None,
+    ) -> Union[
```
cako, line 418: should be `Tuple`

pylops/waveeqprocessing/kirchhoff.py
```diff
@@ -431,7 +454,15 @@ def _identify_geometry(z, x, srcs, recs, y=None):
...
-    def _traveltime_table(z, x, srcs, recs, vel, y=None, mode="eikonal"):
+    def _traveltime_table(
+        z: NDArray,
+        x: NDArray,
+        srcs: NDArray,
+        recs: NDArray,
+        vel: NDArray,
+        y: Optional[NDArray] = None,
+        mode: str = "eikonal",
+    ) -> Union[NDArray, NDArray, NDArray, NDArray, NDArray, NDArray]:
```
cako, line 465: Should be `Tuple`

cako:
I added a couple of comments, should be mostly good to go. Since we don't have mypy in the CI/CD, try to make sure the annotations are consistent by running mypy on those files.

pylops/waveeqprocessing/kirchhoff.py
mrava87: I assume you mean the return right?

mrava87: If this is the case there are a few other cases, eg `Marchenko.apply_onepoint` where this should be fixed

pylops/waveeqprocessing/kirchhoff.py
mrava87: Same as above?

mrava87:
> I added a couple of comments, should be mostly good to go. Since we don't have mypy in the CI/CD, try to make sure the annotations are consistent by running mypy on those files.

Good point, I forgot. Will do now before final merge :)

pylops/waveeqprocessing/kirchhoff.py
cako: Those are `apply_onepoint` are fine because they can actually output different things based on the input. So a `Union` is justified. If you are always returning the same type of output, and that output is composed of many components, choose a Tuple. If you are outputting different outputs depending on the call, use a Union.

cako (approved):

PR #463, opened 2023-01-03
Feature: new NonStationaryConvolve1D and NonStationaryConvolve2D

mrava87:
This PR introduces:

- 2 new operators in the `signalprocessing` module that perform non-stationary convolution. In both cases, the filters are provided on a coarser grid and interpolated on-the-fly on the model grid (of course, for small scale problems one could also provide filters in the fine-scale grid).

- A first-of-a-kind cuda implementation based on numba-cuda that seamlessly integrates with the cupy backend. We should use this PR also to fix the standard for the inclusion of such codes (we can then follow up with implementing `Spread` and `Kirchhoff`).

- A new tutorial performing image-domain least-squares migration (iLSM): this combines the `pylops.waveeqprocessing.Kirchhoff` operator to generate the so-called PSFs and the new `pylops.signalprocessing. NonStationaryConvolve2D` as the modelling operator of iLSM.

For more detailed examples of both CPU and GPU based operator calls, I refer to:

- Nonstat filtering (CPU): https://github.com/PyLops/pylops_notebooks/blob/master/developement/NonStationaryConvolve.ipynb
- Nonstat filtering (GPU): https://github.com/PyLops/pylops_notebooks/blob/master/developement-cupy/NonStationaryConvolve.ipynb
- iLSM(CPU): https://github.com/PyLops/pylops_notebooks/blob/master/developement/LeastSquaresMigration_imagedomainmarmousi.ipynb and https://github.com/PyLops/pylops_notebooks/blob/master/developement/LeastSquaresMigration_imagedomain.ipynb
- iLSM (GPU): https://github.com/PyLops/pylops_notebooks/blob/master/developement-cupy/LeastSquaresMigration_imagedomainmarmousi.ipynb


cako:
This is really great stuff!! The image-domain LSM example is awesome!

I will approve these changes since everything (mostly) runs, but I have a few edits that I will submit in another PR since I cannot edit this code directly.

cako (approved):

mrava87:
Thanks a lot, feel free to make a new PR with edits at any time :)

PR #465, opened 2023-01-15
Improve #463

cako:
This PR fixes some of the issues in #463. Importantly, it limits the number of registers used in the cuda version of `_matvec_rmatvec`. Not doing so causes [this bug](https://stackoverflow.com/questions/68658970/why-launching-a-numba-cuda-kernel-works-with-up-to-640-threads-but-fails-with-6) on my machine.

@mrava87 please check the comments in this PR for further improvements to the code.

### DIFF
--- a/examples/plot_nonstatconvolve.py
+++ b/examples/plot_nonstatconvolve.py
@@ -21,10 +21,10 @@ plt.close("all")
 
 
 ###############################################################################
-# We will start by creating a zero signal of length :math:`nt` and we will
+# We will start by creating a zero signal of length `nt` and we will
 # place a comb of unitary spikes. We also create a non-stationary filter defined by
 # 5 equally spaced `Ricker wavelets <http://subsurfwiki.org/wiki/Ricker_wavelet>`_
-# with dominant frequencies of :math:`f = [10, 15, 20, 25, 30] Hz`.
+# with dominant frequencies of :math:`f = 10, 15, 20, 25` and :math:`30` Hz.
 nt = 601
 dt = 0.004
 t = np.arange(nt) * dt
@@ -38,8 +38,7 @@ Cop = pylops.signalprocessing.NonStationaryConvolve1D(
 )
 
 x = np.zeros(nt)
-for ix in range(64, nt - 64, 64):
-    x[ix] = 1.0
+x[64 : nt - 64 : 64] = 1.0
 
 y = Cop @ x
 
@@ -62,8 +61,8 @@ plt.xlim(0, t[-1])
 plt.tight_layout()
 
 # Spectra
-f = np.fft.rfftfreq(2**9, dt)
-Sh = np.abs(np.fft.rfft(Cop.hsinterp.T, n=2**9, axis=0))
+f = np.fft.rfftfreq(nt, dt)
+Sh = np.abs(np.fft.rfft(Cop.hsinterp.T, n=nt, axis=0))
 
 plt.figure(figsize=(10, 3))
 plt.pcolormesh(t, f, Sh, cmap="jet", vmax=5e0)
@@ -79,15 +78,15 @@ plt.tight_layout()
 # filter
 nx, nz = 601, 501
 
-wav1a, _, wav1c = ricker(t[:17], f0=12)
-wav1b = ricker(t[:17], f0=30)[0]
+wav1a, _, _ = ricker(t[:17], f0=12)
+wav1b, _, _ = ricker(t[:17], f0=30)
 wav2a = gaussian(35, 2.0)
 wav2b = gaussian(35, 4.0)
 
 wav11 = np.outer(wav1a, wav2a[np.newaxis]).T
 wav12 = np.outer(wav1b, wav2a[np.newaxis]).T
 wav21 = np.outer(wav1b, wav2b[np.newaxis]).T
-wav22 = np.outer(wav1b, wav2b[np.newaxis]).T
+wav22 = np.outer(wav1a, wav2b[np.newaxis]).T
 wavsize = wav11.shape
 
 hs = np.zeros((2, 2, *wavsize))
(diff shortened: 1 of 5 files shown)


examples/plot_nonstatconvolve.py
```diff
@@ -79,15 +78,15 @@
...
+wav1a, _, _ = ricker(t[:17], f0=12)
+wav1b, _, _ = ricker(t[:17], f0=30)
 wav2a = gaussian(35, 2.0)
 wav2b = gaussian(35, 4.0)
 
 wav11 = np.outer(wav1a, wav2a[np.newaxis]).T
 wav12 = np.outer(wav1b, wav2a[np.newaxis]).T
 wav21 = np.outer(wav1b, wav2b[np.newaxis]).T
-wav22 = np.outer(wav1b, wav2b[np.newaxis]).T
+wav22 = np.outer(wav1a, wav2b[np.newaxis]).T
```
cako, line 89: Without this change, `wav21` and `wav22` are the same.

examples/plot_nonstatconvolve.py
```diff
@@ -116,16 +115,18 @@
 )
 
 x = np.zeros((nx, nz))
-x[:, 201] = 1.0
-x[:, 401] = -1.0
+line1 = (np.arange(nx) * np.tan(np.deg2rad(25))).astype(int) + (nz - 1) // 4
+line2 = (np.arange(nx) * np.tan(np.deg2rad(-25))).astype(int) + (3 * (nz - 1)) // 4
+x[np.arange(nx), np.clip(line1, 0, nz - 1)] = 1.0
+x[np.arange(nx), np.clip(line2, 0, nz - 1)] = -1.0
```
cako, line 121: This makes the lines at an angle. It allows a better visualization of the nonstationarity of the events

examples/plot_nonstatconvolve.py
```diff
@@ -116,16 +115,18 @@
...
-y = Cop * x
+y = Cop @ x
 
 fig, axs = plt.subplots(1, 2, figsize=(10, 4))
-axs[0].imshow(x.T, cmap="gray", vmin=-0.1, vmax=0.1)
+axs[0].imshow(x.T, cmap="gray", vmin=-1.0, vmax=1.0)
 axs[0].axis("tight")
 axs[0].set_title("Input")
-axs[1].imshow(y.T, cmap="gray", vmin=-0.1, vmax=0.1)
+axs[1].imshow(y.T, cmap="gray", vmin=-3.0, vmax=3.0)
```
cako, line 129: Very clipped

pylops/signalprocessing/_nonstatconvolve2d_cuda.py
```diff
@@ -3,7 +3,7 @@
 from numba import cuda
 
 
-@cuda.jit
+@cuda.jit(max_registers=40)
```
cako, line 6: Fixes https://stackoverflow.com/questions/68658970/why-launching-a-numba-cuda-kernel-works-with-up-to-640-threads-but-fails-with-6

cako:
## Additional suggested changes

* **Refactor bilinear interpolation**
   The [CUDA bilinear interp](https://github.com/PyLops/pylops/blob/ce8a0151cfb6b36a12b5a1f5f6f9e3a5d862bf4c/pylops/signalprocessing/_nonstatconvolve2d_cuda.py#L18-L45) is the same as the [CPU one](https://github.com/PyLops/pylops/blob/ce8a0151cfb6b36a12b5a1f5f6f9e3a5d862bf4c/pylops/signalprocessing/nonstatconvolve2d.py#L175-L197). It then makes sense to wrap the same code under a same function. This can be done by coding a pure Python function and then defining a device function from it. For example:
   ```python
   # CPU
  def fun(a, b):
      return a + b

  # GPU
  dev_fun = cuda.jit(device=True)(fun)
  ```
* **Move CPU to auxiliary file**
    Since the GPU functions are in another file and taking into account the above refactor, it makes sense to move it to another file.
* **Improve thread launch parameters and iteration model**
 Currently, there is a limit to the size of the model, imposed by the number of blocks and threads in the call. There are two approaches to fixing this. The first is by making the total number of threads depend on the model. For example, if the model is 1000 x 1000, launch 1 block of 1024 x 1024. There is a problem which appears when very large models are used: the number of threads that are required exceed the number that the GPU can provide.
 [Grid stride looping](https://towardsdatascience.com/cuda-by-numba-examples-1-4-e0d06651612f) can help with this as allows arbitrarily sized models to be processed by a smaller number of threads. For example, set 1 block of 1024 x 1024 and you can process any model of any size. The issue with naive grid-stride looping is that it may not be optimal in the number of threads. For example, a 32 x 32 model always launches 1024 x 1024 threads.
   A solution to that is to combine both techniques: use an input-dependent number of blocks/threads (that are powers of 2), but set a maximum number of total threads launched. This can even be set per GPU, but is probably ok to hardcode for a reasonable number obtained from the `MAX_GRID_DIM_X` attribute (see [here](https://gist.github.com/cako/6b89a4045918f9428864003a51ab6801#file-cuda-by-numba-examples-01-13-py)).

examples/plot_nonstatconvolve.py
mrava87: Makes sense :)

examples/plot_nonstatconvolve.py
mrava87: Good idea

pylops/signalprocessing/_nonstatconvolve2d_cuda.py
mrava87: I must say I didn't face this issue but:

- I used a pretty good GPU for all of my testing (RTX 3090).
- I didn't really make any model bigger than then one at the bottom of this notebook https://github.com/PyLops/pylops_notebooks/blob/master/developement-cupy/LeastSquaresMigration_imagedomainmarmousi.ipynb

Having said that, lets keep this in and re-evaluate if it is possible to make it more dynamic by finding out properties of a GPU later on

cako: I had to dig really deep to find this fix. I am not sure what causes it. Just to leave it registered, removing the [final lines](https://github.com/cako/pylops/blob/ce8a0151cfb6b36a12b5a1f5f6f9e3a5d862bf4c/pylops/signalprocessing/_nonstatconvolve2d_cuda.py#L79-L82) makes the code run. I tried to manually remove the number of registers but it did not work. The only fix I found was the one above. We should definitely revisit this as I don't fully understand how max_registers imposes this limit.

Issue #467, opened 2023-01-17
Documentation broken after https://github.com/PyLops/pylops/pull/465


mrava87:
After the PR https://github.com/PyLops/pylops/pull/465 was merged the documentation stopped building https://readthedocs.org/projects/pylops/builds/19188920/  (I get the same error when building the doc locally)

I seem to have identified the issue in 

```
@@ -159,9 +159,9 @@ class NonStationaryConvolve2D(LinearOperator):
     def _register_multiplications(self, engine: str) -> None:
         if engine == "numba":
             numba_opts = dict(nopython=True, fastmath=True, nogil=True, parallel=True)
-            self._mvrmv = staticmethod(jit(**numba_opts)(self._matvec_rmatvec))
+            self._mvrmv = jit(**numba_opts)(self._matvec_rmatvec)
         elif engine == "cuda":
-            self._mvrmv = staticmethod(_matvec_rmatvec_cuda_call)
+            self._mvrmv = _matvec_rmatvec_cuda_call
         else:
             self._mvrmv = self._matvec_rmatvec

```

but before I change it and make a PR I would like to hear the rationale for this change and if locally you experienced the same issue. Note that we have already

```
@staticmethod
    def _matvec_rmatvec(
```

so I feel adding `staticmethod` inside the `_register_multiplications` is maybe redundant?

cako:
I don't mind reverting it, but as it was before only the numpy version of `self._matvec_rmatvec` (the one which is defined in the body of the class and has the decorator) is a staticmethod. The other ones are defined outside the class and are not decorated. Hence the change!

Not sure what is causing the bug, but it would be good to investigate a bit further.

mrava87:
Alright, let me dig a bit deeper and see if I understand why adding the staticmethod makes it problematic.

My idea was to add static to the internal one but I don’t really see what is the benefit of adding it also to the other ones… as you say they are external methods. Can you clarify your thought process over this addition?

mrava87:
So far I see you can make a method static outside of a class https://www.digitalocean.com/community/tutorials/python-static-method but I don’t seem to find any link where people make a function declared outside of a class static inside the class…

cako:
We can always just dummy call the correct method:

```python
def __init__(...):
   self._register_matvec(engine)  # attaches self._registered_matvec to the correct function

@staticmethod
def _matvec_rmatvec(self, x):
    return self._registered_matvec(x)
```

cako:
[...]

PR #468, opened 2023-01-17
bug: fixed https://github.com/PyLops/pylops/issues/467

mrava87:
This commit reverts the _register_multiplications method to its original version to bring the documentation back to life. We can assess later how to remake those methods static.


mrava87:
@cako let's bring this back to ensure we can build the documentation with the latest codebase, we can then reintroduce those changes later if needed

Issue #466, opened 2023-01-17
Improvements to NonStationaryConvolve2D


mrava87:
This issue collects a few improvements for the NonStationaryConvolve2D as raised in https://github.com/PyLops/pylops/pull/465 (and a couple of others):

* **Refactor bilinear interpolation**
   The [CUDA bilinear interp](https://github.com/PyLops/pylops/blob/ce8a0151cfb6b36a12b5a1f5f6f9e3a5d862bf4c/pylops/signalprocessing/_nonstatconvolve2d_cuda.py#L18-L45) is the same as the [CPU one](https://github.com/PyLops/pylops/blob/ce8a0151cfb6b36a12b5a1f5f6f9e3a5d862bf4c/pylops/signalprocessing/nonstatconvolve2d.py#L175-L197). It then makes sense to wrap the same code under a same function. This can be done by coding a pure Python function and then defining a device function from it. For example:
   ```python
   # CPU
  def fun(a, b):
      return a + b

  # GPU
  dev_fun = cuda.jit(device=True)(fun)
  ```
* **Move CPU to auxiliary file**
    Since the GPU functions are in another file and taking into account the above refactor, it makes sense to move it to another file.

* **Improve thread launch parameters and iteration model**
 Currently, there is a limit to the size of the model, imposed by the number of blocks and threads in the call. There are two approaches to fixing this. The first is by making the total number of threads depend on the model. For example, if the model is 1000 x 1000, launch 1 block of 1024 x 1024. There is a problem which appears when very large models are used: the number of threads that are required exceed the number that the GPU can provide.
 [Grid stride looping](https://towardsdatascience.com/cuda-by-numba-examples-1-4-e0d06651612f) can help with this as allows arbitrarily sized models to be processed by a smaller number of threads. For example, set 1 block of 1024 x 1024 and you can process any model of any size. The issue with naive grid-stride looping is that it may not be optimal in the number of threads. For example, a 32 x 32 model always launches 1024 x 1024 threads.
   A solution to that is to combine both techniques: use an input-dependent number of blocks/threads (that are powers of 2), but set a maximum number of total threads launched. This can even be set per GPU, but is probably ok to hardcode for a reasonable number obtained from the `MAX_GRID_DIM_X` attribute (see [here](https://gist.github.com/cako/6b89a4045918f9428864003a51ab6801#file-cuda-by-numba-examples-01-13-py)).

* **Create operator that takes filters as input**
[...]

PR #478, opened 2023-01-29
NonStationaryFilters1D and NonStationaryFilters2D operators

mrava87:
This PR introduces two new operators as described in https://github.com/PyLops/pylops/issues/466 (point 4).

For a more detailed example usage see https://github.com/PyLops/pylops_notebooks/blob/master/developement/NonStationaryConvolveFilters.ipynb


mrava87:
@cako i decided to tackle this first as initially I thought it would just mean swapping inputs and filters… didn’t realize that the adjoint is completely different as now the model lives in the filter space. Never mind, the pure python (with numba) code is ready for review. I prefer to wait to add the cuda version after I tackle point 3 in your review

pylops/signalprocessing/nonstatconvolve2d.py
```diff
@@ -292,3 +295,347 @@ def _rmatvec(self, x: NDArray) -> NDArray:
...
+                self.dims[1] + num_threads_per_blocks_z - 1
+            ) // num_threads_per_blocks_z
+            self.kwargs_cuda["num_blocks"] = (num_blocks_x, num_blocks_z)
+        self._register_multiplications(engine)
+
+    def _register_multiplications(self, engine: str) -> None:
+        if engine == "numba":
+            numba_opts = dict(nopython=True, fastmath=True, nogil=True, parallel=True)
+            self._mv = jit(**numba_opts)(self.__matvec)
+            # Currently a race condition seem to occur when updating parts of hs multiple times within
```
mrava87, line 407: @cako if you have time could have take a look into this… initially I though I wrote the adjoint wrong, but once I switched to pure python everything worked fine. I am not sure I understand how this differ from a similar code pattern in the NonStatConvolve2D operator (the only maybe important difference is that here I need to do 4 updates of h instead of just one of x

cako: Right, so it appears you have a line like this: `hs[f(ix), g(iz)] += something`. This will likely create a race condition which you noticed. One solution would be an atomic CPU op (which [doesn't exist so far in Numba](https://github.com/numba/numba/issues/2988)). Otherwise you can create separate vectors for each ix, and then stack them all non-parallel at the end of the parallel for. So you'd have something like this:
```python
tmphs = zeros((nx, *hs.shape))
for ix in prange:
    for iz in range:
        tmphs[ix, f(ix), g(iz)] += something
return tmphs.sum(axis=0)
```

cako:
Looks good! See my comment on the race condition. Up to you if you want to change it in this PR.

cako (approved):

pylops/signalprocessing/nonstatconvolve2d.py
mrava87: Mmh right, I see your point and I think this agrees with what I have read in the numba link I put in the comment about race condition when updating slices of the vector.

However I am then confused why this one works then, https://github.com/PyLops/pylops/blob/92191c82fb56b55e9fba5425fd003e941616e175/pylops/signalprocessing/nonstatconvolve2d.py#L248 ???

I can implement your suggestion, for 3d I am worried it may be very memory consuming but for 2d I think it’s doable

mrava87:
@cako let me know if you agree with these changes and I merge :)

Issue #369, opened 2022-04-16
Feature: Add ABC abstractmethod to LinearOperator


mrava87:
# Motivation
Following the discussion in https://github.com/PyLops/pylops/pull/363 we should consider also doing the same for linear operator

# Solution
Decorate `_matvect` and `_rmatvect` with `@abstractmethod`

mrava87:
Currently, adding `@abstractmethod` to these methods results in the following error:

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [5], in <cell line: 3>()
      1 d = np.random.normal(0., 1., (8, 5))
      2 Dop = pylops.MatrixMult(d)
----> 3 Dop1 = Dop.H @ Dop
      4 Dop.explicit = False
      6 v0 = np.random.normal(0,1, 5)

File /opt/anaconda3/envs/pylops/lib/python3.9/site-packages/scipy/sparse/linalg/_interface.py:484, in LinearOperator.adjoint(self)
    470 def adjoint(self):
    471     """Hermitian adjoint.
    472 
    473     Returns the Hermitian adjoint of self, aka the Hermitian
   (...)
    482         Hermitian adjoint of self.
    483     """
--> 484     return self._adjoint()

File ~/Desktop/KAUST/OpenSource/pylops/pylops/LinearOperator.py:345, in LinearOperator._adjoint(self)
    344 def _adjoint(self):
--> 345     Op = aslinearoperator(super()._adjoint())
    346     self._copy_attributes(Op, exclude=["dims", "dimsd", "name"])
    347     Op.dims = self.dimsd

File ~/Desktop/KAUST/OpenSource/pylops/pylops/LinearOperator.py:1286, in aslinearoperator(Op)
   1284     return Op
   1285 else:
-> 1286     return LinearOperator(Op)

File /opt/anaconda3/envs/pylops/lib/python3.9/site-packages/scipy/sparse/linalg/_interface.py:147, in LinearOperator.__new__(cls, *args, **kwargs)
    145     return super(LinearOperator, cls).__new__(_CustomLinearOperator)
    146 else:
--> 147     obj = super(LinearOperator, cls).__new__(cls)
    149     if (type(obj)._matvec == LinearOperator._matvec
    150             and type(obj)._matmat == LinearOperator._matmat):
    151         warnings.warn("LinearOperator subclass should implement"
    152                       " at least one of _matvec and _matmat.",
[...]

Issue #328, opened 2022-02-13
Alter LinearOperator so as to become independent of scipy


cako:
Motivation
--------------
When PyLops was idealized, it relied heavily on the widely available SciPy implementation of a LinearOperator. This was great in the beginning as it allowed for rapid development and integration into the NumPy/SciPy ecosystem. As PyLops grew, it has implemented a richer environment for LinearOperators. At the same time, we are now finding that, since LinearOperator's internals are subject to change from SciPy, it may be advantageous to offer an implementation fully independent from SciPy.

Definition of solved
--------------------------
* Implement *all* functionality that [SciPy's LinearOperator](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.LinearOperator.html) already offers directly into PyLops' LinearOperator.
* Remove any unecessary usage of  `scipy.sparse.linalg.LinearOperator`
* Ensure that all tests pass ideally without any modification

cako:
Closed by #482

PR #479, opened 2023-02-01
Added Abstraction to `_matvec` and `_rmatvec`

rohanbabbar04:
PR for issue #369 
- Created a New abstract Class `BaseLinearOperator` which extends the LinearOperator with abstract methods `_matvec` and `_rmatvec`. 
- Made all the classes extend the new `BaseLinearOperator`
-  In `docs/source/conf.py` changes were made because of flake8


dikwickley:
@rohanbabbar04 i don't think you should be changing the `makefile`. Also, a lot of your updates are just indentations. maybe try turning off your formatter?

rohanbabbar04:
> @rohanbabbar04 i don't think you should be changing the `makefile`. Also, a lot of your updates are just indentations. maybe try turning off your formatter?

I have reverted the MakeFile, I changed it before but its the same

rohanbabbar04:
@mrava87 do take a look..

mrava87:
thanks a lot for the PR!

In general I think you changed a lot of files, some do not have anything to do with the goal of the PR (eg `conf.py` file for the documentation), so we may ask you to do a clean PR or we can work on this and squash things at the end :)

@cako and I will make a review and get back to you.

rohanbabbar04:
> thanks a lot for the PR!
> 
> In general I think you changed a lot of files, some do not have anything to do with the goal of the PR (eg `conf.py` file for the documentation), so we may ask you to do a clean PR or we can work on this and squash things at the end :)
> 
> @cako and I will make a review and get back to you.

I have removed the `conf.py` formatting changes, maybe it was because of my formatter. Rest all files are related to the PR

cako:
Hi! Awesome that this is looked into, hopefully we can get somewhere we are all comfortable with. With that said, I think users having to derive from a new class is not great. It would require releasing a new version for example. One possibility is to have LinearOperator be a metaclass, or at least set _rmatvec and _matvec as abstract methods in it? I'm not sure that's feasible and would like to hear your opinions on it.

mrava87:
This is what I hoped/tried before but I gave the error in the GitHub issue mentioned above - see for error message.

My understanding at that time was that since LinearOperator has direct dependency on the scipy LinearOperator that

cako:
One suggestion might be to explore a [mixin structure](https://www.pythontutorial.net/python-oop/python-mixin/) where BaseLinearOperator only provides some methods.

mrava87:
Does this still use the ABCMeta concept? In other words, can you have methods that are mandatory and will raise error at class instantiation like for the ABC approach?

rohanbabbar04:
> One suggestion might be to explore a [mixin structure](https://www.pythontutorial.net/python-oop/python-mixin/) where BaseLinearOperator only provides some methods.

I think this will do away with the `ABCMeta` approach and like @mrava87 mentioned some classes have a direct dependency on LinearOperator for functions like `_adjoint`.

mrava87:
Perhaps we should explore what would it take to reach this goal https://github.com/PyLops/pylops/issues/328 first. It will ideally allow us to make LinearOperator the meta class directly instead of relying on 3 base classes :)

marcusvaltonen:
> 

I haven't had so much time to look at this, but considering the current approach with the usage of `aslinearoperator()`, I do think the cleanest approach is to get rid of the dependency on `spLinearOperator`. I would have to dig deeper to see if you could do something about it without sacrificing this.

mrava87:
@rohanbabbar04 probably it is a good time now to go back to this given our new definition of LinearOperator.. it should be much easier now to accomplish this task.

Peraphs make a brand new PR as most of this code is now obsolete :)

rohanbabbar04:
Ok @mrava87 I'll make a new PR

PR #489, opened 2023-02-19
Abstraction to `_matvec` and `_rmatvec`

rohanbabbar04:
For issue #369 

- Added an abstract `BaseLinearOperator` which extends the `LinearOperator` with abstract methods `_matvec` and `_rmatvec`
- Made all the classes extend the new `BaseLinearOperator`


mrava87:
@rohanbabbar04 maybe you didnt understand my suggestion in the previous PR. Since we have new linear operator class, there is no point to create a `BaseLinearOperator`, everything should be possible directly with `LinearOperator`

rohanbabbar04:
@mrava87 I don't think we can directly make the `LinearOperator` the abstract class

```
        """
        if isinstance(x, (LinearOperator, spLinearOperator)):
            # cast x to pylops linear operator if not already (this is done
            # to allow mixing pylops and scipy operators)
            Opx = aslinearoperator(x)
>           Op = LinearOperator(Op=_ProductLinearOperator(self, Opx))
E           TypeError: Can't instantiate abstract class LinearOperator with abstract methods _matvec, _rmatvec

..\venv\lib\site-packages\pylops\linearoperator.py:574: TypeError
=============================================== short test summary info ===============================================
FAILED test_blending.py::test_Blending_continuous[par0] - TypeError: Can't instantiate abstract class LinearOperator with abstract methods _matvec, _rmatvec
FAILED test_blending.py::test_Blending_group[par0] - TypeError: Can't instantiate abstract class LinearOperator with abstract methods _matvec, _rmatvec
FAILED test_blending.py::test_Blending_half[par0] - TypeError: Can't instantiate abstract class LinearOperator with abstract methods _matvec, _rmatvec
================================================== 3 failed in 7.24s ==================================================
```

rohanbabbar04:
We have instantiated the `LinearOperator` class so I think the best possible solution is to make a `BaseLinearOperator`

mrava87:
What if instead of `Op = LinearOperator(Op=_ProductLinearOperator(self, Opx))` you simply had `Op = _ProductLinearOperator(self, Opx)` like for the `_SumLinearOperator ` case?

rohanbabbar04:
Got a new error

```

par = {'dt': 0.3, 'dtype': 'complex128', 'imag': 1j, 'nt': 21, ...}

    @pytest.mark.parametrize(
        "par", [(par1), (par2), (par3), (par4), (par1j), (par2j), (par3j), (par4j)]
    )
    def test_CausalIntegration2d(par):
        """Dot-test and inversion for CausalIntegration operator for 2d signals"""
        dt = 0.2 * par["dt"]  # need lower frequency in sinusoids for stability
        t = np.arange(par["nt"]) * dt
        x = np.outer(np.sin(t), np.ones(par["nx"])) + par["imag"] * np.outer(
            np.sin(t), np.ones(par["nx"])
        )

        for kind, rf in itertools.product(("full", "half", "trapezoidal"), (False, True)):
            Cop = CausalIntegration(
                (par["nt"], par["nx"]),
                sampling=dt,
                axis=0,
                kind=kind,
                removefirst=rf,
                dtype=par["dtype"],
            )
            rf1 = 1 if rf else 0
            assert dottest(
                Cop,
                (par["nt"] - rf1) * par["nx"],
                par["nt"] * par["nx"],
                complexflag=0 if par["imag"] == 0 else 3,
            )

            # test analytical integration and derivative inversion only for
            # cases where a zero c is required
            if kind != "full" and not rf:
                # numerical integration
                y = Cop * x.ravel()
                y = y.reshape(par["nt"], par["nx"])

                # analytical integration
                yana = (
                    np.outer(-np.cos(t), np.ones(par["nx"]))
                    + np.cos(t[0])
                    + par["imag"]
                    * (np.outer(-np.cos(t), np.ones(par["nx"])) + np.cos(t[0]))
                )
                yana = yana.reshape(par["nt"], par["nx"])

                assert_array_almost_equal(y, yana, decimal=2)

                # numerical derivative
>               Dop = FirstDerivative(
                    (par["nt"], par["nx"]), axis=0, sampling=dt, dtype=par["dtype"]
                )
E               TypeError: Can't instantiate abstract class FirstDerivative with abstract methods _matvec, _rmatvec

test_causalintegration.py:165: TypeError
```

rohanbabbar04:
The `FirstDerivative`, `TorchOperator`, and `SecondDerivative` don't have `_matvec` and `_rmatvec` methods

mrava87:
Ok, we need to investigate this deeper... the approach you suggest introduces too many changes.

The worst case scenario (aka we can't make LinearOperator be ABC class), then we should swap the role of `BaseLinearOperator` and `LinearOperator` so that we have very few changes to the codebase for the same outcome

mrava87:
> The `FirstDerivative`, `TorchOperator`, and `SecondDerivative` don't have `_matvec` and `_rmatvec` methods

Oh I see why... for `FirstDerivative` and `SecondDerivative` it is because we use `_register_multiplications`, this can change if needed (though we create methods inside - maybe something ABC doesn't like).

For `TorchOperator`, I think we should reconsider it and simply not subclass it... but I need to look into it.

mrava87:
So, excluding those 3, the approach I suggested would work?

rohanbabbar04:
No here as well `cls_sparsity.py`

```
 # prepare decay (if not passed)
        if perc is None and decay is None:
            self.decay = self.ncp.ones(niter)

        # step size
        if alpha is not None:
            self.alpha = alpha
        elif not hasattr(self, "alpha"):
            if not isinstance(self.Op, LinearOperator):
>              self.Op = LinearOperator(self.Op, explicit=False)
            # compute largest eigenvalues of Op^H * Op
>            Op1 = LinearOperator(self.Op.H * self.Op, explicit=False)
            if get_module_name(self.ncp) == "numpy":
                maxeig: float = np.abs(
                    Op1.eigs(
                        neigs=1,
                        symmetric=True,
                        **self.eigsdict,
                    )[0]
                )
```

mrava87:
Ok, then I need to look into it myself.. this one I dont understand because LinearOperator as such has those methods...

mrava87:
As I said, I am not in favour of making a BaseLinearOperator class... I think this is becoming too messy

Try to see if we can avoid it and what would that entails - if its just changing the 3 methods you mentioned that may be a easier approach :)

rohanbabbar04:
Ok I'll look into this and get back to you :)

rohanbabbar04:
@mrava87 I have updated the code according to the above requirements and also updated the description of this PR.
I also made some changes based on what I was implementing.
Do take a look and give your opinion :)

mrava87:
Great! This looks much better at first sight, I'll get onto it soon and give you a more detailed feedback :)

pylops/basicoperators/firstderivative.py
```diff
@@ -100,34 +100,39 @@ def __init__(
         self.kind = kind
         self.edge = edge
         self.order = order
-        self._register_multiplications(self.kind, self.order)
 
-    def _register_multiplications(
```
mrava87, line 105: The reason we add this is because you are better of finding out once upfront which method to use instead of going through if...else statements every time the operator forward and adjoint methods are called... I have reinstated this :)

mrava87 (commented):
Thanks for updating the PR. 

Whilst I think this is much better than before I still found a major flaw. We want to make sure non-pylops operators which are however following the same structure of pylops operators (eg spicy operators) can still be mixed with our native operator - a simple use case is a 3rd party library that implements spicy operators, this way we can just easily use them... 

Your change in `aslinearoperator` prevent this from being possible... and in fact you had to change `test_scipyop` to make it still pass (see my detailed comment.

Having read that `Abstract base class can't be instantiated.` (see https://blog.teclado.com/python-abc-abstract-base-classes/), I think we need to have a 2 level class definition for `LinearOperator` - one private that is just a template class and then our usual one - see updated code.

mrava87:
@rohanbabbar04 I went through your code and gave you detailed comments above. I still need to look at 'torchoperator' and then do a final check that everything works as expected :)

pytests/test_linearoperator.py
```diff
@@ -72,7 +71,7 @@ def test_scaled(par):
 def test_scipyop(par):
     """Verify interaction between pylops and scipy Linear operators"""
 
-    class spDiag(spLinearOperator):
+    class spDiag(LinearOperator):
```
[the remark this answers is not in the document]
rohanbabbar04: I also didn't want to change this, I just made it to pass the tests, I would have discussed the same after you had reviewed the PR.
Now it's fine that we use a `_LinearOperator` class as a metaclass.

pylops/linearoperator.py
```diff
@@ -1479,5 +1482,3 @@ def aslinearoperator(Op: Union[spLinearOperator, LinearOperator]) -> LinearOpera
     """
     if isinstance(Op, LinearOperator):
         return Op
-    else:
```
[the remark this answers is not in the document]
rohanbabbar04: I thought if we were making the `LinearOperator` the abstract class we would not need the whole `aslinearoperator` method.
So wanted to confirm that after reviewing the PR, but now `_LinearOperator` seems fine

pylops/basicoperators/firstderivative.py
rohanbabbar04: Ok understood, yeah it's better instead of putting the if...else statements in `_matvec` and `_rmatvec` and getting it called again and again

pylops/signalprocessing/patch2d.py
```diff
@@ -271,5 +271,7 @@ def Patch2D(
         int(dims[0] // nwins0),
         int(dims[1] // nwins1),
     ), dimsd
+    OOp.dims = Pop.dims
```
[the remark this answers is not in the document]
rohanbabbar04: Earlier (combining0 * combining1 * OOp) would have been a LinearOperator but when we change `LinearOperator(Op=_ProductLinearOperator(self, Opx))` to `_ProductLinearOperator(self, Opx)`, it takes into consideration the dims of A and B and we run into an error in the `plot_patching` and `plot_sliding` 

`radon = Slid.H * data`
This command over here now that Slid would become a `_ProductLinearOperator`, its adjoint gave me those errors so I fixed it using there in sliding and patching files.

But Now as we are not making `Linear Operator ` the meta class we can write LinearOperator to instantiate it.
See errror in docs for [abf3f54](https://github.com/PyLops/pylops/pull/489/commits/abf3f54295ca1d400f31a481c1df0a016cea1752)

rohanbabbar04:
Thank you for reviewing my PR @mrava87, I have updated the torchoperator.py after taking a look at the updates in the `FirstDerivative` and `SecondDerivative` classes.
The docs currently fail as explained over here. [docs failed](https://github.com/PyLops/pylops/pull/489#discussion_r1117864589)

pylops/signalprocessing/patch2d.py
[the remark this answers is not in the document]
mrava87: Alright, let me take a look. What you did looks like a quick fix, I prefer going deep into the root of the problem and understand if it can be avoided within the LinearOperator class… I worry the same problem may now arise for other operators in users codes, so I’m not happy to have a quick fix unless it’s clear why it’s needed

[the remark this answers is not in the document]
rohanbabbar04: The other fix which seems to be correct is that we change the `_ProductLinearOperator(self, Opx)` to `LinearOperator(Op=_ProductLinearOperator(self, Opx))`

pytests/test_linearoperator.py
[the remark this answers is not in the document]
mrava87: Ok. For the future, it is best to leave a test fail and explain why than do a quick fix that it may have gone unnoticed

pylops/torchoperator.py
```diff
@@ -2,12 +2,12 @@
     "TorchOperator",
 ]
```
mrava87, line 3: Actually this is a very special operator, probably the most special of all in PyLops. It is used to bring a PyLops operator into a torch AD chain, so technically speaking I am not really sure it makes any sense for it inherit from PyLops `LinearOperator`. 

Looking at what you did I started to think maybe there is a point if we could chain multiple TorchOperator in the same way we do with PyLops but this doesn't really work because then the new operator has no apply method, which is what we care here...

In short, I believe the most reasonable thing here is to simply remove the inheritance and make it a stand-alone class.

pytests/test_linearoperator.py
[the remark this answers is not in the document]
rohanbabbar04:
```
def handle_dims_dimsd(self, A: LinearOperator, B: LinearOperator):
       if self.dims != B.dims:
            B.dims = self.dims
       if self.dimsd != A.dimsd:
            A.dimsd = self.dimsd
```
A method like this could also fix our problem we can write a method that checks the dims and dimsd of `_ProductLinearOperator` and updates the dims,dimsd of A and B

pylops/torchoperator.py
rohanbabbar04: Yeah agreed I'll just make the changes and commit..

pylops/signalprocessing/patch2d.py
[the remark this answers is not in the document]
mrava87: I see the issue, some operators like `BlockDiag` do not propagate dims/dimsd as these don't make sense for such operators.

I think in this case we should just wrap `(combining0 * combining1 * OOp)` into a LinearOperator and assign dims/dimsd directly. I find this better than re-instating `LinearOperator` into `_ProductLinearOperator` as in most cases this is not needed. So now if we have `Op= Aop * Bop * Cop` we simply have two `_ProductLinearOperator`, whilst before (and if we bring back `LinearOperator` we will have twice wrap it into an extra `LinearOperator`

pylops/torchoperator.py
mrava87: Done already

rohanbabbar04:
We can create a method that checks for the dims, dimsd of _ProductLinearOperator and updates the dims and dimsd of A and B, I think this should work

something like this...
```
def handle_dims_dimsd(self, A: LinearOperator, B: LinearOperator):
       if self.dims != B.dims:
            B.dims = self.dims
       if self.dimsd != A.dimsd:
            A.dimsd = self.dimsd
```

pylops/signalprocessing/patch2d.py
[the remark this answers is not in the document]
rohanbabbar04: Agreed this solves my query..

mrava87:
@cako this is ready for you to review. It was a bit bumpier that I have initially foreseen after making `LinearOperator` independent of scipy but I think we managed to find the least intrusive solution that still gives us the benefit of having ABC and abstract methods.

mrava87:
@cako let's try to get this one in before https://github.com/PyLops/pylops/pull/496

pylops/torchoperator.py
cako: Since the purpose of this class is to provide a PyTorch function, shouldn't it inherit from `torch.autograd.Function` like `_TorchOperator`?

pylops/signalprocessing/patch2d.py
```diff
@@ -264,7 +264,7 @@ def Patch2D(
             for win_in, win_end in zip(dwin0_ins, dwin0_ends)
         ]
     )
-    Pop = aslinearoperator(combining0 * combining1 * OOp)
+    Pop = LinearOperator(combining0 * combining1 * OOp)
```
cako, line 267: I believe `LinearOperator` can be removed here. Worked on my local tests.

pylops/signalprocessing/patch3d.py
```diff
@@ -443,7 +443,7 @@ def Patch3D(
         ]
     )
 
-    Pop = aslinearoperator(combining0 * combining1 * combining2 * OOp)
+    Pop = LinearOperator(combining0 * combining1 * combining2 * OOp)
```
cako, line 446: I believe `LinearOperator` can be removed here. Worked on my local tests.

pylops/signalprocessing/sliding1d.py
```diff
@@ -180,7 +180,7 @@ def Sliding1D(
             for win_in, win_end in zip(dwin_ins, dwin_ends)
         ]
     )
-    Sop = aslinearoperator(combining * OOp)
+    Sop = LinearOperator(combining * OOp)
```
cako, line 183: I believe `LinearOperator` can be removed here. Worked on my local tests.

pylops/signalprocessing/sliding2d.py
```diff
@@ -214,7 +214,7 @@ def Sliding2D(
             for win_in, win_end in zip(dwin_ins, dwin_ends)
         ]
     )
-    Sop = aslinearoperator(combining * OOp)
+    Sop = LinearOperator(combining * OOp)
```
cako, line 217: I believe `LinearOperator` can be removed here. Worked on my local tests.

cako (changes requested):
Very happy with this PR. The changes are clear, small and don't affect general functionality. I have a few small comments.

My only question before I approve is whether `TorchOperator` should inherit from `torch.autograd.Function`. @mrava87 any comments?

pylops/signalprocessing/patch2d.py
mrava87: Mmh did you try running the doc?

I think it works up to the point you want to pass x as a Nd-array (non-flattened). At that point it stops working as one operator (I checked and I think it’s either Block or a stack or likely both) in the chain tries to flatten. By wrapping around LinearOperator we basically enforce that at the end of the chain the array is brought back to its correct Nd-array state

mrava87:
Mmh not sure, what would the benefit be?

Eventually all the meat is in _TorchOperator, TorchOperator just stores some variables and calls the other operator method(s). I think inheriting will require to implement the forward method at least, but I need to check to be sure :)

pylops/signalprocessing/patch2d.py
cako: Got it!

cako:
Ok, sounds good! I'll leave it with you to merge then when you are done investigating the TorchOpertator inheritance

mrava87:
I did a bit of experimentation and could not find a real benefit from making TorchOperator a child of `torch.autograd.Function`. For now I prefer not to do so, we can revisit in future.

Issue #46, opened 2019-01-27
Sparse transforms: DCT, DTCW, Shearlet, Contourlets


mrava87:
# Motivation 

This Issue is intended to collect pointers to various transforms that we would like to include in PyLops (as well as discuss the level of maturity of the Python libraries providing such transforms).

## Contourlets

- https://github.com/mazayux/pycontourlet

and create tutorial based on https://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=9833541

## DCT: done 🚀 

 ~~- https://docs.scipy.org/doc/scipy/reference/generated/scipy.fftpack.dct.html~~

~~and 2D: https://stackoverflow.com/questions/7110899/how-do-i-apply-a-dct-to-an-image-in-python~~

~~See this for an example: https://inst.eecs.berkeley.edu/~ee123/sp16/Sections/JPEG_DCT_Demo.html~~

## DTCW
Make operator based on:

- https://dtcwt.readthedocs.io/en/0.12.0/

## Shearlet
Make operator for Shearlet transform using either of those libraries:

- https://github.com/stefanloock/pyshearlab and http://na.math.uni-goettingen.de/pyshearlab/

- https://github.com/grlee77/PyShearlets

- https://github.com/skmda37/ShearletX

and create tutorial based on http://citeseerx.ist.psu.edu/viewdoc/download;jsessionid=250093F1970AD5FCFCF9921F97DA9BE1?doi=10.1.1.572.1557&rep=rep1&type=pdf

dikwickley:
@mrava87 i would like to work on this issue

PR #495, opened 2023-02-23
Added `DTCWT` operator

dikwickley:
Added Dual-Tree Complex Wavelet Transform Operator. Issues: #46 

- [x] added operator
- [x] added tests
- [x] added example


pylops/signalprocessing/dtcwt.py
```diff
@@ -0,0 +1,145 @@
...
+    ------
+    NotImplementedError
+        If ``transform`` is 2 or 3.
+    ValueError
+        If ``transform`` is anything other than 1, 2 or 3.
+
+    Notes
+    -----
+    The :py:func:`dtcwt` library uses a Pyramid object to represent the transform domain signal.
+    It has
```
mrava87, line 49: this should ideally be made into a list

pylops/signalprocessing/dtcwt.py
```diff
@@ -0,0 +1,145 @@
...
+    Raises
+    ------
+    NotImplementedError
+        If ``transform`` is 2 or 3.
+    ValueError
+        If ``transform`` is anything other than 1, 2 or 3.
+
+    Notes
+    -----
+    The :py:func:`dtcwt` library uses a Pyramid object to represent the transform domain signal.
```
mrava87, line 48: transform -> transformed

pylops/signalprocessing/dtcwt.py
```diff
@@ -0,0 +1,145 @@
...
+    It has
+        `lowpass` (coarsest scale lowpass signal)
+        `highpasses` (complex subband coefficients for corresponding scales)
+        `scales` (lowpass signal for corresponding scales finest to coarsest)
+
+    To make the dtcwt forward() and inverse() functions compatible with pylops, the Pyramid object is
+    flattened out and all coeffs(highpasses and low pass signal) are appened into one array using the
```
mrava87, line 55: coeffs(highpasses and low pass signal) -> coefficents (high-pass and low pass coefficients)

pylops/signalprocessing/dtcwt.py
```diff
@@ -0,0 +1,145 @@
...
+
+    Notes
+    -----
+    The :py:func:`dtcwt` library uses a Pyramid object to represent the transform domain signal.
+    It has
+        `lowpass` (coarsest scale lowpass signal)
+        `highpasses` (complex subband coefficients for corresponding scales)
+        `scales` (lowpass signal for corresponding scales finest to coarsest)
+
+    To make the dtcwt forward() and inverse() functions compatible with pylops, the Pyramid object is
```
mrava87, line 54: appened -> appended

pylops/signalprocessing/dtcwt.py
```diff
@@ -0,0 +1,145 @@
...
+            self.second_dim = self.dims[1]
+        super().__init__(
+            dtype=np.dtype(dtype),
+            dims=self.dims,
+            dimsd=(self.coeff_array_size,),
+            name=name,
+        )
+
+    def _coeff_to_array(self, pyr: dtcwt.Pyramid) -> NDArray:
+        print("og lowpass ", pyr.lowpass)
```
mrava87, line 111: Remove prints

pylops/signalprocessing/dtcwt.py
```diff
@@ -0,0 +1,145 @@
...
+        if self.transform == 1:
+            self._transform = dtcwt.Transform1d(biort=biort, qshift=qshift)
+        elif self.transform == 2:
+            raise NotImplementedError("DTCWT is not implmented for 2D")
+        elif self.transform == 3:
+            raise NotImplementedError("DTCWT is not implmented for 3D")
+        else:
+            raise ValueError("DTCWT only supports 1D, 2D and 3D")
+
+        pyr = self._transform.forward(
```
mrava87, line 88: Ideally I would hide this away into a method (eg., `_interpret_coeffs`): from line 88 to 102.

Even better, I would see if you can avoid having a dry-run of dtcwt.Transform1d... if they dont provide any routine for that we can stick with this strategy for now.

pylops/signalprocessing/dtcwt.py
```diff
@@ -0,0 +1,145 @@
...
+        )
+        self.coeff_array_size = 0
+        self.lowpass_size = len(pyr.lowpass)
+        self.slices = []
+        for _h in pyr.highpasses:
+            self.slices.append(len(_h))
+            self.coeff_array_size += len(_h)
+        self.coeff_array_size += self.lowpass_size
+        self.second_dim = 1
+        if len(dims) > 1:
```
mrava87, line 99: What happens if dims has size of 3. From the documentation of `dtcwt.Transform1d` it seems they only allow handling 1d or 2d arrays (their doc: `X – 1D real array or 2D real array whose columns are to be transformed`)... but we should allow also 3d (or any nd) and also allow applying the transform on any axis.

This can be easily done by combining:

- adding `axis` to inputs with default to -1. Note that dtcwt applies the transform over columns so we are good :)
- `@reshaped(swapaxis=True)` see CausalIntegration for an example.. basically the axis over which to apply the transform is moved to last
- reshape x before feeding it to the transform such that we have a 2D array `otherdims * dims[axis]` (and reshape back to nd array in adjoint)

mrava87: Also since they say the transform is applied over columns, why do you have second_dim and not first_dim as the one that broadcasts?

pylops/signalprocessing/dtcwt.py
```diff
@@ -0,0 +1,145 @@
...
+        coeffs = pyr.highpasses
+        flat_coeffs = []
+        for band in coeffs:
+            for c in band:
+                flat_coeffs.append(c)
+        flat_coeffs = np.concatenate((flat_coeffs, pyr.lowpass))
+        return flat_coeffs
+
+    def _array_to_coeff(self, X: NDArray) -> dtcwt.Pyramid:
+        lowpass = np.array([x.real for x in X[-1 * self.lowpass_size :]]).reshape(
```
mrava87, line 122: why not just `-self.lowpass_size`

mrava87 (commented):
Well done, this is a very good start :)

I have a few general comments (apart from the specific ones in the code review):

- I always find better to make multiple operators for 1D, 2D, ND. Unless the library that we leverage handles that for us, you will see that we end up having a messy code full of if..else statements if we try to pack all together. Also for 1D you have the special case where the array is 1D, for 2D this doesn't make sense (similar for 3D it doesnt make sense to have 1D and 2D arrays).
-

pylops/signalprocessing/dtcwt.py
dikwickley: I made two funcs that converts the dims to 2D and back to original shape.
[f887031](https://github.com/PyLops/pylops/pull/495/commits/f887031ff0aed0edac70bed9bf1388e2dbfb4b70)

pylops/signalprocessing/dtcwt.py
dikwickley: separated it out in a function [f887031](https://github.com/PyLops/pylops/pull/495/commits/f887031ff0aed0edac70bed9bf1388e2dbfb4b70)

pylops/signalprocessing/dtcwt.py
dikwickley: fixed [f887031](https://github.com/PyLops/pylops/pull/495/commits/f887031ff0aed0edac70bed9bf1388e2dbfb4b70)

pylops/signalprocessing/dtcwt.py
dikwickley: updated [f887031](https://github.com/PyLops/pylops/pull/495/commits/f887031ff0aed0edac70bed9bf1388e2dbfb4b70)

pylops/signalprocessing/dtcwt.py
dikwickley: updated [f887031](https://github.com/PyLops/pylops/pull/495/commits/f887031ff0aed0edac70bed9bf1388e2dbfb4b70)

pylops/signalprocessing/dtcwt.py
dikwickley: done [f887031](https://github.com/PyLops/pylops/pull/495/commits/f887031ff0aed0edac70bed9bf1388e2dbfb4b70)

pylops/signalprocessing/dtcwt.py
```diff
@@ -0,0 +1,145 @@
...
+    dtype : :obj:`DTypeLike`, optional
+        Type of elements in input array.
+    name : :obj:`str`, optional
+        Name of operator (to be used by :func:`pylops.utils.describe.describe`)
+
+    Raises
+    ------
+    NotImplementedError
+        If ``transform`` is 2 or 3.
+    ValueError
```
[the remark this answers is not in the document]
dikwickley: removed [f887031](https://github.com/PyLops/pylops/pull/495/commits/f887031ff0aed0edac70bed9bf1388e2dbfb4b70)

pylops/signalprocessing/dtcwt.py
```diff
@@ -0,0 +1,145 @@
...
+    include_scale: :obj:`bool`, optional
+        Include scales in pyramid. See :py:func:`dtcwt.Pyramid`. Default is False.
+    dtype : :obj:`DTypeLike`, optional
+        Type of elements in input array.
+    name : :obj:`str`, optional
+        Name of operator (to be used by :func:`pylops.utils.describe.describe`)
+
+    Raises
+    ------
+    NotImplementedError
```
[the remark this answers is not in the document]
dikwickley: removed [f887031](https://github.com/PyLops/pylops/pull/495/commits/f887031ff0aed0edac70bed9bf1388e2dbfb4b70)

pylops/signalprocessing/dtcwt.py
```diff
@@ -0,0 +1,145 @@
...
+import numpy as np
+
+from pylops import LinearOperator
+from pylops.utils._internal import _value_or_sized_to_tuple
+from pylops.utils.decorators import reshaped
+from pylops.utils.typing import DTypeLike, InputDimsLike, NDArray
+
+
+class DTCWT(LinearOperator):
+    r"""
```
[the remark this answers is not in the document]
dikwickley: Updated [f887031](https://github.com/PyLops/pylops/pull/495/commits/f887031ff0aed0edac70bed9bf1388e2dbfb4b70)

pylops/signalprocessing/dtcwt.py
dikwickley: updated [f887031](https://github.com/PyLops/pylops/pull/495/commits/f887031ff0aed0edac70bed9bf1388e2dbfb4b70)

pylops/signalprocessing/dtcwt.py
dikwickley: updated in [f887031](https://github.com/PyLops/pylops/pull/495/commits/f887031ff0aed0edac70bed9bf1388e2dbfb4b70)

cako:
@dikwickley let me know when this is ready for review. Thanks!

mrava87:
@cako i was planning to look at it over the weekend (@dikwickley told me is ready) but feel free to go ahead if you have time

examples/plot_dtcwt.py
```diff
@@ -0,0 +1,74 @@
+"""
+Dual-Tree Complex Wavelet Transform
+=========================
+This example shows how to use the :py:class:`pylops.signalprocessing.DCT` operator.
```
cako, line 4: Should be pylops.signalprocessing.DTCWT

cako (changes requested):
The code looks solid, thanks for the feature. One think I think would be nice (but is not necessary) is the addition of a more "intuitive" example. Using random input signals is quite hard to interpret. I would prefer something like two wavelets with different frequencies, and show how they (maybe) will be separated in the DTCWT domain. Just my two cents!

I will let @mrava87 have one final look before we approve.

pylops/signalprocessing/dtcwt.py
```diff
@@ -0,0 +1,147 @@
+__all__ = ["DTCWT"]
+
+from typing import Union
+
+import dtcwt
```
cako, line 5: this needs to be wrapped in a `pylops.utils.deps` check like here:

https://github.com/PyLops/pylops/blob/dev/pylops/_torchoperator.py#LL5-LL10

pylops/signalprocessing/dtcwt.py
```diff
@@ -119,27 +115,33 @@ def _coeff_to_array(self, pyr: dtcwt.Pyramid) -> NDArray:
...
         return dtcwt.Pyramid(lowpass, highpasses)
 
     def get_pyramid(self, X: NDArray) -> dtcwt.Pyramid:
+        """Return Pyramid object from transformed array
+        """
         return self._array_to_coeff(X)
 
     @reshaped
     def _matvec(self, x: NDArray) -> NDArray:
+        x = x.swapaxes(self.axis, -1)
```
mrava87, line 137: @this is not good, you should use `swapaxes` like here https://github.com/PyLops/pylops/blob/e4d914abb7e81eca3ad46f1508c99848771bc4bb/pylops/basicoperators/firstderivative.py#L139

When something is readily available and used in many other operators we should not deviate from it unless there is a special requirement, I do not see it here :)

pylops/signalprocessing/dtcwt.py
```diff
@@ -119,27 +115,33 @@ def _coeff_to_array(self, pyr: dtcwt.Pyramid) -> NDArray:
...
             self._transform.forward(x, nlevels=self.nlevels, include_scale=False)
         )
 
     @reshaped
-    def _rmatvec(self, X: NDArray) -> NDArray:
-        return self._transform.inverse(self._array_to_coeff(X))
+    def _rmatvec(self, x: NDArray) -> NDArray:
+        Y = self._transform.inverse(self._array_to_coeff(x))
+        Y = self._2d_to_nd(Y)
+        return Y.swapaxes(self.axis, -1)
```
mrava87, line 147: Same

mrava87 (commented):
A couple of additional minor comments to avoid diverging from common practice in matvec/rmatvec

examples/plot_dtcwt.py
```diff
@@ -0,0 +1,74 @@
+"""
+Dual-Tree Complex Wavelet Transform
+=========================
```
mrava87, line 3: This needs to go all the way to the end of the title

mrava87:
@dikwickley I am reviewing your code and I found quite a few unclear points:

- If I understand from the doc dtcwt.Transform1d `Perform a n-level DTCWT decompostion on a 1D column vector X (or on the columns of a matrix X).`, so for N-D signals we need to have an array of size `-1 x self.dims[self.axis]` such that the transform is performed over the axis of interest. However your `_nd_to_2d` does `arr_nd.reshape((self.dims[0], -1))` which to me does not make sense as it puts in the last dimension the collection of all dimensions. Also `self.dims[0]` would not work correctly if you want to apply the transform over the axis=0, I think it would just work correctly as a coincidence for 2 arrays if you want to apply it over `axis=1`
- I dont understand the role of `elements = np.prod(T.shape[1:])` it seems to me that the dcwt transform already handles internally multiple dimensions. See an example:
```
import dtcwt

transform = dtcwt.Transform1d()

x = np.ones(50)
transform.forward(x).lowpass.shape

x = np.ones((50, 10))
transform.forward(x).lowpass.shape
```

which will give output (14, 1) and (14, 10). This tells me that the dimensions we need to swap is not -1 but 0 (because of how this transform works) and that we don't need to double up the extra dimensions as `elements` seem to do.



For now, I limited myself to make some cleaning of the docstring and include dtcwt in installation files where missing, but we need to make sure what you do above is correct before moving ahead :)

dikwickley:
@mrava87 
- i created `_nd_to_2d` assuming we would always need to scale down the dimensions of the array to 2 dimensions. For this a check can be added if the dimensions are greater than two, only then, reduce the input.
- `elements = np.prod(T.shape[1:])` is important, it converts array of any shape to 2D,  eg shape (m, n, o, p...) is reduced to (m, n*o*p*..). `dtcwt` does handles multiple dimensions only upto 2D https://dtcwt.readthedocs.io/en/0.12.0/reference.html#dtcwt.Transform1d (see parameter x)

mrava87:
> @mrava87
> 
> * i created `_nd_to_2d` assuming we would always need to scale down the dimensions of the array to 2 dimensions. For this a check can be added if the dimensions are greater than two, only then, reduce the input.

I see the point of `_nd_to_2d` but I am not sure about its implementation. Since you do swapping of dimensions first, you should make sure the 2d array is consistent with it, whilst doing `arr_nd.reshape((self.dims[0], -1))` would always make the first dimension of size self.dims[0] no matter if that is what is there after swapping or not. Try a simple example when you swap and dont swap and then run this function, you will see the output will always have the same size, but this is not correct :)

> * `elements = np.prod(T.shape[1:])` is important, it converts array of any shape to 2D,  eg shape (m, n, o, p...) is reduced to (m, n_o_p*..). `dtcwt` does handles multiple dimensions only upto 2D https://dtcwt.readthedocs.io/en/0.12.0/reference.html#dtcwt.Transform1d (see parameter x)

Ok, I see, but again you have already `_nd_to_2d`  so why do you need to also have `elements`. If you do `print(pyr.lowpass.shape)` inside `_interpret_coeffs` for an example with ND array (say your test test_dtcwt1D_input3D) I see that this is already a 2D-array where the second dimension is the same of that of the input (all your extra dimensions multiplied), so why is this needed `self.lowpass_size *= elements`?

mrava87:
@dikwickley i am checking if you are still interested to finalize this PR? 

Since we were close perhaps we can try to finalize it (I remember I only had a few concerns with handling of ndarrays which I didn’t agree with your code… other than that everything seemed good to me)

mrava87:
@dikwickley I created a new PR as this one is quite hold and has some conflicts. In the new PR you will see that the code is changed quite a bit, because as I mentioned above I think your code is not correct for beyond 1d arrays since it moves `axis` to the last dimension prior to computing the transform, instead this should move to the first dimension.

Let me know if you are still interested to follow that PR or not?

mrava87:
move to https://github.com/PyLops/pylops/pull/538

Issue #500, opened 2023-03-21
Incorrect reshaping if operators maps 1D to ND.


RendersJens:
Since updating to Pylops 2.0, some operators started returning nd arrays instead of raveled or 1d arrays. At first I thought this was intended, so I just kept using 1.18, but I now managed to pinpoint a situation where this happens and it is clearly not intended.

Code to reproduce:
```python
import pylops
import numpy as np

image = np.zeros((5, 5))

S = pylops.Identity(image.size)
F = pylops.signalprocessing.FFT2D(image.shape)

v = image.ravel()

# these are unexpected, they print (5, 5) but should be (25,)
print((F @ S @ v).shape)
print((F.T @ S @ v).shape)

# these all work as expected, they all print (25, )
print((F @ v).shape)
print((F.T @ v).shape)
print((S @ S @ v).shape)
print((S @ F @ v).shape)
print((F @ F @ v).shape)
print((F.T @ F @ v).shape)
print((F @ F.T @ v).shape)
print((F.T @ F.T @ v).shape)


# the problem can be avoided using parethesis, because this avoids the product operator
print((F @ (S @ v)).shape)
print((F.T @ (S @ v)).shape)
```

So the only situation in which I can reproduce it is when an FFT2D (or FFTND) appears as the left factor of a product operator, but not if the right factor is also an FFT. I will try to dig some more, but any insights are already appreciated.

I currently use Pylops 2.1.0, and this situation first appeared in 2.0. It was not present in 1.18.3.

RendersJens:
I have digged further.

The ouput is correctly raveled by .matvec() and .rmatvec(), but it is reshaped back to (5, 5) in the .dot() function:
https://github.com/PyLops/pylops/blob/dev/pylops/linearoperator.py#L616

This happens if `is_dims_shaped` is true, which is `is_dims_shaped = x.shape == self.dims`, but then it gets reshaped to the shape self.dimsd (not self.dims).

- In the case of the FFT2D operator, self.dims = (5, 5) so `is_dims_shaped` is false, so the output is not reshaped and keeps shape (25,)
- In the case of the Identity operator self.dims = (25,) so `is_dims_shaped` is true, so the output is reshaped to self.dimsd which is also (25,)

Now where it goes wrong:
For the product operator P = F @ S, we get P.dims = (25,) so `is_dims_shaped` is true, but P.dimsd = (5, 5). Therefore the output now gets reshaped to (5, 5), which is unexpected.

[...]

PR #502, opened 2023-03-27
Feature: forceflat property

mrava87:
This PR tackles issue https://github.com/PyLops/pylops/issues/500 and aims to provide an easy way for users to drive the behavior of so called ambiguous operators* via keyword arguments, therefore avoid to use global settings or context manager provided by `pylops.config.disabled_ndarray_multiplication`

* An operator is defined ambiguous is the output of either matvec or rmatvec is a 1d array (aka if `dims` or `dimsd` have len=1) . In this case pylops cannot tell whether a user would like a nd array or a 1d array when either rmatvec or matvec is performed on the 1d array from the other method.

## Background 
A couple of remarks based on the discussion in the issue, before introducing the changes:

-  `Now if that is not possible, I don't like breaking the matrix-like functionality. The operations are called .matvec and .rmatvec,`: just to be clear `matvec` and `rmatvec` have always taken 1d arrays and returned 1d arrays. This was the case in v1 and this is still the case in v2. What changed now is that if you use `dot` (or the overloads `@` and `*` some internal checks happen such that one can provide ndarrays to an operator `Op` of shape `Op.dims` (for matvec) or of shape `Op.dimsd` (for rmatvec) and obtain ndarrays of shape `Op.dimsd` (for matvec) or of shape `Op.dims` (for rmatvec); alternatively, as in pylops v1, one can provide 1darrays of size `np.prod(Op.dims)`  (for matvec) and get 1darrays of size `np.prod(Op.dimsd)` - similar for rmatvec. Apart from the ambiguous operators that we are going to tackle here in a special way, all other operators should follow this pattern, if not it means there is a bug (which it may be the case for some operators - we already identified `Identity` for example.
- All pylops solvers that call `matvec` and `rmatvec` instead of `@` and `.H @` do not experience any problem even with the ambiguous operators. We will see in a minute that some still call `@` and `.H @` and explain why

## Changes
This PR introduces the following 3 major changes:

- All pylops solvers are modified to use `matvec` and `rmatvec` instead of `@` and `.H @`. This is good as it makes them run faster (see https://pylops.readthedocs.io/en/stable/tutorials/linearoperator.html) as well as prevents issues with ambiguous operators as they were up until now. For `ISTA` and `FISTA`, we need a special treatment since they allow multiple right-hand sides (https://github.com/PyLops/pylops/pull/266.).
- A new property called `forceflat` is added to `LinearOperator`. This property is always set to `None` other than for the ambiguous operators. In that case, users can choose to set it to `True` such that the output of matvec and/or rmatvec will be always forced to be flat - if instead set to `False`, which is the original behavior, the output  of matvec and/or rmatvec will be reshaped in the shape of `Op.dimsd` or `Op.dims` even when the input was a 1d array. 
- The `__add__` and `dot` methods of `LinearOperator` have been modified to check for inconsistencies in the `forceflat` property of the operators to be combined. Basically, if both of them have `forceflat` set to a boolean and they are inconsistent, an error will be raised.
- The `dot` method is modified such that when the input is a numpy array, the `forceflat` parameters is also used to check whether the input and output arrays should be reshaped into ndarrays or kept as 1darrays
-`Bilinear`, `Identity`,`MatrixMult`,  `Sum`, and composition operators `HStack`, `VStack`, and `BlockDiag` are ambiguous operators. As such we introduce`forceflat` as an additional input parameter. By choosing `forceflat=True` we ensure that the output of matvec/rmatvec is always a 1darray when the input is 1darray. 
- For `HStack` and `VStack`, there may be scenarios where Ndarrays are not a meaningful option (e.g., where `dims` of the two operators differ), in this case we internally force `forceflat=True` (not sure whether we should return a warning or not?)
- For `BlockDiag`, we check if dims/dimsd of all operators are the same and stack over a new zero axis (e.g., [nops, *dims]). However we also want to make sure that if we add an operator to the left or right the output can be either Ndarray or 1darray and chosen by the user via `forceflat` in `BlockDiag`.
- For `Block`, I wasn't able to identify any scenario where `forceflat` would be needed. To be on the safe side, I added it anyways*.
- `Identity` has been modified to accommodate for nd-arrays. For backward compatibiilty, `N` and `M` are kept, but they allow now also passing tuples (effectively becoming as if they were `dimsd` and `dims`).

* See Discussion

## Operators with missing dims/dimsd
- `Kronecker`: not sure how dims/dimsd would apply to this operator
 
## Discussion
I have identified a possibility for `VStack`, `HStack`, and `Block`,  which I think ideally makes sense but it becomes a breaking change. Basically, the idea is that we could have `dimsd` in VStack equal `nops x *dims(op)` and similar for HStack with `dims`. However to be backward compatible we would need to have `forceflat=True` as default. This would however force flat inputs and outputs also when VStack/HStack is chained with other operators, which is not the case... unless someone has any smart idea, we may need to accept this new behavior cannot be added :(

### DIFF
--- a/MIGRATION_V1_V2.md
+++ b/MIGRATION_V1_V2.md
@@ -10,7 +10,7 @@ should be used as a checklist when converting a piece of code using PyLops from 
   for every operator by default. While the change is mostly backwards compatible, there are some operators (e.g. the ``Bilinear``
   transpose/conjugate) which can output reshaped arrays instead of 1d-arrays. To ensure no breakage, you can entirely disable this
   feature either globally by setting ``pylops.set_ndarray_multiplication(False)``, or locally with the context manager
-  ``pylops.disabled_ndarray_multiplication()``. Both will revert to v1.x behavior. At this time, PyLops sparse solvers do
+  ``pylops.disabled_ndarray_multiplication()``. Both will revert to v1.x behavior. At this time, PyLops solvers do
   *not* support N-D array multiplication.
 
   See the table at the end of this document for support ndarray operations.
--- a/pylops/basicoperators/block.py
+++ b/pylops/basicoperators/block.py
@@ -1,7 +1,6 @@
 __all__ = ["Block"]
 
 import multiprocessing as mp
-
 from typing import Iterable, Optional
 
 from pylops import LinearOperator
@@ -15,12 +14,18 @@ class _Block(LinearOperator):
     Used to be able to provide operators from different libraries to
     Block.
     """
-    def __init__(self, ops: Iterable[Iterable[LinearOperator]],
-                 dtype: Optional[DTypeLike] = None,
-                 _HStack=HStack,
-                 _VStack=VStack,
-                 args_HStack: Optional[dict] = None,
-                 args_VStack: Optional[dict] = None, name: str = 'B'):
+
+    def __init__(
+        self,
+        ops: Iterable[Iterable[LinearOperator]],
+        forceflat: bool = None,
+        dtype: Optional[DTypeLike] = None,
+        _HStack=HStack,
+        _VStack=VStack,
+        args_HStack: Optional[dict] = None,
+        args_VStack: Optional[dict] = None,
+        name: str = "B",
+    ):
         if args_HStack is None:
             self.args_HStack = {}
         else:
@@ -30,7 +35,12 @@ class _Block(LinearOperator):
         else:
             self.args_VStack = args_VStack
         hblocks = [_HStack(hblock, dtype=dtype, **self.args_HStack) for hblock in ops]
-        super().__init__(Op=_VStack(ops=hblocks, dtype=dtype, **self.args_VStack), name=name)
+        super().__init__(
+            Op=_VStack(
+                ops=hblocks, forceflat=forceflat, dtype=dtype, **self.args_VStack
+            ),
+            name=name,
+        )
 
     def _matvec(self, x: NDArray) -> NDArray:
         return super()._matvec(x)
@@ -53,6 +63,10 @@ class Block(_Block):
     nproc : :obj:`int`, optional
         Number of processes used to evaluate the N operators in parallel using
         ``multiprocessing``. If ``nproc=1``, work in serial mode.
+    forceflat : :obj:`bool`, optional
+        .. versionadded:: 2.2.0
+
+        Force an array to be flattened after rmatvec.
     dtype : :obj:`str`, optional
         Type of elements in input array.
 
@@ -123,9 +137,16 @@ class Block(_Block):
         \end{bmatrix}
 
     """
-    def __init__(self, ops: Iterable[Iterable[LinearOperator]],
-                 nproc: int = 1,
-                 dtype: Optional[DTypeLike] = None):
+
+    def __init__(
+        self,
+        ops: Iterable[Iterable[LinearOperator]],
+        nproc: int = 1,
+        forceflat: bool = None,
+        dtype: Optional[DTypeLike] = None,
+    ):
         if nproc > 1:
             self.pool = mp.Pool(processes=nproc)
-        super().__init__(ops=ops, dtype=dtype, args_VStack={"nproc": nproc})
+        super().__init__(
+            ops=ops, forceflat=forceflat, dtype=dtype, args_VStack={"nproc": nproc}
+        )
--- a/pylops/basicoperators/blockdiag.py
+++ b/pylops/basicoperators/blockdiag.py
@@ -44,6 +44,10 @@ class BlockDiag(LinearOperator):
     nproc : :obj:`int`, optional
         Number of processes used to evaluate the N operators in parallel using
         ``multiprocessing``. If ``nproc=1``, work in serial mode.
+    forceflat : :obj:`bool`, optional
+        .. versionadded:: 2.2.0
+
+        Force an array to be flattened after matvec and rmatvec.
     dtype : :obj:`str`, optional
         Type of elements in input array.
 
@@ -108,6 +112,7 @@ class BlockDiag(LinearOperator):
         self,
         ops: Sequence[LinearOperator],
         nproc: int = 1,
+        forceflat: bool = None,
         dtype: Optional[DTypeLike] = None,
     ) -> None:
         self.ops = ops
@@ -130,7 +135,12 @@ class BlockDiag(LinearOperator):
 
         dtype = _get_dtype(ops) if dtype is None else np.dtype(dtype)
         clinear = all([getattr(oper, "clinear", True) for oper in self.ops])
-        super().__init__(dtype=dtype, shape=(self.nops, self.mops), clinear=clinear)
+        super().__init__(
+            dtype=dtype,
+            shape=(self.nops, self.mops),
+            clinear=clinear,
+            forceflat=forceflat,
+        )
 
     @property
     def nproc(self) -> int:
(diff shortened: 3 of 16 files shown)


mrava87:
@cako @RendersJens, it would be good to get your feedback on this before I continue working on more ambiguous operators as well as go through all operators to check where we are currently missing `dims\dimsd`.

mrava87:
@cako @RendersJens, any feedback?

RendersJens:
Thanks for this work. I just ran some tests and, pending the nd-extention of identity, this solves all my issues. I do have some thoughts:

* The question of generalizing BlockDiag to nd arrays is for me equivalent to extending the concatenation of 1d arrays to nd. So we may follow the convention of numpy concatenate, which is to specify the axis along which to concatenate. The shape of the arrays can then differ along this axis.
* The behavior of product operator still seems unexpected to me. If I compose `F @ M` where `F` is a 2D FFT and `M` works only on 1D arrays, we still get the previous unexpected behavior. This would be the case if `M`is a `MatrixMult`  for example. Or if `M` is a 1D identity (notice that a user would have to know to use the ND functionality of Identity to get the 1D functionality to work again, which is a bit strange). In this case I cannot set `forceflat=True` by a keyword of FFT. I can do `F.forceflat=True` and this actually solves it.

mrava87:
> Thanks for this work. I just ran some tests and, pending the nd-extention of identity, this solves all my issues. I do have some thoughts:

Thanks! Good to hear that :) I'll move to the Identity operator now.
> 
> * The question of generalizing BlockDiag to nd arrays is for me equivalent to extending the concatenation of 1d arrays to nd. So we may follow the convention of numpy concatenate, which is to specify the axis along which to concatenate, with default `axis=0`. The shape of the arrays can then differ along this axis.

I think BlockDiag and np.concatenate are not so similar. I would say the only way to see a similarity between np.concatenate and BlockDiag is to have this case `np.concatenate([a[np.newaxis], b[np.newaxis]], axis=0)` where you concatenate over a new axis in front... This is what I meant in my comment as the only scenario I can see. I will implement this :)

> * The behavior of product operator still seems unexpected to me. If I compose `F @ M` where `F` is a 2D FFT and `M` works only on 1D arrays, we still get the previous unexpected behavior. This would be the case if `M`is a `MatrixMult`  for example. Or if `M` is a 1D identity (notice that a user would have to know to use the ND functionality of Identity to get the 1D functionality to work again, which is a bit strange). In this case I cannot set `forceflat=True` by a keyword of FFT. I can do `F.forceflat=True` and this actually solves it.

Right, I see. This is because MatrixMult with otherdims=None is also a 1D-to-1D operator. I have added the `forceflatten` option to this operator and now this should work as you expect:

```
import pylops
import numpy as np

M = pylops.MatrixMult(np.ones((30, 25)))
M1 = pylops.MatrixMult(np.ones((30, 25)), forceflat=True)
F = pylops.signalprocessing.FFT2D((5, 5))
Op = M @ F
Op1 = M1 @ F
print(Op.forceflat)
print(Op1.forceflat)

print((Op @ np.ones(Op.dims)).shape, (Op @ np.ones(Op.shape[1])).shape)
print((Op.H @ np.ones(Op.dimsd)).shape, (Op.H @ np.ones(Op.shape[0])).shape)

print((Op1 @ np.ones(Op.dims)).shape, (Op1 @ np.ones(Op.shape[1])).shape)
print((Op1.H @ np.ones(Op.dimsd)).shape, (Op1.H @ np.ones(Op.shape[0])).shape)
```

FFT is not an ambiguous operator, so I dont like the idea of a user having to force forceflatten for that operator.

I am not sure what you mean `notice that a user would have to know to use the ND functionality of Identity to get the 1D functionality to work again, which is a bit strange`? Of course a user needs to know how PyLopsv2 works to benefit from it, otherwise as Carlos said they can just set the pylopsV1 behavior at the start of their code and not learn the new behavior ;)

cako:
Hi all! Thanks for looking into this @mrava87. I have looked over the code and I think it's fine in terms of solving the issues we discussed with minimal impact. The only thing I am still struggling with is accepting the fact that now we have a property common to *all* LinearOperators (aka, expanding the API) but then not offering the possibility of constructing them with this property doesn't seem like a good design choice. In addition we now have three ways of disabling ND-array multiplication: one via class property, and one globally, and one within a context.

Some solutions to discuss:
1. Force the user to wrap their local flattened multiplications with the disable_ndarray decorator. This is the current solution and I agree it feels clunky.
2. Force the user to use matvec and rmatvec to get flattened operations always. This is definitely a bad idea, I would not like to see this ever.
3. Add forceflat as an part of the LinearOperator API and add it to the `__init__` of all operators
4. Add forceflat as a property of only some operators, and upon multiplication/addition/etc, first check if the property is present (as opposed to being None)
5. Add forceflat as part of the LinearOperator API but only add to the `__init__` of ambiguous operators. This is this draft's proposal.

I am currently teetering between 3 and 4, what do you guys think?

mrava87:
Thanks @cako. Let me explain what was my thought process here.

First, what I did here can indeed be done already with the current codebase using the context manager when needed or the global setting. But, as @RendersJens pointed out, I find this my go-to choice when I’m in a rush in research and realize there is a problem with 1D vs ND arrays, but it should not be the default for pylops (from a dev point of view). So 1 goes out of the picture.

2 is also not a good idea as you say because it goes against the whole concept of pylops being a very high-level, expressive library where users write code almost like the math the write on paper. Having said that, within solvers (so for devs), I highly reccomend not using @ and .H @ as this introduced unwanted overhead (see tutorial 1). Now all solvers use matvec/rmatvec.

Now, my idea is that by adding forceflat to LinearOperator, we don’t need to do checks if this is present but just checks on what is the value. If you think about this isn’t very different from for example clinear and explicit. They are in fact present in all operators but ‘make sense’ only for some. Of course, there is a small difference: whilst for clinear and explicit we never ask the user to set them, now forceflat is exposed to the users when it makes sense. And here comes my main point, and the reason why I personally see 5 as the best option.

 My main point is explained with the concept of ambiguous operator: we know upfront when we design an operator if this is ambiguous or any of its combination of input parameters can make it ambiguous. In this case, if we don’t ask the user to solve this ambiguity we cannot do it under the hood and make sure the user is happy with it. See @RendersJens example, he doesn’t want his Identity to behave as it behaves a default (favouring ND or 1D in ambiguous scenarios). But if we take an operator like FFT, which will always go from and to Nd (N=1,2,3,4… BUT the same for model and vector), what is the point to expose an extra parameter to the users that is actually meaningless - I think this would just add more confusion and not really benefit anyone.

In summary, I think I am a bit more in favour thar @RendersJens to having ND arrays as default (reason why I was apply to go this way for pylopsv2) but I still would like pylops to be a easy to use library not just for experts but for anyone with basic knowledge of inverse problems. So I would always encourage not to ask the user to make a choice where we can be smart an handle it under the hood without ambiguity (which is what we are not doing now and I hope we can do again after this PR).

What do you think?

mrava87:
@cako, I believe this is ready, could you please review it :)

Hopefully after we merge this, most of the inconsistencies @RendersJens observed will be gone

mrava87:
@cako did you get any time to look into this yet?

cako (approved):
Great stuff!! I feel like the proposed solution is a good compromise that will make everybody happy. It is non-intrusive enough that most users will not notice it, but offers a way out in case of ambiguous operators. Also nice that it adds dims/dimsd to `Identity` and `Zero`.

In terms of code quality, everything looks pretty good, but I did make a few changes. I added some simple tests for verifying the forceflat behavior in `Sum` (based you your notebook @mrava87), and replaced some of the remaining "naked" multiplications in cls_sparsity for `Op.matvec`. Please check if that is appropriate.

Finally, I also refactored the forceflat logic in the `LinearOperator` definition. Again, please verify that it makes sense, but I think the code is cleaner and more Pythonic. E.g., checking `isinstance(X, bool)` when `X` is `Optional[bool]` is not ideal; best to check `X is not None`. Also, there was an if/if/else block which IMO should be if/elif/else. Anyways, have a look and feel free to change anything.

For now I will approve the changes and let you merge :)

mrava87:
Great, thanks a lot for reviewing the code :) 

I totally agree with your changes, I am going to merge this.

Issue #439, opened 2022-09-24
Improvements to Kirchhoff operator


mrava87:
# Motivation 

The ``pylops.waveeqprocessing.Kirchhoff`` operator has gone through a major restyling in https://github.com/PyLops/pylops/pull/431 mostly aimed at including dynamic capabilities. From the discussion in the associated PR we have identified the following areas that may need further improvement. Moreover some of the new features in such PR are so far only available for the 2D implementation.

# Features to be implemented

- [] Add 2.D wavelet filtering in `_wavelet_reshaping`. Current roadblock is represented by the unclear definition of σ_S and σ_R in equation 2.3.7 of Safron2018 (neither in the Bleistein and Gray, 2002 paper).
- [] Devise a new definition of `aperture` not simply based on the offset/depth ratio to be used for heterogeneous velocity models. Currently we raise a warning when a user decides to activate such aperture limitation
- [] Finalize discussion about the role of `snell`, and if this is actually needed at all.
- [] Finalize discussion about angle scaling and introduce `cospower` as discussed in the PR
- [] Implement angle scaling with normal (instead of vertical) when `anglerefl` is provided
- [] Implement all features when `dynamic=True` in 3D

mrava87:
See https://github.com/PyLops/pylops/pull/431#issuecomment-1272294726

dikwickley:
hello @mrava87 i would like to work on this issue. can you please assign it to me?

dikwickley:
hello @mrava87 @cako 
i would like to work on this issue. can you please assign it to me?

mrava87:
Sure! get in touch with us (@cako) on slack so we can discuss better maybe where it is better to start :)

PR #531, opened 2023-10-01
feature: restyling of Kirchhoff operator

mrava87:
This PR introduces a number of changes to the Kirchhoff operator in an attempt to improve the amplitude fidelity of this modelling process with respect to finite-difference full wavefield modelling. Also follows from https://github.com/PyLops/pylops/issues/439.

Major changes:

- Remove `anglerefl` as the angle `theta` can be computed directly from the incident and reflection angles with normal
- Remove `snell` limitations, this is naturally handled by stationary-phase when summing along all possible image points and src-rec pairs
- Force `amp` to be provided as pair of tables like `trav`
- When `amp` is not provided, compute it as `1/sqrt(dist)` for both source and receiver side in 2D (and `1/dist` in 3D)
- Correct computation of `cos(\theta)`
- Documentation polished and fully consistent with code

### DIFF
--- a/.github/workflows/codacy-coverage-reporter.yaml
+++ b/.github/workflows/codacy-coverage-reporter.yaml
@@ -8,8 +8,8 @@ jobs:
   build:
     strategy:
       matrix:
-        platform: [ ubuntu-latest, macos-latest ]
-        python-version: ["3.8", "3.9", "3.10"]
+        platform: [ ubuntu-latest, ]
+        python-version: ["3.8", ]
 
     runs-on: ${{ matrix.platform }}
     steps:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ repos:
           - --line-length=88
 
   - repo: https://github.com/pycqa/isort
-    rev: 5.10.1
+    rev: 5.12.0
     hooks:
       - id: isort
         name: isort (python)
--- a/pylops/waveeqprocessing/kirchhoff.py
+++ b/pylops/waveeqprocessing/kirchhoff.py
@@ -133,7 +133,7 @@ class Kirchhoff(LinearOperator):
     -----
     The Kirchhoff demigration operator synthesizes seismic data given a
     propagation velocity model :math:`v(\mathbf{x})` and a
-    reflectivity model :math:`m(\mathbf{x})`. In forward mode [1]_, [2]_:
+    reflectivity model :math:`m(\mathbf{x})`. In forward mode [1]_, [2]_, [3]_:
 
     .. math::
         d(\mathbf{x_r}, \mathbf{x_s}, t) =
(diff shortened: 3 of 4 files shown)


mrava87:
@cako, I got some time to revisit the Kirchhoff operator based on some the discussions we had in https://github.com/PyLops/pylops/issues/439. The dynamic version looks much better now in terms of match with FD (see https://github.com/PyLops/pylops_notebooks/blob/master/developement/Modellingcomparison_kirch_vs_fd.ipynb).

Do you think you have some time to review this?

cako:
I should have some time, yes. I will take a look!

.github/workflows/codacy-coverage-reporter.yaml
cako, line 1: Check that this is right

.pre-commit-config.yaml
cako, line 1: I had to do this to be able to run the precommit hook. See this: https://stackoverflow.com/questions/75269700/pre-commit-fails-to-install-isort-5-11-4-with-error-runtimeerror-the-poetry-co

pylops/waveeqprocessing/kirchhoff.py
```diff
@@ -288,81 +274,93 @@ def __init__(
...
                     trav_recs_grad,
                 ) = Kirchhoff._traveltime_table(z, x, srcs, recs, vel, y=y, mode=mode)
                 if self.dynamic:
                     # need to add a scalar in the denominator of amplitude term to avoid
-                    # division by 0, currently set to 1/100 of max distance to avoid having
-                    # to large scaling around the source. This number may change in future
+                    # division by 0, currently set to 1e-4 of max distance to avoid having
+                    # too large scaling around the source. This number may change in future
                     # or left to the user to define
                     epsdist = 1e-2
```
cako, line 308: Should this be set to 1e-4 as per the comment?

cako (approved):
Great stuff! Code LGTM so I'm going to go ahead and approve it.

I do have a couple of comments/questions.

1. Why change the sign of the wavelet? Was it not matching the data as expected?
2. Applying the cosine of the reflection angle or using the apertures ensures some protection against very wide angles, but the Snell parameter is the best way to control this. However, upon rereading the removed code I don't think it was implemented properly so best to remove it for now. (E.g., if ang_src = 0 and ang_rec = 360, it would complain that the reflection angle is too big... it should be implemented on cos((ang_src - ang_rec)/2)).

.pre-commit-config.yaml
mrava87: Interesting, I didnt face this issue but maybe its because I have an old pre-commit installed... happy with the change

pylops/waveeqprocessing/kirchhoff.py
mrava87: This is indeed inconsistent. I honestly do not know what is best to do to avoid blow up near source/rec... my experiments suggest 1e-2 is a reasonable choice, for now I fix the comment, but if you have an idea how to do this better I am happy to amend it

cako:
Interesting! I didn't know that paper. Is there a significant difference between sqrt(jw) and |w|? In terms of frequency content they should be the same. The phase might be a bit altered. I think if yours matches FD simulation then it is probably the preferred way to go... with the caveat that traditionally Kirchhoff modelers try to do a good job at modeling the amplitude of reflections and not diffractions. A bit counter-intuitive in a way! In any case, I don't mind one way or the other, there are dozens of expressions in the literature, I think it's just a matter of choosing one.

Regarding the angle, I definitely agree with keeping the cos of the reflection angle. I just mean that one may want to further modify the diffraction stack based on a user-given weight. This can be useful for higher control when killing the tomographic term. In any case, we can handle it in a future PR if we want to tackle it.

Feel free to merge!

PR #532, opened 2023-10-03
remark on classical sinograms added + jacobian correction; added two …

fedor-goncharov:
…plots - data with and without jacobian correction; added more explanations on implementation of custom curve

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,3 +148,4 @@ A list of video tutorials to learn more about PyLops:
 * Aniket Singh Rawat, dikwickley
 * Rohan Babbar, rohanbabbar04
 * Wei Zhang, ZhangWeiGeo
+* Fedor Goncharov, fedor-goncharov
--- a/docs/source/credits.rst
+++ b/docs/source/credits.rst
@@ -20,3 +20,4 @@ Contributors
 *  `Aniket Singh Rawat <https://github.com/dikwickley>`_, dikwickley
 *  `Rohan Babbar <https://github.com/rohanbabbar04>`_, rohanbabbar04
 *  `Wei Zhang <https://github.com/ZhangWeiGeo>`_, ZhangWeiGeo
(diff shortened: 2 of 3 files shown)


mrava87:
Thanks @fedor-goncharov, I'll take a look in the next few days and let you know if I have any question :D

mrava87:
Definitely, this makes sense. The operator was originally built with geophysical applications in mind but the idea of allowing users to specify a parametric curve was actually meant to give more freedom to users. I do not follow entirely what you mean 'curves with natural parametrization' 5
though, would you envision writing a function that instead of having just a parametric straight lines has different line definitions based on the slope. Say if slope is from 0-1 you use a curve t=f(x) and if bigger you turn the curve around?

Would you think this is feasible with the code as is or you see some logic needs changed?

PS: I'll review your PR soon, and we can always remodify the tutorial after your suggested changes to Radon2D are implemented 😀

mrava87:
Thanks @fedor-goncharov for the contribution. It is great to have added such info to the tutorial so people familiar with CT imaging will not get confused with what is done here vs what they usually do.

Regarding the point of changing Radon2D feel free to open an issue to discuss or make a PR if you have an idea of how this could be easily implemented with the current Radon2D logic :)

mrava87 (approved):

fedor-goncharov:
Thanks! (my first external PR actually!) For the Radon integration, yes it is always better to keep things in issues/PRs :) 

p.s. Just to precise - when the slope of the curve becomes larger - integrating wrt 'x' becomes unstable if you are targeting
value of the integral simply along the line wrt its natural parametrization (natural parametrization - when parameter 
corresponds (up to constant shift) to the length on the curve). 

Anyway, I just wanted to answer your question above. I will think/look on this and will propose an issue when it will be clear for me.

PR #556, opened 2023-11-29
Quick fix for a typo in `tutorials/dottest.py`

alex-rakowski:
changes:
`gs = pltgs.GridSpec(1, 9)` to `gs = plt.GridSpec(1, 9)`


mrava87:
I think is due to the import `import matplotlib.gridspec as pltgs`. Perhaps nowadays it is not needed anymore to import that specific submodule, but if we do the change you suggest we should also remove `import matplotlib.gridspec as pltgs`. Agree?

alex-rakowski:
I made a mistake. I didn't notice the import when I was copying and working through the example.

PR #571, opened 2024-03-04
fix: ensure dtcwt is not used in typing annotations

mrava87:
The previous PR introduced an unwanted import error of pylops when `dtcwt` was not installed; this was due to the fact we used dtcwt for typing annotation. 

The current solution is not to use for any typing annotation to ensure that no problem arises if `dtcwt` is not installed, given that it is an optional dependency


mrava87:
Oh, good idea :) I didn't think about that, but it make sense

mrava87:
@cako I created a PR, have a look if what I did is in line with your suggestion :)

PR #573, opened 2024-03-04
Fix incorrect dips check in `slope_estimation

### LINKED ISSUE #572: incorrect check on dips argument in in `slope_estimation
The documentation states the following: 

image

However, the logic flow in the code is the opposite:

https://github.com/PyLops/pylops/blob/0d19f5689978cb71e743bf673d4ce775e22817c6/pylops/utils/signalprocessing.py#L249-L253

I double checked with the original implementation of `dips_estimate` 

https://github.com/PyLops/pylops/blob/c5844c4a4b25cdb74a5cef12b926d6d91da20095/pylops/utils/signalprocessing.py#L79-L182

And it is inline with the current documentation. It appears that the bug was introduced when `dips_estimate` and slope_estimate` merged.  

I have a PR ready for this.

Beramos:
closes #572. 

Quick fix but I have also added a unit test to see if the regularisation is applied when NOT using dips as a way to check which algorithm is used.

### DIFF
--- a/pylops/utils/signalprocessing.py
+++ b/pylops/utils/signalprocessing.py
@@ -246,7 +246,7 @@ def slope_estimate(
     regdata = l1 > eps
     anisos[regdata] = 1 - l2[regdata] / l1[regdata]
 
-    if not dips:
+    if dips:
         slopes = 0.5 * np.arctan2(2 * gzx, gzz - gxx)
     else:
         regdata = np.abs(gzx) > eps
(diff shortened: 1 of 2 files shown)


mrava87:
Adding @cako here

mrava87:
Alright, great! Then we can merge this :)

mrava87:
Thanks @Beramos, I can merge this now :)

Beramos:
@mrava87, awesome!

PR #575, opened 2024-03-16
doc: added safe typing to dtcwt

mrava87:
This PR adds typing annotations for `dtcwt` objects in a safe way, such that it doesn't result in an error if the library is not installed


cako:
Original was working, but giving some issues in PyLance. To make it happy I just introduced a type variable instead of using a type alias. Should be good to go now!

cako (approved):

mrava87:
Good stuff :)

PR #578, opened 2024-05-18
Feature: new Sliding and Patch operators

mrava87:
In this PR, I provide a new implementation for all Sliding and Patch operators that presents the following features:
- the operators are now directly implemented with `_matvec` and `_rmatvec` methods. The old implementation, although elegant in that it relied on using other pylops operators, showed to be inefficient especially in cases with very large number of windows.
- Two kind of operators `Op` can be provided: the first applies a single transformation to each window separately (as default from before); the second applies the transformation to all of the windows at the same time.
- `nproc` in Sliding3D is deprecated since we do not use anymore HStack/VStack operators that have a multi-threading option. Currently, all of the new implementations are single threaded (although inside they may call numpy methods that are multi-threaded).
- A new parameter `savetaper` in added to all operators: if ``True``, all of the tapers are created and saved in a numpy array so that they can be efficiently applied in one go as element-wise multiplication of the numpy arrays (this is fast but can be memory demanding); if ``False``, only unique tapers are saved and applied one by one (this is a bit slower but more memory friendly)

All of the new operators have been benchmarked against the hold one, and an analysis of both the timings and memory usage of the different operators has been performed here: https://github.com/PyLops/pylops_slidingpatching

### DIFF
--- a/pylops/signalprocessing/patch2d.py
+++ b/pylops/signalprocessing/patch2d.py
@@ -297,6 +297,27 @@ class Patch2D(LinearOperator):
 
         self._register_multiplications(self.savetaper)
 
+    def _apply_taper(self, ywins, iwin0, iwin1):
+        if iwin0 == 0 and iwin1 == 0:
+            ywins[0, 0] = self.taps[0, 0] * ywins[0, 0]
+        elif iwin0 == 0 and iwin1 == self.dims[1] - 1:
+            ywins[0, -1] = self.taps[0, -1] * ywins[0, -1]
+        elif iwin0 == 0:
+            ywins[0, iwin1] = self.taps[0, 1] * ywins[0, iwin1]
+        elif iwin0 == self.dims[0] - 1 and iwin1 == 0:
+            ywins[-1, 0] = self.taps[-1, 0] * ywins[-1, 0]
+        elif iwin0 == self.dims[0] - 1 and iwin1 == self.dims[1] - 1:
+            ywins[-1, -1] = self.taps[-1, -1] * ywins[-1, -1]
+        elif iwin0 == self.dims[0] - 1:
+            ywins[-1, iwin1] = self.taps[-1, 1] * ywins[-1, iwin1]
+        elif iwin1 == 0:
+            ywins[iwin0, 0] = self.taps[1, 0] * ywins[iwin0, 0]
+        elif iwin1 == self.dims[1] - 1:
+            ywins[iwin0, -1] = self.taps[1, -1] * ywins[iwin0, -1]
+        else:
+            ywins[iwin0, iwin1] = self.taps[1, 1] * ywins[iwin0, iwin1]
+        return ywins
+
     @reshaped()
     def _matvec_savetaper(self, x: NDArray) -> NDArray:
         ncp = get_array_module(x)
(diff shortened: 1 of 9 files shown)


mrava87:
@cako not sure if you have time to take a look at this, any comment/suggestion will be welcome 😄

cako:
Will have a look at it! Could you make this repo public: https://github.com/mrava87/pylops_slidingpatching

mrava87:
Ops, yes :) I also moved it to the PyLops organization

pylops/signalprocessing/patch2d.py
```diff
@@ -172,104 +173,268 @@ def Patch2D(
...
+            ywins[iwin0, -1] = self.taps[1, -1] * ywins[iwin0, -1]
+        else:
+            ywins[iwin0, iwin1] = self.taps[1, 1] * ywins[iwin0, iwin1]
+        return ywins
 
-    combining0 = HStack(
-        [
-            Restriction(dimsd, range(win_in, win_end), axis=0, dtype=Op.dtype).H
-            for win_in, win_end in zip(dwin0_ins, dwin0_ends)
+    @reshaped()
```
cako, line 321: No need for the parens: `@reshaped`

pylops/signalprocessing/patch2d.py
```diff
@@ -172,104 +173,268 @@ def Patch2D(
...
+    def _matvec_savetaper(self, x: NDArray) -> NDArray:
+        ncp = get_array_module(x)
+        if self.tapertype is not None:
+            self.taps = to_cupy_conditional(x, self.taps)
```
cako, line 325: I realized this function may have a "bug", the GPU array gets created on the current device, not the device of `x`.

The function should probably look like this (untested!):

```python
def to_cupy_conditional(x: npt.ArrayLike, y: npt.ArrayLike) -> NDArray:
    if deps.cupy_enabled:
        if cp.get_array_module(x) == cp and cp.get_array_module(y) == np:
            with cp.cuda.Device(x.device):
                y = cp.asarray(y)
    return y
```

pylops/signalprocessing/patch2d.py
```diff
@@ -172,104 +173,268 @@ def Patch2D(
...
+                    tapbottom,
+                    taprightbottom,
+                ]
+                self.taps = np.vstack(taps).reshape(3, 3, nwin[0], nwin[1])
+
+        # define scalings
+        if scalings is None:
+            self.scalings = [1.0] * nwins
+        else:
+            self.scalings = scalings
```
cako, line 282: if else definitions are generally clearer inline:

```python
self.scalings = [1.0] * nwins if scalings is None else scalings
```

pylops/signalprocessing/patch2d.py
```diff
@@ -172,104 +173,268 @@ def Patch2D(
...
+        else:
+            y = ncp.zeros(self.dims, dtype=self.dtype)
+            for iwin0 in range(self.dims[0]):
+                for iwin1 in range(self.dims[1]):
+                    y[iwin0, iwin1] = self.Op.rmatvec(
+                        ywins[iwin0, iwin1].ravel()
+                    ).reshape(self.dims[2], self.dims[3])
+        return y
+
+    @reshaped()
```
cako, line 368: parens

pylops/signalprocessing/patch3d.py
```diff
@@ -185,272 +186,581 @@ def Patch3D(
...
+            ywins[iwin0, -1, iwin2] = self.taps[1, -1, 1] * ywins[iwin0, -1, iwin2]
+        elif iwin2 == 0:
+            ywins[iwin0, iwin1, 0] = self.taps[1, 1, 0] * ywins[iwin0, iwin1, 0]
+        elif iwin2 == self.dims[2] - 1:
+            ywins[iwin0, iwin1, -1] = self.taps[1, 1, -1] * ywins[iwin0, iwin1, -1]
+        else:
+            ywins[iwin0, iwin1, iwin2] = self.taps[1, 1, 1] * ywins[iwin0, iwin1, iwin2]
+        return ywins
+
+    @reshaped()
```
cako, line 579: parens

cako (approved):
Added a few small comments, code looks good. I ran your benchmark repo on CPU, and everything worked as expected. The benchmarks were also followed your numbers pretty closely in which transforms were the fastest. So, LGTM!

mrava87:
@cako thanks a lot for the comments, and great to hear that you also run the benchmarks on your machine with similar results!

I’ll address your comments in the next few days and then merge this PR 😀

pylops/signalprocessing/patch2d.py
mrava87: Done

pylops/signalprocessing/patch2d.py
mrava87: Done

pylops/signalprocessing/patch2d.py
mrava87: I think you are right, well spotted. I am going to open an issue and take this separetely.

Issue #579, opened 2024-05-22
bug: to_cupy_conditional device selection


mrava87:
The `to_cupy_conditional` method creates cupy array on the current device instead of the device of `x`.

Proposed solution in https://github.com/PyLops/pylops/pull/578:

```
def to_cupy_conditional(x: npt.ArrayLike, y: npt.ArrayLike) -> NDArray:
    if deps.cupy_enabled:
        if cp.get_array_module(x) == cp and cp.get_array_module(y) == np:
            with cp.cuda.Device(x.device):
                y = cp.asarray(y)
    return y
```

mrava87:
Indeed, after the suggested fix, this code:

```
cp.cuda.Device(device=0).use()

xcurrent = cp.ones(10)

with cp.cuda.Device(1):
    xother = cp.ones(10)

print(xcurrent.device, xother.device)

y = np.zeros(5)

# from current device
ycurrent = to_cupy_conditional(xcurrent, y)

# from other device
yother = to_cupy_conditional(xother, y)

print(ycurrent.device, yother.device)
```

gives:
```
<CUDA Device 0> <CUDA Device 1>
<CUDA Device 0> <CUDA Device 1>
```

PR #580, opened 2024-05-22
bug: to_cupy_conditional

mrava87:
Fixes https://github.com/PyLops/pylops/issues/579


mrava87:
@cako just to make you aware, no need to review really as i just implemented your suggested fix 😄

PR #586, opened 2024-06-18
Feature: migrate to numpy v2.0.0 (OLD)

mrava87:


mrava87:
Too early, too many of our other dependencies still do not support numpy v2...

mrava87:
Seems like the only library that could take a while to support numpyv2 is `dtcwt`. For this reason, we currently build the documentation with numpy<2.0.0

PR #589, opened 2024-07-02
Patch: scipy v1.11

mrava87:
This PR is aimed at solving some of the issues arising in our test suite when scipy v1.14 is used. This is mostly related to an (announced) breaking change in the new version for the `scipy.sparse.linalg.cg` solver, namely:

```
The tol argument of scipy.sparse.linalg.{bcg,bicstab,cg,cgs,gcrotmk, mres,lgmres,minres,qmr,tfqmr} 
has been removed in favour of rtol. Furthermore, the default value of atol for these 
functions has changed to 0.0.
```

See https://docs.scipy.org/doc/scipy/release/1.14.0-notes.html

As a result of this change, we decided to:

- Bump the minimal python requirement to 3.9
- Bump the minimal scipy requirement to 1.13
- Change `tol` into `atol` when passed to the scipy cg solver in `NormalEquationsInversion`

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -7,7 +7,7 @@ jobs:
     strategy:
       matrix:
         platform: [ ubuntu-latest, macos-13 ]
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11"]
 
     runs-on: ${{ matrix.platform }}
     steps:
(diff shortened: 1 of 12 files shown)


mrava87:
@cako can you take a look and let me know what you think. This is the first step towards supporting newer versions of scipy and numpy (with numpy v2 as target once all other dependencies do)

mrava87:
Hi @cako, just checking you saw this… it is not a major thing but since effectively it is an inner breaking change (we don’t really trigger it, scipy does but someone may have the same code we had in one of the tests, which will crash…) I wanted a second opining before going ahead :)

PR #594, opened 2024-07-25
Update that increases performance (AcousticWave2D)

guaacoelho:
Hey, my name is Gustavo Coelho. I hope this PR is useful for you.

During the creation of the Born operator, Devito uses an optimization technique that caches the result so that for subsequent executions with the same parameters, it doesn't need to recreate the operator, but simply retrieves it from the cache.
 
However, with a new solver being created for each shot, Pylops couldn't take advantage of this feature. This PR aims to address this issue by moving the creation of the solver outside the shots iteration loop.

Additionally, we are working on other features for Pylops and plan to make further contributions soon.

### DIFF
--- a/pylops/waveeqprocessing/twoway.py
+++ b/pylops/waveeqprocessing/twoway.py
@@ -327,13 +327,13 @@ class AcousticWave2D(LinearOperator):
                 self.src_wavefield.append(src_wav)
             self.src_illumination += src_ill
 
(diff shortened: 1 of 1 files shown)


mrava87:
Hello @guaacoelho,
thanks for this PR!

I knew about the caching of Devito but didn't realize that it would only work if the solver is not re-created, that definitely makes sense.

I am going to merge this PR. Looking forward to more features :)

PR #615, opened 2024-10-06
bug: fix nans in Spread

mrava87:
This PR fixes a problem introduced in Spread with latest versions of numba. This is due to the way we used to handle nans, a much cleaner way is used now with math.isnan without having to first convert to integrers.

### DIFF
--- a/pylops/basicoperators/_spread_numba.py
+++ b/pylops/basicoperators/_spread_numba.py
@@ -1,3 +1,4 @@
+import math
 import os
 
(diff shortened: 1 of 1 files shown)


mrava87:
@cako FYI, I just installed pylops on my new mac book air and started to have issues with tests related to the Spread operator; turns out that probably something is changed in numba and the way we were handling nans stopped working... the good news is that it forced me to find a much cleaner way (I think!) to do it 😄

PR #621, opened 2024-11-17
Adds interoperability with PyTensor (and PyMC)

cako:
## Description

Adds the ability to call use PyLops Linear Operators as PyTensor [Op](https://pytensor.readthedocs.io/en/latest/extending/op.html)s. This allows PyMC to use PyLops LinearOperators inside of their models.

The PR also includes an example of Bayesian linear regression (`examples/plot_bayeslinearregr.py`) demonstrating this capability.

## Changes
In addition to the main functionality, this PR issues several fixes:
* Ignore `__pycache__`
* Disables html_theme_options not available for current theme
* Various formatting issues in RST, math, etc.

## Notes
Although I added everything to intersphinx, the documentation links is not appearing in my local documentation. Don't know if this is by design or if I need to change anything.


cako:
@mrava87 I've been having a very hard time satisfying the version requirements of all libraries. PyTensor does not support numpy 2. Devito only supports numpy 2 for the latest versions, so dropping to numpy 1.x forces one to use a very old version of devito... which in turns makes it not support newer versions of sympy (above 0.13). On the other hand, PyTorch requires versions of sympy>=0.13.1.

cako:
@mrava87 I think I managed to build it, the trick was to prevent the pip command which install pytorch to overwrite the sympy installation. I did this setting pytorch version to be below 2.5.

Note that these version changes are only required for the `requirements-dev.txt` and `requirements-torch.txt` files because they are used separately when installing. They are not needed in the requirements-doc.txt or the yml files because the dependency managers have access to all packages at once and therefore are able to figure out that you need numpy<2 and pytorch<2.5 to be compatible with pytensor.

cako:
@mrava87 managed to get everything working after a lot of back and forth with versions :)

https://pylops--621.org.readthedocs.build/en/621/gallery/plot_bayeslinearregr.html#sphx-glr-gallery-plot-bayeslinearregr-py

mrava87:
Great! I’ll take a look in the next couple of days 😀

examples/plot_bayeslinearregr.py
```diff
@@ -0,0 +1,223 @@
...
+    "k",
+    lw=4,
+    label=rf"True: $x_0$ = {x[0]:.2f}, $x_1$ = {x[1]:.2f}",
+)
+ax.plot(
+    np.array([t.min(), t.max()]),
+    np.array([t.min(), t.max()]) * xnest[1] + xnest[0],
+    "--g",
+    lw=4,
+    label=rf"MAP Estimated: $x_0$ = {xnest[0]:.2f}, $x_1$ = {xnest[1]:.2f}",
```
mrava87, line 90: To be very picky, this is an ML estimator, not a MAP as no prior knowledge is added... as you say later a MAP will have for example a Tikhonov regularization term... ok, we could say the term is added but the scaling is 0 😉

requirements-doc.txt
```diff
@@ -3,7 +3,7 @@
 # same reason, we force devito==4.8.7 as later versions of devito
 # require numpy>=2.0.0
 numpy>=1.21.0,<2.0.0
-scipy>=1.11.0
+scipy>=1.11.0,<1.13
```
mrava87, line 6: I guess this is for pytensor/pymc to work?

mrava87 (approved):
This looks great 🚀 

I just left a few minor comments... I am happy for this to be merged as soon as you feel it's ready 😄

requirements-doc.txt
cako: This is for arviz to work:

https://discourse.pymc.io/t/importerror-cannot-import-name-gaussian-from-scipy-signal/14170

If we could get arviz to 0.18 we wouldn't need it, but other package dependencies prevent that from happening. It's a mess!

cako:
@mrava87 I actually tested the yml file as it current stands, it didnt install the correct numpy. The only way I could get it to work was to version pin:
```
- numpy>=1.21.0,<2
- scipy>=1.11.0,<1.13
- pytorch>=1.2.0,<2.5
 - pip:
  - devito==4.8.6
```

I'm reluctant to add this to the "standard" environment-dev.yml since that would force everyone to use old versions just so that pytensor still works. I think instead I will remove pytensor from the yml files, and put a check to not execute the pytensor tests if numpy>=2. What do you think?

examples/plot_bayeslinearregr.py
cako: Very good point! I added the MAP for completeness :)

Issue #620, opened 2024-11-08
ToCupy operator


mrava87:
## Motivation
Following up a discussion started in https://github.com/PyLops/pylops/pull/617, our current design of PyLops allows one to solve inverse problems in two modalities: end-to-end CPU and end-to-end GPU.

In practice, when working with arrays that don't fit in the memory of a single machine, the problem of interest can be lifted one level up using `pylops-mpi`. This is the case for both CPU and GPU solutions.

However, there exist scenarios where one may have a model fits into a single GPU, but the data doesn't. This case cannot be handled in the current version of PyLops. Similarly, one may want to only run part of an operator on the GPU or run a stack of operators on the GPU in a serial manner lifting part of the model/data onto the GPU as needed by each operator.

The same approach can be of course lifted one level up again if the entire data does not fit either on a CPU, so splitting it across machines with `pylops-mpi`, and then in each machine using the above suggested strategy when the portion of the data that sits on a single machine does not fit in a single GPU.

## Solution
I suggest to create a simple operator called `ToCupy` whose forward simply lifts a numpy array to a cupy array and whose adjoint does the reverse. This operator can be used with a VStack operator in front of each operator (and its adjoint in the back) so that when the VStack operator is invoked the different operators will run sequentially but each of them can internally operator on cupy arrays.

To ease the understanding, I attach visual description of the two currently available use cases and the two new use cases (the first is already working, the second requires some changes in the design choices of some operators and solvers... I am not sure yet whether it is worth).

A prototype solution is in https://github.com/PyLops/pylops/pull/617/commits/fbdbcfbd759d294af6269d06709dd2fcf1a35470 with an example use case in https://github.com/PyLops/pylops_notebooks/blob/master/developement-cupy/Kirchhoff.ipynb

<img width="932" alt="Screen Shot 2024-11-08 at 2 29 58 PM" src="https://github.com/user-attachments/assets/f0177f3e-fa3e-461e-8759-0df738a0774a">
<img width="964" alt="Screen Shot 2024-11-17 at 10 34 21 PM" src="https://github.com/user-attachments/assets/5bef9487-69de-45c0-9589-54d82cd85a33">
<img width="965" alt="Screen Shot 2024-11-17 at 10 34 24 PM" src="https://github.com/user-attachments/assets/afccdb38-734d-484f-8957-7ad85a059584">

cako:
This is a very good idea :)

PR #622, opened 2024-11-18
feat: ToCupy operator

mrava87:
Addresses https://github.com/PyLops/pylops/issues/620


mrava87:
@cako I'd like to hear your opinion on this before implementing the changes to all the other solvers.

Basically when I started to work on this I wasn't sure how much effort it would be required to have these sort of mixed CPU-GPU operators (and inverse problems). It turned out to be easier than expected, but it required some modifications to he solvers (hidden from the user) and the introduction of a new optional parameter for the stacking operators to be able to handle case 4 in https://github.com/PyLops/pylops/issues/620. To handle case 3, it is instead simply required to have this new `ToCupy` operator...

docs/source/gpu.rst
```diff
@@ -450,15 +466,90 @@ your GPU/TPU:
...
    xadj = Gop.rmatvecad(x, y)
 
+Again, the code is almost unchanged apart from the fact that we now use ``jax`` arrays.
+
+
+Mixed CPU-GPU powered inverse problems
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Let us now consider a more intricate scenario where we have acess to 
+a GPU-powered operator, however the model and/or data vectors are too large
```
cako, line 486: I would be specific here and say GPU memory or VRAM

docs/source/gpu.rst
```diff
@@ -450,15 +466,90 @@ your GPU/TPU:
...
+Mixed CPU-GPU powered inverse problems
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Let us now consider a more intricate scenario where we have acess to 
+a GPU-powered operator, however the model and/or data vectors are too large 
+to fit onto the memory. 
+
+For the sake of clarity, we consider a problem where 
+the operator can be written as a :class:`pylops.BlockDiag` of
+PyLops operators. Note how, by simply sandwitching any of the GPU-powered
```
cako, line 491: sandwiching

docs/source/gpu.rst
```diff
@@ -409,23 +409,36 @@ Geophysical subsurface characterization:
    when ``explicit=False``.
 
 
-Example
--------
+Examples
```
cako, line 412: I think we should move this section to be after "Overview" and have a separate section "Supported Operators" the full tables. Otherwise this section will be lost to the reader

docs/source/gpu.rst
```diff
@@ -450,15 +466,90 @@ your GPU/TPU:
...
+
+   # Create operator
+   Mops = []
+   for iop in range(nops):
+      Mop = MatrixMult(cp.asarray(Ms[iop], dtype=dtype))
+      Top = ToCupy(Mop.dims, dtype=dtype)
+      Top1 = ToCupy(Mop.dimsd, dtype=dtype)
+      Mop = Top1.H @ Mop @ Top
+      Mops.append(Mop)
+   Mops = BlockDiag(Mops, forceflat=True)
```
cako, line 513: I think a diagram showing all the operations and where they reside would be really useful. Something like this:
to_cupy_diagram drawio

(I can share the draw.io xml if you want).

cako:
Looks really good! The only part I'd say could be improved is the documentation.

docs/source/gpu.rst
mrava87: Thanks! I agree, it would be nice to have a few diagrams like this to explain the different behaviors 😊

Sure, can you please share it

docs/source/gpu.rst
mrava87: Good point :smile:

mrava87:
@cako what about now? I added 3 graphics for end-to-end GPU, and the 2 mixed scenarios 😄

PR #623, opened 2024-11-20
doc: added info to install pytensor and pymc

mrava87:


mrava87:
@cako just realized we forgot to add `pytensor` and `pymc` to the installation instructions in the doc...

cako (approved):

PR #626, opened 2024-12-04
Add button to GitHub Repo

prisae:
I personally dislike it a lot when I come to the documentation of a source code, but there is no easy/quick link to the code repo, which means I have to google to find the repo ;-)

Adding the `"github_url"` to the `"html_theme_options"` should add a GitHub-link-button on the top-right corner, next to the switch for light/dark theme.

I haven't built the PyLops-docs to check it, but this is how it looks on the empymod website:
GitHub-Button

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -124,6 +124,7 @@ html_show_copyright = True
 # Theme config
 html_theme = "pydata_sphinx_theme"
 html_theme_options = {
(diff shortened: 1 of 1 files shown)


mrava87:
Nice addition 😄  Our CI builds the doc so you don't need to do it locally, and it seems to have worked: https://pylops--626.org.readthedocs.build/en/626/

I am going to merge this.

PR #631, opened 2024-12-13
Bug: fix sampling when not given

AmirMardan:
For trace interpolation, it is necessary to define either `sampling` or both `taxis` and `spataxis`. In case of using the latter, the function `SeismicInterpolation` is supposed to calculate the sampling. However, the calculated `sampling` was always `[0, 0]`. 

This PR fixes this problem.

### DIFF
--- a/pylops/waveeqprocessing/seismicinterpolation.py
+++ b/pylops/waveeqprocessing/seismicinterpolation.py
@@ -272,8 +272,8 @@ def SeismicInterpolation(
                     )
                 else:
                     sampling = (
-                        np.abs(spataxis[1] - spataxis[1]),
-                        np.abs(taxis[1] - taxis[1]),
+                        np.abs(spataxis[1] - spataxis[0]),
+                        np.abs(taxis[1] - taxis[0]),
                     )
             Pop = FFT2D(dims=dims, nffts=nffts, sampling=sampling)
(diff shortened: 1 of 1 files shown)


AmirMardan:
Please note that this modification is applied only for 2D problems. It seems, the same modification should be applied to 3D, [here](https://github.com/PyLops/pylops/blob/64de4796222d682983ec5467318d8e1323a82af8/pylops/waveeqprocessing/seismicinterpolation.py#L259C1-L263C22)

mrava87:
@AmirMardan oh yeah, good spot 😄 it should have been 

```
 sampling = (
                        np.abs(spataxis[1] - spataxis[0]),
                        np.abs(taxis[1] - taxis[0]),
                    )
```

but since these are pre-computed above, your fix is actually better. Do you mind push the same change for the 3D case before I merge the PR?

AmirMardan:
@mrava87  It's done

> @AmirMardan oh yeah, good spot 😄 it should have been
> 
> ```
>  sampling = (
>                         np.abs(spataxis[1] - spataxis[0]),
>                         np.abs(taxis[1] - taxis[0]),
>                     )
> ```
> 
> but since these are pre-computed above, your fix is actually better. Do you mind push the same change for the 3D case before I merge the PR?

PR #638, opened 2025-01-31
Added missing `iiter` and `xnorm` to the documentation of `pylops.lsqr`

### LINKED ISSUE #637: [Documentation] `pylops.lsqr` documentation lists 9 results even though there is 11
While looking at [`pylops.lsqr`](https://pylops.readthedocs.io/en/stable/api/generated/pylops.optimization.basic.lsqr.html) to find out what's wrong with [this `scipy` bug](https://github.com/scipy/scipy/issues/22232), I found a minor issue with the documentation of the returns.

According to the docs, there are 9 results returned:

Image

However, when I run the following, it reveals that there are 11 returns:

```python
import numpy as np
from scipy.sparse import eye as speye
from scipy.sparse.linalg import aslinearoperator
from pylops import LinearOperator, lsqr

A = aslinearoperator(speye(10, 10, format="csr"))
b = np.ones(A.shape[0])

# solving with lsqr
lsqr_result = lsqr(LinearOperator(A), b, calc_var=True)

print(len(lsqr_result))  # 11
```

Looking at the code,

Image

the documentation is missing
- `iiter` as the 3-rd output after `istop` (index 2)
- `xnorm` as the 9-th output after `arnorm`  (index 8)

IruNikZe:
The documentation of [pylops.lsqr](https://pylops.readthedocs.io/en/stable/api/generated/pylops.optimization.basic.lsqr.html) was missing the returns of `iiter` and `xnorm`. 
They were added to the docs of all related functions and methods.

For `iiter` the same phrasing as for [`scipy.sparse.linalg.lsqr`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.lsqr.html) was used.
For `xnorm`, the mathematical style of `pylops` was adopted.

Closes #637

### DIFF
--- a/pylops/optimization/basic.py
+++ b/pylops/optimization/basic.py
@@ -242,6 +242,8 @@ def lsqr(
 
         ``7`` means the iteration limit has been reached
 
+    iiter : :obj:`int`
+        Iteration number upon termination
     r1norm : :obj:`float`
         :math:`||\mathbf{r}||_2^2`, where
         :math:`\mathbf{r} = \mathbf{y} - \mathbf{Op}\,\mathbf{x}`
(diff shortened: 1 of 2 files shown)


mrava87:
@IruNikZe thanks for noticing this, the PR looks great. I approve it :)

Issue #603, opened 2024-08-16
Enable caching in other methods of AcousticWave2D


mrava87:
Following https://github.com/PyLops/pylops/pull/594, we should consider making the same changes to:

- `_srcillumination_oneshot` and `_srcillumination_allshots`
- ` _bornadj_oneshot` and ` _bornadj_allshots`

PR #650, opened 2025-02-27
Removes the creation of the solver from within the shots iteration loop.

guaacoelho:
Hi!

This pull request is related to #603

### DIFF
--- a/pylops/waveeqprocessing/twoway.py
+++ b/pylops/waveeqprocessing/twoway.py
@@ -250,7 +250,7 @@ class AcousticWave2D(LinearOperator):
 
         # assign source location to source object with custom wavelet
         if hasattr(self, "wav"):
-            self.wav.coordinates.data[0, :] = self.geometry.src_positions[isrc, :]
+            self.wav.coordinates.data[0, :] = solver.geometry.src_positions[:]
 
         # source wavefield
         u0 = solver.forward(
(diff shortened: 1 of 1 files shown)


pylops/waveeqprocessing/twoway.py
```diff
@@ -245,17 +247,6 @@ def _srcillumination_oneshot(self, isrc: int) -> Tuple[NDArray, NDArray]:
...
-            self.geometry.src_positions[isrc, :],
-            self.geometry.t0,
-            self.geometry.tn,
-            f0=self.geometry.f0,
-            src_type=self.geometry.src_type,
-        )
-        solver = AcousticWaveSolver(self.model, geometry, space_order=self.space_order)
 
         # assign source location to source object with custom wavelet
         if hasattr(self, "wav"):
```
mrava87, line 252: To be consistent with the `_born_oneshot` method, shall we make this 

```
self.wav.coordinates.data[0, :] = solver.geometry.src_positions[:]
```

?

pylops/waveeqprocessing/twoway.py
```diff
@@ -375,22 +382,10 @@ def _bornadj_oneshot(self, isrc, dobs):
...
-        )
         # create boundary data
         recs = self.geometry.rec.copy()
         recs.data[:] = dobs.T[:]
 
-        solver = AcousticWaveSolver(self.model, geometry, space_order=self.space_order)
-
         # assign source location to source object with custom wavelet
         if hasattr(self, "wav"):
             self.wav.coordinates.data[0, :] = self.geometry.src_positions[isrc, :]
```
mrava87, line 391: Same here?

mrava87 (changes requested):
@guaacoelho thanks, this looks great!

I just left two small comments, other than that I think this is ready to be merged 😄

pylops/waveeqprocessing/twoway.py
guaacoelho: You're right! I'll make the changes and update the PR.

PR #681, opened 2025-07-08
Support JAX backend in ChirpRadon2D/3D by removing in-place assignment.

GeophyAI:
This PR replaces in-place assignments in ChirpRadon2D with `pylops.backend.inplace_set`, enabling compatibility with JAX arrays.


mrava87:
Thanks @GeophyAI

I have just pushed a similar change for `ChirpRadon3D` and checked they both work and produce same results to the numpy backend. Happy to merge this 😄

PR #684, opened 2025-07-10
build: disable dtcwt example and test to be able to support numpy v2 in CI and doc

mrava87:
This PR is primarily focused on removing `dtcwt` from all environment/requirement files as by not supporting numpy v2 it was making it very hard to have a working local environment on ARM MacOSX (and like this is going to be the same soon for Linux).

Similarly, a skip `@pytest.mark.skipif(int(np_version[0]) >= 2, reason="dtcwt does not support numpy v2")` is added to the `test_dtcwt`, currently disabling any test for this operator.

Finally RTD is now using python 3.11 and numpy v2 and the `plot_dtcwt` example is currently disabled (we still keep the file in the repository in case `dtcwt` starts to support numpy v2, so we will be able to bring it back...).

Additionally, all requirements/environments files have been slightly revamped with the following changes:
- `python>=3.11` is now the new default in environment files 
- `numpy>=2.0.0` and `scipy>=1.13.0` (the first version supporting numpy2) are now the new default in environment/requirement files
- `torch` is now installed via pip and not conda in environment files as this is now the reccomended way (https://pytorch.org/get-started/locally/)
- `Jax` is also now installed via pip and not conda in environment files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ docs/build
 docs/source/api/generated
 docs/source/gallery
 docs/source/tutorials
+docs/source/sg_execution_times.rst
 
 # Pylint #
 pylint_plot.py
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -9,7 +9,7 @@ version: 2
 build:
   os: ubuntu-20.04
   tools:
-    python: "3.9"
+    python: "3.11"
 
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -60,14 +60,16 @@ numpydoc_show_inherited_class_members = False
 numpydoc_class_members_toctree = False
 
 sphinx_gallery_conf = {
-    # path to your examples scripts
+    # Path to examples scripts
     "examples_dirs": [
         "../../examples",
         "../../tutorials",
     ],
-    # path where to save gallery generated examples
+    # Path where to save gallery generated examples
     "gallery_dirs": ["gallery", "tutorials"],
     "filename_pattern": r"\.py",
+    # Examples to skip when building the gallery
+    "ignore_pattern": r"plot_dtcwt\.py$",
     # Remove the "Download all examples" button from the top level gallery
     "download_all_examples": False,
     # Sort gallery example by file name instead of number of lines (default)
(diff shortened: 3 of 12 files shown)


mrava87:
@cako I noticed that we have the following note in the Installation section for PyTensor:

```
PyTensor does not support NumPy 2 yet, so make sure you use NumPy 1.x with PyTensor and PyMC.
```

From a quick look at the documentation and changelogs of PyTensor and PyMC it seems to me that this comment is outdated and they have started supporting NumPy 2, do you agree? If so I am going to take this out?

And do you agree in general with my changes (we did not pay attention to the environments for long time so I felt things were starting to get a bit outdated 😉 )

cako:
@mrava87 Agreed! Pytensor supports numpy 2 since [2.28](https://github.com/pymc-devs/pytensor/releases/tag/rel-2.28.0) and PyMC since [5.21.0](https://github.com/pymc-devs/pymc/releases/tag/v5.21.0). So I think as long as we bump the versions it should be fine!

mrava87:
@cako perfect! Then let me remove the comment about numpy v1 add these as the minimum versions in requirement/environment files 😀

mrava87:
For the record: seems like that if I change the version of `pytensor` and  `pymc` to `pytensor>=2.28.0`
`pymc>=5.21.0` in the environment files, `devito` produces a strange compilation error in `plot_twoway` in my local ARM environment.

```
self = JITCompiler[CustomCompiler], ext_file = '/var/folders/k2/5r6ztqx10jscz2r52gj5ry2c0000gn/T/devito-jitcache-uid501/790839b6ddf8bcf29121cdd8508fbe03d7782ca4.dylib'
source_files = ['/var/folders/k2/5r6ztqx10jscz2r52gj5ry2c0000gn/T/devito-jitcache-uid501/790839b6ddf8bcf29121cdd8508fbe03d7782ca4.c'], debug = False

    def build_extension(self, ext_file, source_files, debug=False):
        cc_cmdline = (
                self._cmdline(source_files, False)
                + ["-o", ext_file]
                )
    
        from pytools.prefork import call
        if debug:
            print(" ".join(cc_cmdline))
    
        result = call(cc_cmdline)
    
        if result != 0:
            import sys
            print("FAILED compiler invocation: {}".format(" ".join(cc_cmdline)),
                  file=sys.stderr)
>           raise CompileError("module compilation failed")
E           codepy.CompileError: module compilation failed

/opt/miniconda3/envs/pylops/lib/python3.13/site-packages/codepy/toolchain.py:211: CompileError
```

For now, I revert back the environment files to `pymc>=5` and `pytensor` as this seems to work fine.

PR #685, opened 2025-07-15
test: reinstate TorchOperator tests on macosx

mrava87:


mrava87:
I still get the same error `RuntimeError: Numpy is not available`... torch2 does not seem to work with numpy2 on macosx GA.

PR #687, opened 2025-07-21
feat: improved used of logging library

mrava87:
This PR is aimed at improving the use of the `logging` library. 

The most notable change is the removal of all `logging.basicConfig` (which should be left to the user application) and the use of `logger = logging.getLogger(__name__)` instead. 

Some of the `logging.error` messages are also removed (as accompanied by a raise exception) and some `logging.warning` converted to `warnings.warn` based on the definition of how these two ways of raising a warning compare in https://docs.python.org/3/howto/logging.html)

### DIFF
--- a/pylops/__init__.py
+++ b/pylops/__init__.py
@@ -45,6 +45,8 @@ utils
 
 """
 
+import logging
+
 from .config import *
 from .linearoperator import *
 from .torchoperator import *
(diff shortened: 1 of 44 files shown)


mrava87:
@cako not sure how familiar you are with the `logging` library... as I was learning more I realized our use wasn't that great so I decided to go through a full re-style. Let me know if you have any feedback before I can go ahead and merge this 😄

PR #694, opened 2025-08-30
Add Intel `mkl_fft` support for fft module and GitHub Action with Intel Python

### LINKED ISSUE #386: Support Intel MKL FFT
## Description

A conda installation of numpy provides MKL-accelerated linear algebra subroutines (BLAS) but does not substitute FFTs for
those from the Intel MKL FFT. As far as I have been able to establish, there are three ways of installing MKL FFTs: 1) through the [`intel-numpy` package](https://pypi.org/project/intel-numpy/) which substitutes `numpy.fft` for MKL FFTs; 2) through [`intel-scipy`](https://pypi.org/project/intel-scipy/) which does the same for `scipy.fft`; 3) through the standalone [`mkl_fft` package](https://github.com/IntelPython/mkl_fft).

It would be interesting to ensure that MKL FFT is supported through all these formats.

## Definition of done
* Test and possibly fix current implementation of FFTs but using `intel-numpy` and `intel-scipy`. ~There is a known issue that breaks PyLops when using these libraries (https://github.com/IntelPython/mkl_fft/pull/69). This issue is reported upstream, but can be bypassed downstream.~
* Implement a new FFT backend (like `pyfftw`) which relies on the standalone `mkl_fft`.

rohanbabbar04:
Closes #386 
- Add new `engine=mkl_fft` option for Fourier transforms (FFT) which uses the [mkl_fft](https://github.com/IntelPython/mkl_fft) package.
- Add tests and an example using this engine.


mrava87:
Nice @rohanbabbar04 😄 I'll take a look soon, @cako would be great if you can do the same.

PS: one thing I noticed quickly skimming through the files changed is that we should also modify https://github.com/PyLops/pylops/blob/dev/docs/source/installation.rst to include `mkl_fft`.

requirements-dev.txt
```diff
@@ -30,3 +30,4 @@ black
 flake8
 mypy
 pytensor>=2.28.0
+mkl-fft; sys_platform != "darwin"
```
mrava87: What does this mean? 

If I look at the documentation of `mkl_fft` I see that they suggest that to install the library with `pip` one has to do 

```
   python -m pip install --index-url https://software.repos.intel.com/python/pypi --extra-index-url https://pypi.org/simple mkl_fft
```

but you seem to install it from the standard url... is it because of this: https://pypi.org/project/mkl-fft/?

And why not on mac? I guess the macos tests run fin as `mkl_fft` is not installed and so it falls back to numpy?

mrava87 (commented):
@rohanbabbar04 I do not seem to be able to install `mkl_fft` on macOS? Seeing also your change in the requirements file it makes me suspect it is only available on linux platforms?

Does that make sense? If so I will review this on my other laptop but we should state that clearly in the `installation.rst` file and should not run tests on macOS when `"engine": "mkl_fft"` as otherwise it will just run twice the tests with the `numpy` engine...

requirements-dev.txt
rohanbabbar04: > What does this mean?
> 
> If I look at the documentation of `mkl_fft` I see that they suggest that to install the library with `pip` one has to do
> 
> ```
>    python -m pip install --index-url https://software.repos.intel.com/python/pypi --extra-index-url https://pypi.org/simple mkl_fft
> ```
> 
> but you seem to install it from the standard url... is it because of this: https://pypi.org/project/mkl-fft/?
> 
> And why not on mac? I guess the macos tests run fin as `mkl_fft` is not installed and so it falls back to numpy?

This is just a check to avoid installing mkl_fft on macOS, as mkl_fft, starting from, I believe, April [no longer supports macOs](https://github.com/IntelPython/mkl_fft/pull/147).

We can update the requirements to include `--index-url` and `--extra-index-url`. I have also updated the tests to skip any that involve `mkl_fft` on macOS.

requirements-dev.txt
```diff
@@ -1,3 +1,6 @@
+--index-url https://software.repos.intel.com/python/pypi
```
mrava87: mmh I think this may not work as you may expect as it would always favor the intel repo over the official pypi for every package in the file (which I am not in favor of).... what we did for torch (after realizing that) was to make an additional `requirements-torch` file where there we switch the index just to install `torch` and then we do (eg taken from Makefile)

```
pip install -r requirements-dev.txt && pip install -r requirements-torch.txt && $(PIP) install -e .
```

We could probably do something similar here and maybe provide make targets for intel where we install an intel flavored environment (see below for more on this)?

mrava87: The only thing I am a bit worried about / don't fully understand is this comment in the README of mkl_fft

```
If command above installs NumPy package from the PyPI, please use following command to install Intel optimized NumPy wheel package from Intel PyPI Cloud
```

So does this mean we need their NumPy as it won't work with the one from the official PyPI. 

If so, I think we need a more targeted strategy where we provide files to create an environment with `mkl` where we use the Intel registry and we can also add a build GA to test it... I do not want to have our CI using the Numpy distributed by Intel as default as 99% of our users are more likely to use that from PyPI so I want that to be the one we run test against 😉

docs/source/installation.rst
```diff
@@ -358,11 +358,11 @@ Install it via ``pip`` with
 
 FFTW
```
mrava87: I would change this into `FFTW and MKL-FFT`

environment-dev.yml
```diff
@@ -1,5 +1,6 @@
 name: pylops
 channels:
+  - https://software.repos.intel.com/python/conda
```
mrava87: Same as above, would this also install intel `numpy` over the one from coda-forge? 

If so, I feel this may be too strong of a change as now `Intel` takes priority over anything and we should maybe provide an environment specific for Intel?

mrava87:
@rohanbabbar04 thanks for the explanation and changes!

I left a few more comments as I am not sure exactly how invasive is the use of Intel conda channel and Intel PyPI cloud... hopefully you have a better grasp than me on this and can comment on.

Regarding the code, I see that you switched from `mkl_fft.interfaces.scipy_fft` to `mkl_fft.interfaces.numpy_fft` and that led to removing a lot of the type casting. Again can you explain?

pytests/test_ffts.py
```diff
@@ -267,6 +323,8 @@ def test_unknown_engine(par):
 
 @pytest.mark.parametrize("par", pars_fft_small_real)
 def test_FFT_small_real(par):
+    if par["engine"] == "mkl_fft" and sys.platform == "darwin":
```
mrava87: I think it should be "Darwin" as in https://github.com/PyLops/pylops/blob/cbd7d1bf4b5937992e64048d8e813c9fdad3d84d/pytests/test_torchoperator.py#L35

I tried switching to "darwin" but it won't go into the condition... in fact on my macOS if I do

```
In [1]: import platform

In [2]: platform.system()
Out[2]: 'Darwin'
```

requirements-dev.txt
rohanbabbar04: > mmh I think this may not work as you may expect as it would always favor the intel repo over the official pypi for every package in the file (which I am not in favor of).... what we did for torch (after realizing that) was to make an additional `requirements-torch` file where there we switch the index just to install `torch` and then we do (eg taken from Makefile)
> 
> ```
> pip install -r requirements-dev.txt && pip install -r requirements-torch.txt && $(PIP) install -e .
> ```
> 
> We could probably do something similar here and maybe provide make targets for intel where we install an intel flavored environment (see below for more on this)?

Agreed, it’s better to create another requirements file, which I just added in my latest commits — `requirements-intel-mkl.txt`. Even better, we can define targets that call `pip`, for example:
```
pip install -r requirements-intel-mkl.txt && pip install -r requirements-dev.txt && pip install -r requirements-torch.txt && pip install .
``` 

Similarly, for `conda`, it’s better to create a new file `environment-dev-intel-mkl.yml`, giving priority to `software.intel.repos`. This way, Python, NumPy, SciPy, and MKL can be installed directly from the Intel repository for maximum performance.

rohanbabbar04: > The only thing I am a bit worried about / don't fully understand is this comment in the README of mkl_fft
> 
> ```
> If command above installs NumPy package from the PyPI, please use following command to install Intel optimized NumPy wheel package from Intel PyPI Cloud
> ```
> 
> So does this mean we need their NumPy as it won't work with the one from the official PyPI.
> 
> If so, I think we need a more targeted strategy where we provide files to create an environment with `mkl` where we use the Intel registry and we can also add a build GA to test it... I do not want to have our CI using the Numpy distributed by Intel as default as 99% of our users are more likely to use that from PyPI so I want that to be the one we run test against 😉

By default, NumPy is built with OpenBLAS, not MKL. For maximum performance, it’s better to ensure NumPy is configured with MKL (although it will still work with OpenBLAS, just not at the best performance).

As you suggested, the best approach would be to create a new GA workflow specifically for Intel MKL. In this workflow, we can set up an Intel Python environment and install all required libraries inside it. Ideally, we should create a `conda` environment with Python 3.10, 3.11, and 3.12 (from Intel’s software repos), install the necessary packages there, and run the pytest suite.

I’ve already implemented this in my latest commits by adding a new workflow file, build-mkl.yml, which performs these steps.

pytests/test_ffts.py
rohanbabbar04: Since we now have a new GA, it’s probably better to run it on `ubuntu-latest` (since macOS is not supported by the Intel software repositories). We could also skip the tests when `mkl_fft` is not enabled. For the normal GA checks, I think it’s fine to just skip `mkl_fft`. Do you have any suggestions for a better way to skip these tests in our normal GA, other than the approach I’ve taken in the tests?

rohanbabbar04:
One more thing to note is `pyFFTW`, which sometimes does not behave correctly when installed with pip alongside NumPy + MKL. I came across [this PR](https://github.com/pyFFTW/pyFFTW/pull/116), which seems to provide a solution. Another good option is to use the `conda-forge`, which is updated and works reliably with NumPy + MKL. I’ve mentioned this in the `installation.rst` so that users are aware of it.

mrava87:
@rohanbabbar04 thanks, this already looks much better and less intrusive to the basic NumPy+SciPy installations that we 
want to support as default 🚀 

I'll take a more detailed look during the week and provide any further feedback. In the meantime, @cako would you be able to review this PR as well, since you have actually done more work than me in the `FFT*` operators so it would be good to get your approval on these updates 😄

rohanbabbar04:
@mrava87 @cako Just checking if you got some time to review this PR 😅

mrava87:
@rohanbabbar04 sorry not yet on my side… will try to do it at the weekend 😀

But I won’t merge this until I hear from @cako as well…

pytests/test_ffts.py
mrava87: No I think what you did is the best way, since we have the `mkl_fft_enabled` it is a good idea to use it 😄

mrava87 (approved):
@rohanbabbar04 I finally had time to look at your PR in more details, and I think it looks great 😄 everything seems to be aligned with the other FFT implementations and the changes to environment/requirements files and the new GA are perfect.

I'll wait a couple of days, if I do not hear from @cako, I'll go ahead and merge

cako:
Sorry guys, was on holiday and then back from a conference and then ill. I will take a look at this today!

cako (commented):
This looks great, thanks @rohanbabbar04!! I have a few questions:

1. This was mentioned in the new docs, but just to confirm, my FFTW tests are not passing when I install `requirements-intel-mkl.txt`.
2. I added a "Supported backends" section to the `plot_fft.py` example, please make sure it's correct!
3. My understanding is that installing `intel-numpy`/`intel-scipy` hijacks the `numpy.fft` and `scipy.fft` libraries to under the MKL FFTs instead. Therefore, the way the installation currently works is that both `intel-numpy` and `intel-scipy` will be installed alongside `mkl_fft`, this means that **every backend except FFTW** (which will incidentally also be broken) will actually be using MKL. This is mentioned under the BLAS section of the Advanced Installation, but I think it bears repeating under the MKL-FFT section.
4. If we want to make the MKL installation **only** install `mkl_fft` and not `intel-numpy`/`intel-scipy`, then we need to make 3 changes: 1) install dev requirements before mkl requirements, 2) use `numpy>=2.0.0,<=2.2.5` in the dev requirements because otherwise `mkl_fft` will downgrade to numpy==2.2.5 and in the process, replace it with the intel version. I don't mind either solution, but I think it needs to be made very clear in the docs what's going on.
5. I made some small annotation changes, please review them!

mrava87:
Thanks @cako!

For point 1, I see that pyfftw is in that requirement file too, but I see that in the action it is installed first with conda… but I think we should avoid too many exceptions, to avoid confusion (we, ourselves, will forget these little details with time). If the Make target dev-install_intel_mkl leads to the pyfftw tests not running / passing, we should either say that for intel_mkl we provide only a conda recipe or avoid installing pyfftw and maybe let FFT fallback to numpy?

For 3,4, I let @rohanbabbar04 reply but in general I agree with you… repeating always help 😝

cako (approved):

rohanbabbar04:
Thanks @cako for your review.

1)Yes, this may be caused by pyFFTW, which creates issues when installed with pip, since pyFFTW does not work with MKL-backed NumPy. An alternative suggestion is already mentioned in the docs and can also be found [here](https://github.com/pyFFTW/pyFFTW/pull/116). @mrava87 I think we should just remove it from `requirements-dev`, keep `pyFFTW` in `requirements-pyfftw`, and let FFT fall back to NumPy when people install via pip. This way, the FFT routines and the computation backend for everything else will use MKL, and users should still be able to work with it. I made my latest change on this, do you think this is the best way to tackle this?

3)@cako, you are almost right. Having NumPy and SciPy installed from the Intel repositories allows not only FFT routines but also linear algebra operations to be carried out using MKL, which provides better performance. I think repeating this inside mkl-fft makes sense, along with a warning about pyFFTW (since it breaks with MKL-backed NumPy when installed using pip).

4)You are right. We can do that, and it works without issues. However, this would create an environment where NumPy uses its default backend, while MKL FFT uses the MKL backend. To achieve maximum performance, we should have `NumPy`, `SciPy`, and `mkl-fft` all installed from the Intel repositories.

mrava87 (approved):
This also looks good to me now 😄 I am just going to push a few minor changes in the example and merge it.

PR #697, opened 2025-09-11
Feat: add support for CT using ASTRA Toolbox

askorikov:
Roadmap:
- [ ] Fix macOS build or decide for an alternative
- [ ] Add 3D operator


askorikov:
One potential fix is to put ASTRA into the CuPy workflow, because they have the same CUDA requirement unsatisfiable on ARM macOS.

mrava87:
Thank @askorikov, I am going to take a look at this next week as soon as I can.

Just to be sure I understand what you wrote above: everything works for Linux but not for Mac? Ideally it would be of course better to support both Linux and Mac but we had some other scenarios where things don't play nice on Mac, hopefully just temporarily but maybe also for long time. If this is the case we could just add a skip to the test if running on Mac (https://github.com/PyLops/pylops/blob/0396fff9b837d77cb967a76c22418696e2e34e1e/pytests/test_torchoperator.py#L34) and add a note in the `installation.rst` file saying that to work on Mac this or that is required 😄 ]

If instead the entire build process breaks on Mac, because of trying to install `astra-toolbox`, I am also happy to consider splitting the test process in two jobs, one for linux and one for Mac and have different requirement files as the solution

askorikov:
At the moment, the entire build will not work, unfortunately... The main problem is, on mac arm we can only provide a CPU-only version of ASTRA. It actually already exists on conda-forge as a build variant of ASTRA package, but I'm not quite sure sow to do it with pip. We could probably distribute an `astra-toolbox-cpu` package there, but then the workflow (for building PyLops) would anyway split between CUDA and CPU-only version.

mrava87:
Alright, I see.

Well, let me try to push something to try to see how easy is to just have two different requirements for arm and Linux like we have for the conda environment and use them in the CIs.

(In general I always found it hard to have the exact same build process for CPU and GPU whatever I do, PyLops and anything DL/PyTorch related, so the name difference won’t be a major problem but I understand if you don’t wanna put effort on mac arm as it isn’t really the primarily platform for scientific computing anyways 😝)

mrava87:
@askorikov: Ok, my changes seem to streamline the CIs avoiding errors in the OSX systems (by not installing astra-toolbox and skipping tests).

At this point, I think this could go in as first minimal PR. But:
- if you see room for improvement for the `CT2D` and feel like, please go ahead, I am happy to keep this here waiting until you are happy
- adding a 3D operator would be great, but I have no time at the moment and it's a long time I don't touch ASTRA so I won't be able to do it right away... if you want please go ahead. 
- since ASTRA works for GPUs, if you think you could make our this operator work we could avoid skipping the tests in CuPy CI. For now I wasn't sure it/how to do it, so the ``@pytest.mark.skipif(int(os.environ.get("TEST_CUPY_PYLOPS", 0)) == 1, reason="Not CuPy enabled")`` added to the tests

askorikov:
Cool! I will be looking at this in the next few days.

mrava87:
@askorikov checking if you had time to take a look at this?

askorikov:
Hi Matteo! Do you have on opinion on when stuff should run on GPU? In principle, Astra primarily focuses on GPU accelerated operations that are currently unused. The speed comparison is something like milliseconds (GPU) vs hundreds of milliseconds (CPU) for a 256 x 256 image. Given that, I see the following options for the CT2D operator:

1. Make GPU the default option (with a disadvantage of excluding CPU-only users).
2. Make it an option to chose GPU or CPU.
3. Make it run on CPU for NumPy-based PyLops and on GPU for the CuPy version.

For the 3D version, we only have GPU operators available in Astra, since CPU would be prohibitively slow for the typically large 3D data.

mrava87:
> Hi Matteo! Do you have on opinion on when stuff should run on GPU? In principle, Astra primarily focuses on GPU accelerated operations that are currently unused. The speed comparison is something like milliseconds (GPU) vs hundreds of milliseconds (CPU) for a 256 x 256 image. Given that, I see the following options for the CT2D operator:
> 
> 1. Make GPU the default option (with a disadvantage of excluding CPU-only users).
> 2. Make it an option to chose GPU or CPU.
> 3. Make it run on CPU for NumPy-based PyLops and on GPU for the CuPy version.
> 
> For the 3D version, we only have GPU operators available in Astra, since CPU would be prohibitively slow for the typically large 3D data.

@askorikov right, our approach when we bring it external libraries is to provide GPU support if the library supports it. So yes, I am in total favor of having a double backend for the `CT2D` (and `CT3D`) operator. Having said that, we prefer not to have black-box kind of operators that take numpy arrays, convert internally, run operator on GPU and convert back. This is to avoid unexpected behaviors and also because if one chains an operator like `CT2D` to another PyLops operator that has a GPU backend, one would like to avoid any redundant host-2-device and device-2-host copies...

So my suggestion would be option 2 or 3, depending on when you need to know that the operator would run on the GPU (I haven't dug deep enough into Astra to know the answer 😉 ). If you need to know it in the `__init__` of  `CT2D` then we would add a `engine` parameter like in for example  in https://github.com/PyLops/pylops/blob/0396fff9b837d77cb967a76c22418696e2e34e1e/pylops/signalprocessing/nonstatconvolve2d.py#L59. However, if this can be deferred to the `_ matvec/_rmatvec` then we can just use the strategy we use in all our numpy/cupy operators, like in https://github.com/PyLops/pylops/blob/0396fff9b837d77cb967a76c22418696e2e34e1e/pylops/basicoperators/pad.py#L89. Which one do we need?

And for 3D, I do not see a problem to have an operator that has only GPU support as long as we document it here https://github.com/PyLops/pylops/blob/dev/docs/source/gpu.rst (in the `Supported Operator` section). It would be a first time, but there is always a first time 😸

mrava87:
@askorikov ?

mrava87:
@askorikov as discussed offline I am going to merge this and I just created an issue to capture the further developments that we discussed here https://github.com/PyLops/pylops/issues/701

mrava87 (approved):

PR #699, opened 2025-10-03
Feat: integers in stacking operators

mrava87:
## Motivation

PyLops' stacking operators allow one to pass PyLops multiple LinearOperators or numpy/scipy matrices. However, in many practical scenarios, especially for the `Block` Operator, one may have some zero blocks (either many small ones or few large ones). Currently, one must rely on `pylops.Zero` to fill those zero blocks. However the inner working of `pylops.Zero` is such that every time matvec/rmatvec is called a zero array is instantiated. For stacking operators, this is actually redundant as one could simply skip filling part of the output vector. The current approach is particularly problematic when using the CuPy backend as instantiating new arrays on GPU memory is tremendously slow and unefficient (compared to performing arithmetic operations).

To solve this issue, this PR proposes to allow passing integer numbers (alongside LinearOperators or numpy/scipy matrices), which represent the number of zero columns (in `HStack` and `Block`) and the number of zero rows (in `VStack`). By doing so, matvec/rmatvec simply do not perform any operation for the indices of the `ops` list that contain `int` values.

## Definition of Done

- [x] Modify `HStack`
- [x] Modify `VStack`
- [x] Modify `Block`
- [x] Add tests
- [x] Add examples 

⏰ https://github.com/PyLops/pylops/pull/702 seems to solve the same issue with much less code changes - and since `Zero` instantiation has pretty much no overhead, this PR seems to be an overkill...

### DIFF
--- a/examples/plot_stacking.py
+++ b/examples/plot_stacking.py
@@ -58,7 +58,7 @@ plt.tight_layout()
 plt.subplots_adjust(top=0.8)
 
 ###############################################################################
-# We now want to *vertically stack* three operators
+# We now want to *vertically stack* two operators
 #
 #    .. math::
 #       \mathbf{D_{Vstack}} =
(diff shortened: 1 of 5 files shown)


mrava87:
@cako would be great to get a feedback 😄

PR #700, opened 2025-10-08
Doc: add Attributes section

### LINKED ISSUE #488: Attributes field in docstrings of operators
# Motivation
Historically, every operator has a docstring containing an Attributes section with 2 fields: shape and explicit. This isn’t very informative and outdated as by now each linear operator has more ‘shared’ attributes (common to all operators) than those two. Moreover different operators have different attributes that may be useful to tell users about.

# Definition of done 
- Add an Attributes section to LinearOperator which details all of the attributes that are shared among operators (eg shape, dims, dimsd, explicit)
- Remove or modify the Attributes section in every operator. Remove if the operator does not define interesting attributes in the __init__ method, of modify removing shape and explicit and including those if it does.

mrava87:
Closes https://github.com/PyLops/pylops/issues/488.

This PR aims to update all Attributes sections of the docstrings of operators and solvers, including all of the internal attributes that we want users/devs to be aware of (previously we simply reported shape/explicit and missed many useful ones) - note that Parameters that are saved as class members are not added to the Attributes list, only new class members are.

Moreover, since the `Attributes` sections are now meaningful we make them appear in the Sphinx documentation.

Additional: 
- some of the attributes that I feel should be kept hidden from users (to prevent anyone from tinkering with them) are now private (ie added `_*`). Ideally I would have like to be more aggressive but I was worried about breaking backward compatibility for some attributes that users may print or check so I decided to be very conservative.
- all `np.ndarray` in docstrings have been switched to `numpy.ndarray` to be able to have hyperlinks pointing to the NumPy documentation.
- all :obj:`func` in the documentation are turned into :obj:`callable`
- fixed all outstanding warnings in the Sphinx documentation build


mrava87:
@cako this was long overdue and with some help with Copilot it took less than expected (I run out of the autocompletion allowance this months ahah).

I am not expecting you to look at this line by line, but let me know if you have any general suggestion 😄

cako (approved):
This is amazing!! Huge effort but will result in much better docs!! Thanks

PR #702, opened 2025-10-12
Feat: change internal handling of Zero in hstack/vstack

mrava87:
## Motivation

PyLops' stacking operators allow one to pass PyLops multiple LinearOperators or numpy/scipy matrices. However, in many practical scenarios, especially for the `Block` Operator, one may have some zero blocks (either many small ones or few large ones). Currently, one must rely on `pylops.Zero` to fill those zero blocks. However the inner working of `pylops.Zero` is such that every time matvec/rmatvec is called a zero array is instantiated. For stacking operators, this is actually redundant as one could simply skip filling part of the output vector. The current approach is particularly problematic when using the CuPy backend as instantiating new arrays on GPU memory is tremendously slow and unefficient (compared to performing arithmetic operations).

To solve this issue, this PR proposes to add a check prior to calling the matvec/rmatvec of an operator and skip it if the operator is of instance `Zero`

## Definition of Done

- [x] Modify `HStack`
- [x] Modify `VStack`
- [x] Add tests

⏰ This is a much simpler solution to the one initially proposed in https://github.com/PyLops/pylops/pull/699

### DIFF
--- a/examples/plot_stacking.py
+++ b/examples/plot_stacking.py
@@ -13,9 +13,10 @@ for the creation of complex operators as well as in the definition of various
 types of optimization problems with regularization or preconditioning.
 
 Some of this operators naturally lend to embarassingly parallel computations.
-Within PyLops we leverage the multiprocessing module to run multiple processes
-at the same time evaluating a subset of the operators involved in one of the
-stacking operations.
+Within PyLops we leverage the
+`Multiprocessing <https://docs.python.org/3/library/multiprocessing.html>`_
+module to run multiple processes at the same time evaluating a subset of the
+operators involved in one of the stacking operations.
 """
 import matplotlib.pyplot as plt
 import numpy as np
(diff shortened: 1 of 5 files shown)


cako (approved):
LGTM! I will leave it as approved but I suggest including this behavior in the class documentation and/or the gallery examples. It is important for people to know that if they use the proper operators (Zero as opposed to e.g., MatrixMult) they will get performance benefits.

Also, seems like no need for me to review #699?

mrava87:
Thanks!

Sure, let me add something to the example like I had done inhttps://github.com/PyLops/pylops/pull/699

And yes, no need to look at https://github.com/PyLops/pylops/pull/699... after I implemented that way I started to feel it was an overkill given the minimal overhead of `Zero` at initialization time, what we want to avoid is calling the matvec/rmatvec.

Issue #701, opened 2025-10-08
Extend medical submodule


mrava87:
This issue is mostly aimed at linking the conversation in this PR (https://github.com/PyLops/pylops/pull/697) for future development of the `medical` submodule.

The main directions for further development identified in the PR are:

- [ ] support for GPU backend in `CT2D`
- [ ] implementation of `CT3D`

aman-coder03:
Can I work on this?

PR #703, opened 2025-10-16
Update ASTRA integration

askorikov:
This PR adds support for CUDA backend of ASTRA, including the support for CuPy inputs (partially addressing #701). As you can see, the code becomes quite a bit more messy, but that is to cover several scenarios at the same time:

1. CPU input + CPU execution
2. CPU input + GPU execution (either via setting `engine`  or `projector_type` to `cuda`)
3. GPU input + GPU execution (including zero-copy GPU data exchange, which is only supported for 3D data in ASTRA at the moment, so we need to do a hack by representing 2D geometry as 3D with one slice).

With this, I have the following observations/questions:

1. The adjoint for the CUDA backend of ASTRA is much less matched, so the tests produce >10% relative error. We do expect the adjoint to be mismatched, but this error is still way too high. I will be looking at that on the ASTRA side in the coming time. In the meantime, the tests with CUDA backend are skipped.
2. Do we need to assert that the input is on CPU for `cpu` engine and on GPU for `cuda` engine?
3. In principle, this operator almost works with JAX on both CPU and GPU, except for ensuring devices appropriately. What is now the ambition for adding JAX support in PyLops operators?
4. The default `projector_type` now (`strip`) favors accuracy at the expense of performance. For experimental data the accuracy difference is usually not noticeable, so the faster `linear` projector is more commonly used. It's also the only projector available in CUDA backend. Let me know what you think fits the expected audience better.
5. ASTRA only supports float32 dtype internally at the moment. How do we go about it?


mrava87:
Thanks @askorikov!

Let me reply to the questions below and then I will do a more general review of the code.

> This PR adds support for CUDA backend of ASTRA, including the support for CuPy inputs (partially addressing #701). As you can see, the code becomes quite a bit more messy, but that is to cover several scenarios at the same time:
> 
> 1. CPU input + CPU execution
> 2. CPU input + GPU execution (either via setting `engine`  or `projector_type` to `cuda`)
> 3. GPU input + GPU execution (including zero-copy GPU data exchange, which is only supported for 3D data in ASTRA at the moment, so we need to do a hack by representing 2D geometry as 3D with one slice).

It is kind of expected that the code becomes a bit more complicated for operators having to handle multiple backend when it's not a pure numpy/cupy switch, so in principle I have no problem. 

So far PyLops' philosophy has been that we don't want operators to be magic black boxes so we want users to always provide numpy arrays when the operator works on CPU and cupy (or Jax) arrays for GPUs. In special cases where you cannot solve the entire problem on the GPU, we have an auxiliary operator called `ToCupy` that can be chained like any other PyLops operator and can be used to lift data in and out of the GPU as needed. An example for your case would be, split the CT operator into a stack of N CT operators, each with a portion of angles, and put them into a VStack; this way, even if the entire data does not fit the GPU one, one can still apply the operator on the GPU and then move the partial data out prior to moving to the next bit. Now, if your GPU operator does fancier things (like streaming), I am happy to keep option 2 but if you just move everything to the GPU and then apply the operator, probably having options 1 and 3 is enough, and more in line with the rest of PyLops operators?

> 
> With this, I have the following observations/questions:
> 
> 1. The adjoint for the CUDA backend of ASTRA is much less matched, so the tests produce >10% relative error. We do expect the adjoint to be mismatched, but this error is still way too high. I will be looking at that on the ASTRA side in the coming time. In the meantime, the tests with CUDA backend are skipped.

Well, in general we strive for forward-adjoint pairs that pass the dot test, for fp64 quite tighly (say atol=1e-6) and for fp32 of course much less.... if you know that your operator isn't perfectly matched we can still add it and add a test with the threshold you expect, I would prefer that to skip (or at least if you skip it I would like the tests to run both the forward and adjoint somehow to make sure we test they at least run 😄 

> 2. Do we need to assert that the input is on CPU for `cpu` engine and on GPU for `cuda` engine?

See above; in general, we do not assert but we have this as a convention... or in other words, if one expects to pass a CuPy array the other input parameters of the operator should also be passed as CuPy arrays - though I don't think this applies to `CT2D` as you dispatch everything to the Astra operator....

> 3. In principle, this operator almost works with JAX on both CPU and GPU, except for ensuring devices appropriately. What is now the ambition for adding JAX support in PyLops operators?

We do not aim a 100% coverage with JAX - see this tables https://pylops.readthedocs.io/en/latest/gpu.html#supported-operators. So if you support it great, if not we just need to make sure the row for CT2D is updated accordingly.

> 4. The default `projector_type` now (`strip`) favors accuracy at the expense of performance. For experimental data the accuracy difference is usually not noticeable, so the faster `linear` projector is more commonly used. It's also the only projector available in CUDA backend. Let me know what you think fits the expected audience better.

Mmh I guess this was me making this choice... I trust your judgement of what you think is best. Maybe we can just follow the ASTRA default?

> 5. ASTRA only supports float32 dtype internally at the moment. How do we go about it?

I think this is fine. I would suggest we do something like that

```
xdtype = x.dtype
...
y = y.astype(x)
return y
```

so when we chain operators if one wants to use fp64 we don't break the chain by all of a sudden passing out a fp32. We can add a note to the doc saying that even if you pass fp64 the internal operations are done in fp32.

pylops/medical/ct.py
```diff
@@ -50,7 +52,11 @@ class CT2D(LinearOperator):
         Distance between origin and detector along the source-origin line
         (only for "proj_geom_type=fanflat")
     projector_type : :obj:`int`, optional
-        Type of projection geometry (``strip``, or ``line``, or ``linear``)
+        Type of projection kernel (``strip``, or ``line``, or ``linear``).
+        For ``cuda`` computation engine, only ``cuda`` (hardware-accelerated ``linear``)
```
mrava87: Type of projection geometry: ``strip``, or ``line``, or ``linear`` for ``engine=cuda``,
and  ``cuda`` (i.e., hardware-accelerated ``linear``) for ``engine=cuda``.

pylops/medical/ct.py
```diff
@@ -86,54 +92,125 @@ class CT2D(LinearOperator):
...
-        det_count: float,
+        det_width: float,
+        det_count: int,
         thetas: NDArray,
         proj_geom_type: Optional[str] = "parallel",
         source_origin_dist: float = None,
         origin_detector_dist: float = None,
-        projector_type: Optional[str] = "strip",
+        projector_type: Optional[str] = None,
+        engine="cpu",
```
mrava87: engine: str = "cpu",

pylops/medical/ct.py
```diff
@@ -86,54 +92,125 @@ class CT2D(LinearOperator):
...
+            if projector_type == "cuda":
+                logging.warning("'cuda' projector type specified with 'cpu' engine.")
+        elif engine == "cuda":
+            if projector_type in [None, "cuda"]:
+                self.projector_type = "cuda"
+            else:
+                raise ValueError(
+                    "Only 'cuda' projector type is supported for 'cuda' engine."
+                )
+
```
mrava87, line 144: else:
     raise NotImplementedError("Engine must be cpu or cuda")

pylops/medical/ct.py
```diff
@@ -86,54 +92,125 @@ class CT2D(LinearOperator):
...
-        # create projector
-        self.proj_id = astra.create_projector(
-            projector_type, self.proj_geom, self.vol_geom
+        self.projector_id = astra.create_projector(
+            self.projector_type, self.proj_geom, self.vol_geom
         )
-        super().__init__(
-            dtype=np.dtype(dtype), dims=dims, dimsd=(len(thetas), det_count), name=name
+
+    def _init_1_slice_3d_geometries(self):
```
mrava87, line 175: maybe add a comment explaining what this is, like you did in the PR description

pylops/medical/ct.py
```diff
@@ -86,54 +92,125 @@ class CT2D(LinearOperator):
...
     @reshaped
     def _matvec(self, x):
-        y_id, y = astra.create_sino(x, self.proj_id)
-        astra.data2d.delete(y_id)
+        ncp = get_array_module(x)
+        backend = get_module_name(ncp)
+        if backend == "numpy":
+            y_id, y = astra.create_sino(x, self.projector_id)
+            astra.data2d.delete(y_id)
+        else:
```
mrava87: does this also work if ``backend=="jax" as this is one of the options that `get_module_name` could return?

pylops/medical/ct.py
```diff
@@ -86,54 +92,125 @@ class CT2D(LinearOperator):
...
+            "cuda3d", self._3d_proj_geom, self._3d_vol_geom
         )
 
     @reshaped
     def _matvec(self, x):
-        y_id, y = astra.create_sino(x, self.proj_id)
-        astra.data2d.delete(y_id)
+        ncp = get_array_module(x)
+        backend = get_module_name(ncp)
+        if backend == "numpy":
```
mrava87: Ok maybe I see here what is your second case... if you have engine="cuda" but pass numpy you will go for the else branch and I guess internally in `astra.experimental.direct_FP3D` x is moved to the GPU?

If this is the case maybe we can leave the 3 cases as you had and let the user to choose if they want to lift data to the GPU (especially if they want it to flow between operators in a chain) or let ASTRA do it if they want numpy in - numpy out...

mrava87 (changes requested):
Very nice addition!

Overall the code changes look great, just left a few minor suggestions 😄

pylops/medical/ct.py
askorikov: Adjusted the wording as requested, but I suggest changing "geometry" -> "kernel" (or a better word?), because to me the geometry is more about it being parallel or fanbeam, etc.

pylops/medical/ct.py
askorikov: Resolved. Also see the note in the `__init__` when calling this method.

pylops/medical/ct.py
askorikov: Yeah, if `engine="cuda"`, we have ensured that `projector_type="cuda"` before, in which case ASTRA will send data to GPU, do the computations, and send the result back to CPU.

pylops/medical/ct.py
askorikov: Yes, it should! In principle, the other branch also works if the JAX array in on CPU device.

pylops/medical/ct.py
mrava87: Mmh you may have forgotten to push the code changes?

askorikov: Yes sorry, I thought I would push one bigger batch, but testing some things is taking more time than expected :)

mrava87: Ahaha no worries 😄

mrava87: And sure, feel free to change to `kernel` if you think it is a better terms 😄 this operator has not made it to a release yet, so we are also still in time to make 'breaking' changes to the parameter names if any are not ideal

askorikov:
A couple more questions:

1. I've added casting of the output to the dtype of the input (potentially with a warning), but what is the semantics of `dtype` argument in the `__init__`? Does it need to force the dtype as well?
2. I forgot that for GPU input we also require the input to be contiguous. Now I added this check, but when testing I discovered that the `.dot` method of `LinearOperator` already (implicitly) makes the input contiguous here, potentially at the expense of copying the array: https://github.com/PyLops/pylops/blob/18321da8f07d0d4fc897400cd1867f56f70001f9/pylops/linearoperator.py#L685 Is it an intended behavior? In this case, I can remove the redundant check.
3. I guess it would be nice to set `engine="cuda"` when using CuPy. Is checking `pylops.utils.deps.cupy_enabled` a good way to determine this default?
4. Is JAX with CPU backend relevant, or do you expect mostly GPU?

mrava87:
> A couple more questions:
> 
> 1. I've added casting of the output to the dtype of the input (potentially with a warning), but what is the semantics of `dtype` argument in the `__init__`? Does it need to force the dtype as well?

So `dtype` is there for a bit of an historical reason.. when we started PyLops we were subclassing from `scipy.sparse.linalg.LinearOperator` and there `dtype` was mandatory. In principle we use it to keep track of the overall type of an operator when we chain/combine multiple operators, eg:

```
d = np.ones(10)
D1 = Diagonal(d.astype(np.float32), dtype="float32")
D2 = Diagonal(d*2, dtype="float64")
D = D1 @ D2
print(D)
> <10x10 _ProductLinearOperator with dtype=float64>
```
but in practice we are not so strict, in the sense that the below does not really respect the dtype of the operator.... 
```
d = np.ones(10, dtype="float32")
D1 = Diagonal(d, dtype="float32")
D2 = Diagonal(d*2, dtype="float32")
D = D1 @ D2

x = np.ones(10, dtype="float64")
y = D @ x

print(D, y.dtype)
><10x10 _ProductLinearOperator with dtype=float32>
>dtype('float64')
```

I have been always tempted to make this more strict but i) it would require a major version bump, ii) it would require some conventions to be set (ie if the input and operator do not match in dtype, which one wins) or being very strict raising errors all the time there is no match.. so not sure at this point 😉 

> 2. I forgot that for GPU input we also require the input to be contiguous. Now I added this check, but when testing I discovered that the `.dot` method of `LinearOperator` already (implicitly) makes the input contiguous here, potentially at the expense of copying the array: https://github.com/PyLops/pylops/blob/18321da8f07d0d4fc897400cd1867f56f70001f9/pylops/linearoperator.py#L685
>     Is it an intended behavior? In this case, I can remove the redundant check.
Yes and no. Dot is invoked if you do D @ x but one could also call directly D.matvec(x) (we tend to do that inside solvers for example as it avoids doing a bunch of checks that make dot a bit slower that just invoking the matvec/rmatvec directly.. so if you really need contiguous arrays, I would suggest to keep your internal check 😄 
> 3. I guess it would be nice to set `engine="cuda"` when using CuPy. Is checking `pylops.utils.deps.cupy_enabled` a good way to determine this default?
You mean if a user does not sets `engine="cuda"` but then passes a CuPy array?  We tend to avoid outsmarting users, we expect users to be smart... so if they do something silly I rather an error is raised than something is changed under the hood for them. So `pylops.utils.deps.cupy_enabled` (and similar) is something we only use to check if a library is present and import it (or import some method), not for the check you want to do (if I understand what you want to do...)
> 4. Is JAX with CPU backend relevant, or do you expect mostly GPU?
Yeah why not.. we see JAX as a replacement for the pair numpy/cupy since JAX claims to be a library where you write one code and run it on different hardware 😉 

Hope this makes sense?

askorikov:
Okay, points 1 and 2 are clear :+1: 

For point 3, I would certainly not automatically change the engine for the user, but I thought that's just not very clear what the default engine for this operator should be. We could also make it a non-optional argument I guess. Let me know what you prefer in the end. Also, do you think CuPy input + CPU engine should be supported?

Point 4: I will try to add a comprehensive support then. By the way, is it a good idea to patch `to_numpy` method in this PR to support JAX, or is it out of scope?

mrava87:
> Okay, points 1 and 2 are clear 👍
> 
> For point 3, I would certainly not automatically change the engine for the user, but I thought that's just not very clear what the default engine for this operator should be. We could also make it a non-optional argument I guess. Let me know what you prefer in the end. Also, do you think CuPy input + CPU engine should be supported?
> 
> Point 4: I will try to add a comprehensive support then. By the way, is it a good idea to patch `to_numpy` method in this PR to support JAX, or is it out of scope?

3: Ok given the fact that a lot of combinations are possible, let's make it non-optional so one cannot forget to set it... and my preference is to allow cpu engine with Numpy, cuda engine with Numpy (you lift in and out of the GPU, and cuda engine with CuPy, but not  cpu engine with CuPy.

4: not sure what you have in mind, if it allowing to also take in JAX and give NumPy, I am happy for it to be added here but I would like to keep consistency, so also `to_cupy` should be changed to be symmetrical, so also allow JAX as input and same for the conditional ones 😄

askorikov:
Ready for review now. In the end it turned out that enabling CuPy + CPU backend is easier than not allowing it :sweat_smile: I'm thinking it's probably fine with the non-optional `engine` arg, but let me know if I should disable it after all.

mrava87:
Perfect, I'll do one final review by the end of the week 😄

pylops/medical/ct.py
```diff
@@ -86,54 +94,149 @@ class CT2D(LinearOperator):
...
+        self.source_origin_dist = source_origin_dist
+        self.origin_detector_dist = origin_detector_dist
+
+        # make "strip" projector type default for cpu engine and only allow cuda otherwise
+        if engine == "cpu":
+            if projector_type is None:
+                projector_type = "strip"
+            # fanflat geometry projectors need an appropriate suffix (unless it's "cuda")
+            if projector_type == "cuda":
+                logging.warning("'cuda' projector type specified with 'cpu' engine.")
```
mrava87: Please use this logic (https://github.com/PyLops/pylops/blob/18321da8f07d0d4fc897400cd1867f56f70001f9/pylops/basicoperators/sum.py#L13) to be consistent with the rest of PyLops' logging messages

pylops/medical/ct.py
```diff
@@ -86,54 +94,149 @@ class CT2D(LinearOperator):
...
+            elif proj_geom_type == "fanflat":
+                projector_type += "_fanflat"
+        elif engine == "cuda":
+            if projector_type in [None, "cuda"]:
+                projector_type = "cuda"
+            else:
+                raise ValueError(
```
mrava87, line 138: Ideally we add a `Raises` section. (see https://github.com/PyLops/pylops/blob/18321da8f07d0d4fc897400cd1867f56f70001f9/pylops/basicoperators/linearregression.py#L310) if we raise any error in the constructor 😄

pylops/utils/backend.py
```diff
@@ -486,32 +486,32 @@ def get_real_dtype(dtype: DTypeLike) -> DTypeLike:
     return np.real(np.ones(1, dtype)).dtype
 
 
-def to_cupy(x: NDArray) -> NDArray:
-    """Convert x to cupy array
+def to_cupy(x: npt.ArrayLike) -> npt.ArrayLike:
```
mrava87: Can you please add we use `ArrayLine` in the `typing.py` as done for `NDArray` and use it throughout: https://github.com/PyLops/pylops/blob/18321da8f07d0d4fc897400cd1867f56f70001f9/pylops/utils/typing.py#L22

pylops/utils/backend.py
```diff
@@ -486,32 +486,32 @@ def get_real_dtype(dtype: DTypeLike) -> DTypeLike:
...
     """
     if deps.cupy_enabled:
-        if cp.get_array_module(x) == np:
+        if cp.get_array_module(x) != cp:
             x = cp.asarray(x)
     return x
 
 
-def to_numpy(x: NDArray) -> NDArray:
+def to_numpy(x: npt.ArrayLike) -> NDArray:
```
mrava87: same

mrava87: well technically we should create CuPy/JAX types but I haven't got around that yet.. for another PR 😉

mrava87 (changes requested):
@askorikov this looks great!

I just left a few minor suggestions, after you have addressed them I am happy to merge this PR 😄

askorikov:
Done! I've chosen `warnings.warn` mechanism, as the warnings we have should be avoidable by the user. And one more question: shouldn't this be ImportError, or something like it?

https://github.com/askorikov/pylops/blob/de7b99554717d8a9895253b35086b6a093f8a71e/pylops/medical/ct.py#L114

mrava87:
Warnings is good, we use it as well for this kind of purpose.

To the NotImplementedError/ImportError I think it is a bit of semantics... yes it is technically an import error but since we avoid the import to raise an error if a library is not present, this error is more to say that we don't have an implementation of operator such and such if you don't have a given library installed 😉

askorikov:
Okay, clear :+1: The PR is good to go then.

mrava87:
Merging now 😄

askorikov:
Cool, thanks for reviewing! I will look at adding a CT3D operator in the future.

mrava87:
Thanks 😄 that would be very welcome at any time!

PR #705, opened 2025-10-22
fix: standard-imghdr is removed from standard library in 3.13

cako:


mrava87:
@cako where do we rely on this?

cako:
We don't right now! This is for if we want to ever build docs in 3.13 and above. I was testing on 3.13 and noticed this issue of the library being dropped from stdlib.

mrava87:
Oh right! Well, I am all in favor of keeping the python version high in CI and doc (CI is already supporting 3.13, soon I'll try to see what happens if I add 3.14 😉 )... let me try

mrava87:
Mmh 3.13 fails on intel_mkl: https://app.readthedocs.org/projects/pylops/builds/30058430/

mrava87 (approved):
Ok, looks like we can't go up to 3.13 yet, but if you agree I am happy to keep your change and bump to 3.12 for now?

cako:
@mrava87 sounds good! I'll merge this ones for now

PR #706, opened 2025-10-22
PWD based operators

FB-I:
Proposing addition to PyLops of PW based Structural Smoothing (inherently based on Spraying) and PWD local slope estimation. Both operators work in 2D. Simple example provided.


examples/plot_pwdslope.py
```diff
@@ -0,0 +1,117 @@
...
+plt.close("all")
+np.random.seed(10)
+
+###############################################################################
+# Load sigmoid model
+# ------------------
+# Load the synthetic sigmoid model used throughout the notebooks.
+sigmoid_path = os.path.join(
+    os.path.dirname(__file__), "../testdata/slope_estimate/sigmoid_model.npy"
+)
```
cako:
```suggestion
inputfile = "../testdata/sigmoid.npz"
sigmoid = np.load(inputfile)["sigmoid"]
```

examples/plot_pwdslope.py
```diff
@@ -0,0 +1,117 @@
...
+sigmoid = np.load(sigmoid_path)
+nz, nx = sigmoid.shape
+
+###############################################################################
+# Slope estimation comparison between PWD and Structure Tensor
+# ---------------------------------------
```
cako:
```suggestion
# ------------------------------------------------------------
```

examples/plot_pwdslope.py
```diff
@@ -0,0 +1,117 @@
...
+fig.colorbar(im2, ax=ax[2], fraction=0.046, pad=0.04)
+
+for a in ax:
+    a.set_xlabel("x (samples)")
+ax[0].set_ylabel("z (samples)")
+fig.tight_layout()
+
+###############################################################################
+# Structure-aligned smoothing via local slopes
+# ---------------------------------------
```
cako:
```suggestion
# --------------------------------------------
```

examples/plot_pwdslope.py
```diff
@@ -0,0 +1,117 @@
...
+
+st_sigma, _ = slope_estimate(
+    sigmoid,
+    dz=1.0,
+    dx=1.0,
+    smooth=5,
+    eps=1e-6,
+    dips=False,
+)
+st_sigma = np.asarray(st_sigma, dtype=np.float32)
```
cako:
```suggestion
st_sigma = st_sigma.astype(np.float32)
```

pylops/signalprocessing/pwd_spraying.py
```diff
@@ -0,0 +1,199 @@
...
+        x2 = np.asarray(x, dtype=np.float32, order="C").reshape(self.nz, self.nx)
+        y2 = np.zeros_like(x2)
+        spray_forward_numba(x2, self.sigma, self.radius, self.alpha, y2)
+        return y2.ravel().astype(self.dtype, copy=False)
+
+    def _rmatvec(self, x):
+        x2 = np.asarray(x, dtype=np.float32, order="C").reshape(self.nz, self.nx)
+        y2 = np.zeros_like(x2)
+        spray_adjoint_numba(x2, self.sigma, self.radius, self.alpha, y2)
+        return y2.ravel().astype(self.dtype, copy=False)
```
cako:
```suggestion
    def __init__(
        self,
        dims: Tuple[int, int],
        sigma: NDArray,
        radius: int = 8,
        alpha: float = 0.9,
        dtype: str | np.dtype = "float32",
        name: str = "PWSp",
    ):
        if len(dims) != 2:
            raise ValueError("dims must contain exactly two elements (nz, nx)")
        self._radius = int(radius)
        self._alpha = float(alpha)
        self._sigma = np.ascontiguousarray(sigma.astype(np.float32))
        super().__init__(dtype=np.dtype(dtype), dims=dims, dimsd=dims, name=name)

    @reshaped
    def _matvec(self, x: NDArray) -> NDArray:
        y = np.zeros_like(x)
        spray_forward_numba(x, self._sigma, self._radius, self._alpha, y)
        return y

    @reshaped
    def _rmatvec(self, x: NDArray) -> NDArray:
        y = np.zeros_like(x)
        spray_adjoint_numba(x, self._sigma, self._radius, self._alpha, y)
        return y
```

pylops/signalprocessing/pwd_spraying.py
```diff
@@ -0,0 +1,199 @@
...
+
+    def _matvec(self, x):
+        # y = Spray^T (Spray x)
+        y = self._sprayer @ x
+        y = self._sprayer.H @ y
+        return y
+
+    def _rmatvec(self, x):
+        # symmetric
+        return self._matvec(x)
```
cako, line 187:
```suggestion
    def __init__(
        self,
        dims: Tuple[int, int],
        sigma: NDArray,
        radius: int = 8,
        alpha: float = 0.9,
        dtype: str | np.dtype = "float32",
        name: str = "PWSm",
    ):
        if len(dims) != 2:
            raise ValueError("dims must contain exactly two elements (nz, nx)")
        self._sprayer = PWSprayer2D(
            dims=dims, sigma=sigma, radius=radius, alpha=alpha, dtype=dtype
        )
        super().__init__(dtype=np.dtype(dtype), dims=dims, dimsd=dims, name=name)

    @reshaped
    def _matvec(self, x: NDArray) -> NDArray:
        # y = Spray^T (Spray x)
        y = self._sprayer @ x
        y = self._sprayer.H @ y
        return y.todense()

    @reshaped
    def _rmatvec(self, x: NDArray) -> NDArray:
        # symmetric
        return self._matvec(x)
```

pylops/signalprocessing/pwd_spraying.py
cako, line 1: To agree with the example, this should file should be called `pwd.py`.

pylops/utils/signalprocessing.py
```diff
@@ -1,20 +1,140 @@
...
 from pylops.utils.backend import get_array_module, get_toeplitz
 from pylops.utils.typing import NDArray
 
+try:  
+    from ._pwd2d_numba import conv_allpass_numba
+except ImportError:  
+    try:
+        from _pwd2d_numba import conv_allpass_numba  
+    except ImportError:  
+        conv_allpass_numba = None
```
cako: I don't think this is needed! Numba is a core dependency of PyLops

pylops/utils/signalprocessing.py
```diff
@@ -307,3 +427,124 @@ def dip_estimate(
...
+    sigma = np.zeros((nz, nx), dtype=np.float32)
+    delta_sigma = np.zeros_like(sigma)
+    u1 = np.zeros_like(sigma)
+    u2 = np.zeros_like(sigma)
+
+    smoothing_lower = smoothing.lower()
+    if smoothing_lower == "triangle":
+        Sop = _triangular_smoothing_from_boxcars(nsmooth=nsmooth, dims=(nz, nx), dtype="float32")
+    elif smoothing_lower == "boxcar":
+        from pylops.basicoperators import Smoothing2D
```
cako: It's probably fine to just import this at the start since it is part of the library

cako (changes requested):
@FB-I thanks a lot for this PR! This is really awesome! I found the PR very easy to read and very much in line with the rest of the operators in the library. Overall great stuff.

I have left a few comments and code suggestions, mostly bringing it up to speed with the latest PyLops features (a.k.a, reshaped, automatic shape, etc.). Otherwise just a few nits here and there.

To expand a bit more on the topic, PyLops since version 2 automatically reshapes arrays based on its inputs. So you don't need to pass `x.ravel()` to the operator or reshape it at the output. This is done with a bit of decorator magic (see [reshaped](https://pylops.readthedocs.io/en/stable/api/generated/pylops.utils.decorators.reshaped.html)) as well as proper super class initialization. Basically if you decorate your _matvec and _rmatvec with @reshaped and then ensure that you call super with dims and dimsd arguments, you are all good to go. See the [Sum](https://github.com/PyLops/pylops/blob/master/pylops/basicoperators/sum.py) operator for an example.

I haven't had time to look at the results yet, but I have noticed that the smoothing looks a but weird with the PWD slopes (but works with the structural-tensor slopes). Could point to some incorrect implementation but I haven't had time to dig:
image

pylops/signalprocessing/pwd_spraying.py
FB-I: Hi @cako, thank you for the previous suggestions. 

I modified the file name in pwd_spraying2d.py to be more consistent with the 2d/3d nomenclature used with the signal_processing operators. I also modified the import in the example.

pylops/utils/signalprocessing.py
FB-I: Ok, I've deleted that.

pylops/utils/signalprocessing.py
FB-I: Sounds good. I'll import it from the start.

pylops/utils/signalprocessing.py
mrava87: @cako no, currently `numba` is still a optional dependency... so we need protected imports.

But what I see now is different from this and in line with what we do everywhere in PyLops 😄

mrava87:
@FB-I indeed, great work!

I agree with Carlos that the code is well written and well organized and that these new features are going to be very useful!

I have so far only looked at the example in details and cleaned up a bit (trying to make it as consistent as possible to the rest of our example). I also add a small 3D example at the end... I am not aware of any real 3D implementation of PWD, usually what one does is cascaded 2D versions (yes, one could at least solve the estimation process for all slices together and impose some smoothness in the extra dimension, but I don't think that even Madagascar does that 😉 ) and this is the approach I took here.

I am now going through the other files, I will push some minor re-organization of functions in files tomorrow, then I think this is very close to be ready to be merged 🚀

pylops/utils/signalprocessing.py
```diff
@@ -1,20 +1,158 @@
...
+
+def _conv_allpass(
+    din: np.ndarray,
+    dip: np.ndarray,
+    order: int,
+    u1: np.ndarray,
+    u2: np.ndarray,
+) -> None:
+    """Dispatch to numba kernel when available, otherwise use Python fallback."""
+    global _pwd_warning_emitted
```
mrava87: Can you explain the rationale of this? Would just 

```
if _jit_message_pwd is not None:
         warnings.warn(_jit_message_pwd)
        _conv_allpass_python(din, dip, order, u1, u2)
```

not work equally well?

pylops/utils/signalprocessing.py
cako: Oh my bad then! You guys got it :muscle:

cako:
Looking pretty good to me! I'll let you handle the final changes that @mrava87 suggested and let him approve it. On my end I made a few commits fixing some typing, improving conciseness and adjusting some Numba code. I also change the `slope_estimate` smoothing and eps to better match the PWD results.

pylops/utils/signalprocessing.py
FB-I: My bad, I assumed that without it, the users would get the warning message of the fallback every time _conv_allpass is called, but actually, it is not.

examples/plot_pwdslope.py
```diff
@@ -8,29 +8,41 @@
...
 # Sigmoid model
 # ------------------
-# To start we import the same `2d image <http://ahay.org/blog/2014/10/08/program-of-the-month-sfsigmoid/>`_
+# To start we import the same `2D image <http://ahay.org/blog/2014/10/08/program-of-the-month-sfsigmoid/>`_
 # that we used in the seislet transform example. This image contains curved
 # reflectors that we will try to follow during the smoothing operation.
 inputfile = "../testdata/sigmoid.npz"
 
-sigmoid = np.load(inputfile)["sigmoid"].T
+sigmoid = 1e3 * np.load(inputfile)["sigmoid"].T
```
mrava87, line 45: @cako any specific reason for adding 1e3 * ?

FB-I:
I think that Madagascar is actually doing this "one could at least solve the estimation process for all slices together and impose some smoothness in the extra dimension, but I don't think that even Madagascar does that", all slices are updated in one linear solve and they use a 3d smoother for regularization to link the 3rd dim. But I can look into that if it is needed and later update the 3d implementation .

Also in @mrava87 new example numpy.concat() does not seem to work unless changed in numpy.concatenate()

mrava87:
> I think that Madagascar is actually doing this "one could at least solve the estimation process for all slices together and impose some smoothness in the extra dimension, but I don't think that even Madagascar does that", all slices are updated in one linear solve and they use a 3d smoother for regularization to link the 3rd dim. But I can look into that if it is needed and later update the 3d implementation .
> 
> Also in @mrava87 new example numpy.concat() does not seem to work unless changed in numpy.concatenate()

Sounds good! No need to do it now 😄 

`numpy.concat()` was added in some recent version of numpy, it is an alias to `numpy.concatenate()`  so I switched to the latter to avoid other users facing the the issue you faced

mrava87:
@FB-I as you will see I did a bit of reshuffling of codes (mostly the internal ones) to try to keep as much consistency as possible with the rest of PyLops codebase. 

Other than that, I am also happy now with how things look. If you are ok, I am ready t merge? 🚀

FB-I:
@mrava87 Sure, thank you. It is ok to merge from my side.

examples/plot_pwdslope.py
cako: Oh, I normally do this to sigmoid to bring it to order of magnitude ~1. Totally optional.

PR #708, opened 2025-10-29
feat: added multithreading support to stacking ops

mrava87:
This PR introduces support for multithreading (alongside the already present multiprocessing) in the stacking operators. 

The main motivation of this stems from the recently introduced [Python support for free threading](https://docs.python.org/3/howto/free-threading-python.html) (aka Python with GIL disabled). The changes are rather trivial and are tested in https://github.com/PyLops/pylops-free_threading (currently private, only visible to PyLops org. owners).

@cako I wonder if you have already some experience with the new support for real multithreading in Python and want to take a look. The performance comparison between serial/multiprocess/multithreading tells me that there is clearly a benefit (see https://github.com/PyLops/pylops-free_threading/blob/main/pyblockdiag.ipynb and https://github.com/PyLops/pylops-free_threading/blob/main/pyhstack.ipynb) but it looks like that similar benefit can also be seen in python 3.13 with GIL enabled (so not truly with multithreaded code but just threads alternating each other)....

### DIFF
--- a/examples/plot_multiproc.py
+++ b/examples/plot_multiproc.py
@@ -1,15 +1,14 @@
 """
-Operators with Multiprocessing
-==============================
+Operators with Multithreading/Multiprocessing
+=============================================
 This example shows how perform a scalability test for one of PyLops operators
-that uses ``multiprocessing`` to spawn multiple processes. Operators that
-support such feature are :class:`pylops.basicoperators.VStack`,
-:class:`pylops.basicoperators.HStack`, and
-:class:`pylops.basicoperators.BlockDiagonal`, and
-:class:`pylops.basicoperators.Block`.
+that uses ``concurrent.futures``/``multiprocessing`` to spawn multiple
+threads/processes. Operators that support such feature are
+:class:`pylops.VStack`, :class:`pylops.HStack`, and
+:class:`pylops.BlockDiag`, and :class:`pylops.Block`.
 
-In this example we will consider the BlockDiagonal operator which contains
-:class:`pylops.basicoperators.MatrixMult` operators along its main diagonal.
+In this example we will consider the :class:`pylops.BlockDiag` operator which
+contains :class:`pylops.MatrixMult` operators along its main diagonal.
 """
 import matplotlib.pyplot as plt
 import numpy as np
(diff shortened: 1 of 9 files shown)


cako:
Nice! I haven't played around with it much, but I had a quick look at it makes sense to me. There's no write access to shared objects so you should be free from race conditions. Only thing I'd comment as a possibility is to make a `parallelism_kind` (or any other better name) where one can choose multiprocessing, threading or [subinterpreters](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.InterpreterPoolExecutor). We'd need to benchmark but my gut feeling is the default should be an "auto" like this:
1. Use free threading when available (`not sys._is_gil_enabled()`)
2. Use subinterpreters when available (`sys.version_info >= (3, 14)`
3. Use multiprocessing otherwise (is it faster 3.12 and below?)

mrava87:
Thanks @cako!

I do not have experience with subinterpreters, let me take a look 😄 

`There's no write access to shared objects so you should be free from race conditions.`: well for the multiprocessor case this is not possible, so what I do for HStack/VStack is to create partial solutions and then sum them together, but I have updated the multithreading version to leverage the fact all threads can write in a shared array (using lock to avoid race conditions...), this seems to give quite a bit of speedup over the other solution

Regarding the `parallelism_kind="auto"`, it looks like a good idea; I guess you suggest we don't add `multiproc` but `parallelism_kind`? It is be a bit backward incompatible as we would technically need to have multiprocess as default but I think since Python is changed a lot recently in that space it makes sense to have an automatic default that is smart in that sense. The only question I have, you suggest that if `sys.version_info >= (3, 14)` subinterpreters are favoured even over threads? 

And regarding multiprocessing being faster than threads in python 3.12 and below (or even 3.13), I do not think this is always the case, I see some timings where threads are still faster (although they don't really run in parallel their overhead is less than for processes)...

mrava87:
Mmh if this is the case (see towards the end https://chatgpt.com/share/69111d2e-a5dc-8011-9295-386dfab1debb), I am not so sure I am keen to re-engineer an entire solution unless it gives a clear speedup over multiprocess.

What do you think?

PS: I still think that the idea of having `parallel_kind` is better than `multiproc` so we can keep the door open to any other option in the future... will modify that 😄

mrava87:
@cako I think this is ready to go. When you have time, have a look at my replies to your comments and we can pick this up again at any later time if we see ways to further speed up computations 😄

PR #719, opened 2025-12-26
Cubic Spline Interpolator

### LINKED ISSUE #632: [Enhancement] Add Cubic Spline Interpolator
Thank you very much for this amazing package! I'm still in the process of diving deeper into it, but I would like to apply it for problems that involve 1D-interpolation.

Unfortunately, the [`pylops.signalprocessing.Interp`](https://pylops.readthedocs.io/en/stable/api/generated/pylops.signalprocessing.Interp.html) only offers interpolation which are either not accurate enough (``"linear"``, ``"nearest"``) or too computationally expensive (``"sinc"``).

## Proposed Feature

I would like to propose to incorporate a cubic spline interpolation into `Interp` which would be a good middle ground in terms of accuracy and computational load.

While I managed to implement it, I don't know how easy this can be cast into the backend-agnostic nature of `pylops`. On top of that, I'm not 100% sure if this can easily be generalised to arbitrary dimensions, altough I think that a 1D-only implementation would already be great.

## Theoretical Background

I derived the following matrix notation from the [German Wikipedia article on spline interpolation](https://de.wikipedia.org/wiki/Spline-Interpolation#Der_kubische_C%C2%B2-Spline). It was a bit of tedious work with index-bookkeeping and reformulation of the equations, but I got it to work.

The matrix formulation for the interpolation would be 

image

where

- $\mathbf{X}$ is the design matrix that depends on the x-coordinates $x_{fit}$ of the data points to be fitted as well as the x-coordinates $x_{interp}$ of the data points to be interpolated
- $\mathbf{Q}$ is the matrix that obtains the coefficients for the cubic spline interpolation
- $\mathbf{y_{fit}}$ are the $n$ data points to be fitted
- $y_{interp}$ are the interpolated data points
- $\mathbf{O}$ is the interpolation operator

For natural boundary conditions (second derivative at the boundaries is zero), the matrix $\mathbf{Q}$ looks like

image

where

- $\mathbf{I}_{n}$ is the identity matrix of size $n\times n$
- $\mathbf{0}_{1}^{T}$ is a zero vector of size $1\times n$
- $\mathbf{A}$ is a tridiagonal matrix (see below) of size $\left(n-2)\times(n-2\right)$
- $\mathbf{D}$ is a second order finite difference matrix (see below) of size $\left(n-2\right)\times n$

The matrix $\mathbf{A}$ is a tridiagonal strictly diagonally dominant symmetric matrix with the following structure:

image

The matrix $\mathbf{D}$ is a second order finite difference matrix with the following structure:

image

$\mathbf{X}$ has to be constructed to assign the correct x-coordinates to the data points to be fitted and interpolated. Each row in $\mathbf{X}$ assigns weights to the 2 nearest values in $y_{fit}$ as well as their corresponding solutions of the linear system $\mathbf{A}^{-1}\cdot\mathbf{D}\cdot\mathbf{y_{fit}}$.

Consequently, the transpose operation $\mathbf{O}^T$ looks like

image

In both the $\mathbf{O}$ and $\mathbf{O}^T$ case, the solution of $\mathbf{A}^{-1}$ can be achieved very fast by using the [Thomas algorithm](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.solve_banded.html).

## Implementation

I would like to share my implementation that currently relies on `scipy.sparse.linalg.LinearOperator` because I didn't manage to inherit from `Interp` yet. Besides, it invokes `scipy` further for handing the sparse matrix ``X`` and solving the banded matrix with LAPACK.
I also implemented a custom finite difference operator with Numba to be fast.

The code was written in a Jupyter notebook and all the `%magic` operations are commented out. The timing results can be found below.

```python
# === Imports ===

from typing import Tuple, Type

import numpy as np
from numba import jit
from matplotlib import pyplot as plt
from pylops.signalprocessing import Interp
from pylops.utils import dottest
from scipy.linalg import solveh_banded
from scipy.sparse import csr_matrix
from scipy.sparse.linalg import LinearOperator

# %matplotlib widget

# === Second Order Finite Differences ===

@jit(nopython=True)
def second_order_finite_differences(x: np.ndarray) -> np.ndarray:
    """
    Computes the second order finite differences of ``x``.

    Parameters
    ----------
    x : :class:`numpy.ndarray` of shape (n,)
        The input array.

    Returns
    -------
    x_diffs : :class:`numpy.ndarray` of shape (n - 2,)
        The second order finite differences of ``x``.

    """

    assert x.ndim == 1
    assert x.size >= 3
    x_diffs = np.empty(shape=(x.size - 2,), dtype=x.dtype)

    for index in range(0, x.size - 2):
        x_diffs[index] = x[index + 2] - 2.0 * x[index + 1] + x[index]

    return x_diffs


@jit(nopython=True)
def second_order_finite_differences_transposed(x: np.ndarray) -> np.ndarray:
    """
    Computes the transposed operation of the second order finite differences operator,
    i.e.,

    - ``x`` is padded with 2 leading and 2 trailing zero,
    - the second order finite differences of the padded array are computed

    Parameters
    ----------
    x : :class:`numpy.ndarray` of shape (n,)
        The input array.

    Returns
    -------
    x_diffs : :class:`numpy.ndarray` of shape (n + 2,)
        The result of the transposed second order finite differences operator.

    """

    assert x.ndim == 1
    if x.size == 1:
        return np.array([x[0], -2.0 * x[0], x[0]])
    if x.size == 2:
        return np.array([x[0], -2.0 * x[0] + x[1], x[0] - 2.0 * x[1], x[1]])

    x_diffs = np.empty(shape=(x.size + 2,), dtype=x.dtype)

    x_diffs[0] = x[0]
    x_diffs[1] = -2.0 * x[0] + x[1]

    for index in range(0, x.size - 2):
        x_diffs[index + 2] = x[index] - 2.0 * x[index + 1] + x[index + 2]

    x_diffs[x.size] = x[x.size - 2] - 2.0 * x[x.size - 1]
    x_diffs[x.size + 1] = x[x.size - 1]

    return x_diffs


# === Custom Cubic Spline Interpolator ===

@jit(nopython=True)
def _make_cubic_spline_left_hand_side(dims: int) -> np.ndarray:
    """
    Constructs the banded matrix ``A`` for the linear system of equations ``A @ m = b``
    where

    - ``A`` is a symmetric diagonally dominant tridiagonal matrix with the main diagonal
        being 2/3 and the off-diagonals being 1/6,
    - ``m`` is the unknown vector of spline coefficients,
    - ``b`` is the second order finite differences of the ``y`` values for which the
        spline should interpolate.

    Parameters
    ----------
    dims : :class:`int`
        The number of points the spline should interpolate.

    Returns
    -------
    lhs_banded_matrix : :class:`numpy.ndarray` of shape (2, dims - 2)
        The banded matrix ``A`` in the LAPACK lower banded storage format for Hermitian
        tridiagonal matrices.

    """

    lhs_banded_matrix = np.empty(shape=(2, dims - 2))
    lhs_banded_matrix[0, ::] = 0.6666666666666666  # 2/3
    lhs_banded_matrix[1, ::] = 0.16666666666666666  # 1/6

    return lhs_banded_matrix

@jit(nopython=True)
def _make_cubic_spline_x_csr_specs(
    dims: int,
    iava: np.ndarray,
    base_indices: np.ndarray,
    iava_remainders: np.ndarray,
) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
    """
    Constructs the specifications ``data``, ``indices``, and ``indptr`` for a
    :class:`scipy.sparse.csr_matrix` ``X`` that can be used to interpolate the
    equally spaced ``y`` values with a cubic spline like ``X @ Q @ y`` where

    - ``X`` is the interpolation matrix to be constructed,
    - ``Q`` is the linear operator that obtains the spline coefficients from ``y``
        (by a vertical concatenation of the ``y`` values and the spline coefficients
        ``m``),
    - ``y`` are the values to be interpolated.

    Parameters
    ----------
    dims : :class:`int`
        The number of points the spline should interpolate.
    iava : :class:`numpy.ndarray` of shape (n,)
        Floating indices of the locations to which the spline should interpolate.
    base_indices : :class:`numpy.ndarray` of shape (n,)
        The indices of the respective first data point in ``y`` for the intervals
        in which the corresponding ``iava`` values lie.
    iava_remainders : :class:`numpy.ndarray` of shape (n,)
        The remainders of the ``iava`` values after subtracting the respective
        ``base_indices``.

    Returns
    -------
    data : :class:`numpy.ndarray` of shape (4 * ``iava.size``,)
        The data array of the CSR matrix ``X``
    indices : :class:`numpy.ndarray` of shape (4 * ``iava.size``,)
        The column indices of the CSR matrix ``X``
    indptr : :class:`numpy.ndarray` of shape (``iava.size`` + 1,)
        The row pointers of the CSR matrix ``X``

    """

    # some auxiliary variables are required
    iava_remainders_cubic = (  # (x - x[i])**3
        iava_remainders * iava_remainders * iava_remainders
    )
    one_minus_iava_remainders = 1.0 - iava_remainders  # (x[i + 1] - x)
    one_minus_iava_remainders_cubic = (  # (x[i + 1] - x)**3
        one_minus_iava_remainders
        * one_minus_iava_remainders
        * one_minus_iava_remainders
    )

    # for each data point, except for the first and the last one, we need 4 entries
    # to multiply with ``y[i]``, ``y[i + 1]``, ``m[i]``, and ``m[i + 1]``;
    data = np.empty(shape=(4 * iava.size,), dtype=np.float64)
    indices = np.empty(shape=(4 * iava.size,), dtype=np.intp)
    indptr = np.arange(0, 4 * (iava.size + 1), 4, dtype=np.int64)

    # the rows are specified by means of a loop
    for row_index in range(0, iava.size):
        index_offset = 4 * row_index
        data[index_offset] = one_minus_iava_remainders[row_index]
        data[index_offset + 1] = iava_remainders[row_index]
        data[index_offset + 2] = 0.16666666666666666 * (
            one_minus_iava_remainders_cubic[row_index]
            - one_minus_iava_remainders[row_index]
        )
        data[index_offset + 3] = 0.16666666666666666 * (
            iava_remainders_cubic[row_index] - iava_remainders[row_index]
        )

        indices[index_offset] = base_indices[row_index]
        indices[index_offset + 1] = base_indices[row_index] + 1
        indices[index_offset + 2] = base_indices[row_index] + dims
        indices[index_offset + 3] = base_indices[row_index] + dims + 1

    return data, indices, indptr


class CubicSplineInterp(LinearOperator):
    """
    Custom cubic spline interpolator.

    Parameters
    ----------
    dims : :class:`int`
        The number of points the spline should interpolate.
    iava : :class:`numpy.ndarray` of shape (n,)
        Floating indices of the locations to which the spline should interpolate.
    dtype : type
        The data type of the input and output arrays.

    """

    def __init__(
        self,
        dims: int,
        iava: np.ndarray,
        dtype: Type = np.float64,
    ) -> None:

        self.dims: int = dims
        self.iava: np.ndarray = iava
        self.shape: Tuple[int, int] = (iava.size, dims)
        self.dtype: Type = dtype

        self.base_indices = np.clip(
            iava.astype(np.int64),
            a_min=0,
            a_max=dims - 2,
        )
        self.iava_remainders = iava - self.base_indices
        self.lhs_banded_matrix: np.ndarray = _make_cubic_spline_left_hand_side(
            dims=dims
        )
        self.X_matrix: csr_matrix = csr_matrix(
            _make_cubic_spline_x_csr_specs(
                dims=dims,
                iava=iava,
                base_indices=self.base_indices,
                iava_remainders=self.iava_remainders,
            ),
            shape=(iava.size, 2 * dims),
        )
        self.X_matrix_transposed: csr_matrix = self.X_matrix.transpose().tocsr()  # type: ignore

    def _matvec(self, x: np.ndarray) -> np.ndarray:
        m_coeffs = np.empty_like(x)
        m_coeffs[0] = 0.0
        m_coeffs[x.size - 1] = 0.0
        m_coeffs[1:x.size -1] = solveh_banded(
            ab=self.lhs_banded_matrix,
            b=second_order_finite_differences(x),
            lower=True,
            check_finite=False,
        )

        return self.X_matrix @ np.concatenate((x, m_coeffs))

    def _rmatvec(self, x: np.ndarray) -> np.ndarray:
        x_mod = self.X_matrix_transposed @ x
        return x_mod[0:self.dims] + second_order_finite_differences_transposed(
            solveh_banded(
                ab=self.lhs_banded_matrix,
                b=x_mod[self.dims + 1: x_mod.size - 1],
                lower=True,
                check_finite=False,
            )
        )


# === Interpolators Test ===

# A complex signal is created to test the interpolators

x = np.arange(50)
x_new = np.linspace(0, x.max() - np.spacing(x.max()), 50_001)
y = np.sin(6.0 * np.pi * (x / x.size)) + 1.0j * np.cos(4.0 * np.pi * (x / x.size))

linop_linear = Interp(
    dims=x.size,
    iava=x_new,
    kind="linear",
)[0]
y_new_linear = linop_linear @ y

print("Timing PyLops Linear Interpolator")
print("Creation")
# %timeit linop_linear = Interp(dims=x.size, iava=x_new, kind="linear", dtype="complex128",)
print("Matrix-Vector Multiplication")
# %timeit y_new_linear = linop_linear @ y

linop_sinc = Interp(
    dims=x.size,
    iava=x_new,
    kind="sinc",
)[0]
y_new_sinc = linop_sinc @ y

print("\nTiming PyLops Sinc Interpolator")
print("Creation")
# %timeit linop_sinc = Interp(dims=x.size, iava=x_new, kind="sinc", dtype="complex128",)
print("Matrix-Vector Multiplication")
# %timeit y_new_sinc = linop_sinc @ y

linop_spline = CubicSplineInterp(dims=x.size, iava=x_new, dtype=np.complex128,)
print("\nTesting custom Spline Interpolator")
assert dottest(linop_spline, nr=x_new.size, nc=x.size, complexflag=2)
print("Passed!")
y_new_spline = linop_spline @ y

print("\nTiming custom Spline Interpolator")
print("Creation")
# %timeit linop_spline = CubicSplineInterp(dims=x.size, iava=x_new, dtype=np.complex128,)
print("Matrix-Vector Multiplication")
# %timeit y_new_sline = linop_spline @ y

# === Plotting ===

fig, ax = plt.subplots(nrows=2, sharex=True, figsize=(10, 6))

ax[0].plot(x, y.real, "o", label="Original data")
ax[0].plot(x_new, np.sin(6.0 * np.pi * (x_new / x.size)), label="Ground truth")
ax[0].plot(x_new, y_new_spline.real, label="Cubic spline interpolation")
ax[0].plot(x_new, y_new_linear.real, label="Linear interpolation", linestyle=":")
ax[0].plot(x_new, y_new_sinc.real, label="Sinc interpolation", linestyle="--")

ax[0].set_ylabel("Real part")

ax[0].legend()

ax[1].plot(x, y.imag, "o", label="Original data")
ax[1].plot(x_new, np.cos(4.0 * np.pi * (x_new / x.size)), label="Ground truth")
ax[1].plot(x_new, y_new_spline.imag, label="Cubic spline interpolation")
ax[1].plot(x_new, y_new_linear.imag, label="Linear interpolation", linestyle=":")
ax[1].plot(x_new, y_new_sinc.imag, label="Sinc interpolation", linestyle="--")

ax[1].set_xlabel("Index")
ax[1].set_ylabel("Imaginary part")

ax[1].legend()

```

### Correctness and Timing
The timings for this test with a relatively large complex Array show that the cubic spline interpolation in this basic version is competitive. It passes `dottest` and its runtime is comparable to the linear interpolation.

```bash
Testing correctness of the finite differences operator ...
Passed!

Timing PyLops Linear Interpolator
Creation
1.96 ms ± 19.2 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
Matrix-Vector Multiplication
305 µs ± 10.2 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

Timing PyLops Sinc Interpolator
Creation
13.2 ms ± 317 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
Matrix-Vector Multiplication
4.01 ms ± 555 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

Testing custom Spline Interpolator
Passed!

Timing custom Spline Interpolator
Creation
864 µs ± 31.2 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
Matrix-Vector Multiplication
272 µs ± 2.78 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

### Visualisation

The real and imaginary part of the complex test data are interpolated separately as intended. For this example, the cubic spline gives higher accuracy than the linear method. Besides, it also does not suffer from the ringing introduces by the sinc-interpolation of the imaginary part.

image

MothNik:
This pull request aims to close #632 by adding the `kind="cubic_spline"`-option to `pylops.signalprocessing.Interp`.
It is implemented in a way that can handle 1D-Arrays and 2D-Arrays along a single axis (0 or 1) in a vectorized fashion that heavily reuses pre-computations (optimised to match speed of `kind="linear"` while providing similar accuracy as `kind="sinc"` with less oscillations).

Currently, only a `numpy`-backend is available.


MothNik:
# Pull Request Checklist
- [x] implement ~`CubicSplineInterpolator`~ `InterpCubicSpline` basic logic
- [x] handle real and complex data types
- [x] optimize runtime to be comparable with `Interp(..., kind="linear")`
- [x] handle 1D-vectors
- [x] ~handle 2D-Arrays~
- [x] handle ND-Arrays with `axis`-argument (thanks to @mrava87 🚀 )
- [ ] ~protect against wrong~ document availability of backends (tridiagonal solver seems unavailable for `cupy`)
- [x] check if 2D-Array memory layout (C/F-order) matters
- [x] integrate ~`CubicSplineInterpolator`~ `InterpCubicSpline` into existing tests
- [x] validate ~`CubicSplineInterpolator`~ `InterpCubicSpline` against [`scipy.interpolate.CubicSpline(..., bc_type="natural")`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.CubicSpline.html)
- [x] add tiny documentation section to `Interp`
- [ ] add mathematical documentation `Notes` section to `CubicSplineInterpolator` 
- [x] move ~`CubicSplineInterpolator`~ `InterpCubicSpline` to its own linear operator file
- [x] fix floating point rounding error in `iava`-clipping for very large samples
- [ ] fix typing of `iava` to `SamplingLike`
- [x] integrate `InterpCubicSpline` into `pylops` documentation

~<sup>1)</sup> What's not obvious to me right now is how `dims` and `dimsd` work and why the `@` (`__matmul__`) operator decides to flatten 2D-input before passing it down to the `_matmat` and `_rmatmat` methods.~

mrava87:
@MothNik thanks a lot for taking the time to turn the long stand Github issue into a PR ❤️ 

I am going to try to reply to all the points you have raised here and in your last comment in the Github issue, I will later do a proper review of the code and provide inline comments:

- handling of 2D/ND arrays: I recall we discussed this briefly in the Github issue. Basically in PyLops we historically allowed two type of inputs to an operator (I discuss for simplicity the case of forward, the same applies to adjoint)

     1. 1d arrays with size equal to the number of columns of the operator. Applying the operator is equivalent to a matvec if the operator was a dense matrix, and the output is a 1d array of size equal to the number of rows of the operator.
     2. 2d arrays (or matrices) where the number of rows equals to the number of columns of the operator and an arbitrary number of columns. Here the operator is applied to each column with a naive for...loop - https://github.com/PyLops/pylops/blob/eeb8c474e71609bb29d2aab0ae5931a5692c4d3c/pylops/linearoperator.py#L451

  In PyLops v2 we also started allowing one to pass inputs that are ND-arrays in the natural shape an operator would work (so case 1 but with the input not being flattened)... this is the scenario we implement with the help of `dims/dimsd` as they are the unflattened version of `shape[1]/shape[0]` (basically `np.prod(dims)=shape[1]` and `np.prod(dims)=shape[0]`), and this is what I think you did not really implement - hence the tests failing. I have played a bit and found out that the current code wasn't that far for allowing this, see my commit for the changes that allow ANY ND-array (so not just 2d but also 3/4/ND) to work with any choice of axis.

- matmat: I see you implemented some custom _matmat/_rmatmat... as I explained above we do not do it as matmat isn't a very common use case for PyLops (it is there more for historical reasons as SciPy supports it)... I wonder if you actually did it thinking it will handle the 2D/ND array scenario I was asking?
Also I tried this:
```
np.random.seed(1)
x = np.random.normal(0, 1, (10, 50))

perc_subsampling = 0.4
Nsub = int(np.round(10 * perc_subsampling))
iava = np.sort(np.random.permutation(np.arange(10))[:Nsub])

Iop, _ = Interp(
    10,
    iava,
    axis=0,
    kind="cubic_spline",
    dtype=par["dtype"],
)

Iop._matmat(x)
```
and seems to fail `ValueError: all the input array dimensions except for the concatenation axis must match exactly, but along dimension 1, the array at index 0 has size 50 and the array at index 1 has size 52` 🤔 

- cupy: it is ok to have an operator that has only a numpy (aka cpu) backend, and your scenario (ie lack of equivalent in cupy to numpy/scipy method used in the operator) is a common one... for those operators we however do not feel it is important to have checks if one passes cupy arrays, but we simply state the capabilities of each operator in https://github.com/PyLops/pylops/blob/dev/docs/source/gpu.rst. In this case, since `Interp` will be partially CuPy/JAX compatible, we can use the warning emoji and had a warning at the bottom specifying that for spline the only backend is numpy... see Convolve1D for an example

- Notes: I think this is definitely important as this is not a trivial operator, and you had a nice description in the Github issue. And here comes a suggestion: since this operator is effectively implemented in `CubicSplineInterpolator`, I would make this a public class (calling the file `interpspline` and the operator `InterpCubicSpline`). We can still also have it as an option of `Interp` but having the class public we can more easily expose it in the doc and add a good documentation of what it does, I feel it would be too much to be part of the Notes of Interp (basically we can treat it the same way we treat `nearest`. I also suggest to add the spline interpolator to this example https://github.com/PyLops/pylops/blob/dev/examples/plot_restriction.py, and feel free to add more (like the example you had in the Github issue)

- Minor Finding No. 1 - Typing of iava: you are right, there is an inconsistency here, which I think originated from this massive PR (https://github.com/PyLops/pylops/commit/d765af9285e4b11651901483f3ffd7a10f77c02c#diff-db0feeec6c30154190addf833c81fba3c7a670f4fce6579966fcc942307094a4) when we added a lot of new type hints. The correct type for `iava` should be `NDArray` as we expect floats as element of the array. In fact, if one passed just ints we will have already the correct values in the input and we could just fallback to the Restriction operator (and so the operator would work with int values, but it will be an overkill). So if you don't mind, could you fix the types to `SamplingLike` since in the doc we say we also allow lists 😄 

- Minor Finding No. 2 - Clipping of iava - Interesting, I never noticed this (probably because I never had such large inputs), but worth fixing it now that you spotted a problem. I am just not sure why you have added the `1e6`; this still seems to work and gives a number close to the actual last value:
```
>>> n * (1 - np.finfo(np.float64).eps)
np.float64(1999999.9999999995)
```
probably nothing that will be ever perceived, but just curious 😉 

<sup>1)</sup> not sure what you mean, if I run the code I gave you above with `Iop @ x` instead of directly `Iop._matmat(x)` and I print the shape of x in your _matmat method, I get  (10, 50), which is what I expect...

MothNik:
@mrava87 
Thanks a lot for taking the time for such detailed feedback, the explanations, and providing a fix!
I misunderstood the 2D-Array-handling then, and I must say that the handling of ND-Arrays is really a genius addition 🤩 
I will go through your feedback and adapt the PR 👍

mrava87:
Perfect 😄 Do not worry about some of the CI failing, these are problems unrelated to your PR (as long as the doc and main Github action CI passes it is all good...)

MothNik:
@mrava87 
I was adapting the tests in `test_interpolation` and I uncovered and fixed a lingering test bug ([commit](https://github.com/PyLops/pylops/pull/719/commits/fb696dcb34edeeb7620c466e664d3f3936dc3ae6)) in the tests for `Bilinear` (the tests, not the operator itself).

The lines 

```python
    with pytest.raises(ValueError):
        iava_rep = iava.copy()
        iava_rep[-2] = [0, 0]
        iava_rep[-1] = [0, 0]
        _, _ = Bilinear(
            iava_rep, dims=(par.y_num, par.x_num, par.t_num), dtype=par.dtype
        )
```

passed the assertion, but never hit the `iava`-duplicate check in `Bilinear`.
Changing this using regex-matching with a relevant part of the error message like

```python
    with pytest.raises(ValueError, match="repeated"):
        iava_rep = iava.copy()
        iava_rep[-2] = [0, 0]
        iava_rep[-1] = [0, 0]
        _, _ = Bilinear(
            iava_rep, dims=(par.y_num, par.x_num, par.t_num), dtype=par.dtype
        )
```

revealed the following

```bash
>       with pytest.raises(ValueError, match="repeated"):
E       AssertionError: Regex pattern did not match.
E         Expected regex: 'repeated'
E         Actual message: 'could not broadcast input array from shape (2,) into shape (10,)'
```

The error happened already in `iava_rep[-2] = [0, 0]` due to a shape mismatch, which also raises a `ValueError`.

Checking against common exceptions like `TypeError` or `ValueError` with `pytest.raises` alone is relatively risky because they are frequently raised in other contexts. Then, `pytest.raises` just swallows the context and nobody can infer what happened unless closely inspecting the coverage reports. I recommed to use the `match=...` with a meaningful regex pattern to avoid tests passing without hitting the targeted branch.

Altneratively, and probably more readable, one can use something more verbose like

```python
    with pytest.raises(ValueError) as exception_info:
        iava_rep = iava.copy()
        iava_rep[-2] = [0, 0]
        iava_rep[-1] = [0, 0]
        _, _ = Bilinear(
            iava_rep, dims=(par.y_num, par.x_num, par.t_num), dtype=par.dtype
        )
        
        error_message = str(exception_info.value)
        assert "iava" in error_message
        assert "repeated" in error message
```

mrava87:
> @mrava87 
> 
> I was adapting the tests in `test_interpolation` and I uncovered and fixed a lingering test bug ([commit](https://github.com/PyLops/pylops/pull/719/commits/fb696dcb34edeeb7620c466e664d3f3936dc3ae6)) in the tests for `Bilinear` (the tests, not the operator itself).
> 
> 
> 
> The lines 
> 
> 
> 
> ```python
> 
>     with pytest.raises(ValueError):
> 
>         iava_rep = iava.copy()
> 
>         iava_rep[-2] = [0, 0]
> 
>         iava_rep[-1] = [0, 0]
> 
>         _, _ = Bilinear(
> 
>             iava_rep, dims=(par.y_num, par.x_num, par.t_num), dtype=par.dtype
> 
>         )
> 
> ```
> 
> 
> 
> passed the assertion, but never hit the `iava`-duplicate check in `Bilinear`.
> 
> Changing this using regex-matching with a relevant part of the error message like
> 
> 
> 
> ```python
> 
>     with pytest.raises(ValueError, match="repeated"):
> 
>         iava_rep = iava.copy()
> 
>         iava_rep[-2] = [0, 0]
> 
>         iava_rep[-1] = [0, 0]
> 
>         _, _ = Bilinear(
> 
>             iava_rep, dims=(par.y_num, par.x_num, par.t_num), dtype=par.dtype
> 
>         )
> 
> ```
> 
> 
> 
> revealed the following
> 
> 
> 
> ```bash
> 
> >       with pytest.raises(ValueError, match="repeated"):
> 
> E       AssertionError: Regex pattern did not match.
> 
> E         Expected regex: 'repeated'
> 
> E         Actual message: 'could not broadcast input array from shape (2,) into shape (10,)'
> 
> ```
> 
> 
> 
> The error happened already in `iava_rep[-2] = [0, 0]` due to a shape mismatch, which also raises a `ValueError`.
> 
> 
> 
> Checking against common exceptions like `TypeError` or `ValueError` with `pytest.raises` alone is relatively risky because they are frequently raised in other contexts. Then, `pytest.raises` just swallows the context and nobody can infer what happened unless closely inspecting the coverage reports. I recommed to use the `match=...` with a meaningful regex pattern to avoid tests passing without hitting the targeted branch.
> 
> 
> 
> Altneratively, and probably more readable, one can use something more verbose like
> 
> 
> 
> ```python
> 
>     with pytest.raises(ValueError) as exception_info:
> 
>         iava_rep = iava.copy()
> 
>         iava_rep[-2] = [0, 0]
> 
>         iava_rep[-1] = [0, 0]
> 
>         _, _ = Bilinear(
> 
>             iava_rep, dims=(par.y_num, par.x_num, par.t_num), dtype=par.dtype
> 
>         )
> 
>         
> 
>         error_message = str(exception_info.value)
> 
>         assert "iava" in error_message
> 
>         assert "repeated" in error message
> 
> ```
> 
> 

Oh yeah, good catch! I agree, sometimes our with... tests have too generic assertments leading to unexpected behaviors like the one you spotted. I like the second one so we can check multiple words or even partial sentences in the error message 😉

pylops/signalprocessing/_interp_utils.py
```diff
@@ -0,0 +1,46 @@
+import warnings
+
+import numpy as np
+
+from pylops.utils.typing import NumericNDArray
+
+
+def ensure_iava_is_unique(
```
mrava87: Can you please add `_` to both methods to re-enforce the fact they are private methods :)

pylops/signalprocessing/_interp_utils.py
```diff
@@ -0,0 +1,46 @@
...
+    iava: NumericNDArray,
+    axis: int | None = None,
+) -> None:
```
mrava87, line 11: Add docstring `"""Ensure that iava has only unique values"""`... since it is a private method it is enough to just state what the method does.

pylops/signalprocessing/_interp_utils.py
```diff
@@ -0,0 +1,46 @@
...
+        raise ValueError("Found repeated/non-unique values in iava.")
+
+    return
+
+
+def clip_iava_above_last_sample_index(
+    iava: NumericNDArray,
+    sample_size: int,
+) -> None:
+    # ensure that samples are not beyond the last sample, in that case set to
```
mrava87: convert this comment into docstring

pytests/test_interpolation_spline.py
```diff
@@ -0,0 +1,103 @@
...
+    y_eval_scipy = CubicSpline(
+        x=x_fit,
+        y=y_fit,
+        bc_type="natural",
+        axis=axis,
+    )(x=x_eval_for_scipy)
+
+    assert np.allclose(y_eval_pylops, y_eval_scipy)
+
+    return
```
mrava87: return is not needed

pytests/test_interpolation_spline.py
```diff
@@ -0,0 +1,103 @@
...
+    ],
+)
+@pytest.mark.parametrize(
+    "subsample_fraction",
+    [
+        pytest.param(0.5, id="decimation"),
+        pytest.param(5.0, id="upsampling"),
+    ],
+)
+def test_natural_cubic_spline_against_scipy(
```
mrava87, line 45: nice test to have 😄

pytests/test_interpolation_spline.py
```diff
@@ -0,0 +1,103 @@
...
+    axis: int,
+    with_complex: bool,
+) -> None:
+    """
+    Tests ``pylops.signalprocessing.InterpCubicSpline`` against the equivalent
+    implementation ``scipy.interpolate.CubicSpline`` for the natural boundary condition.
+
+    """
+
+    # --- Setup ---
```
mrava87: Here and everywhere, we do not favour using ``---`` in PyLops comments, simply use plain comments, here : `# Setup`

pylops/utils/typing.py
```diff
@@ -20,9 +20,13 @@
     import torch
 
 IntNDArray = npt.NDArray[np.int_]
+NumericNDArray = npt.NDArray[np.number]
 NDArray = npt.NDArray
 ArrayLike = npt.ArrayLike
 
+Int64Vector = np.ndarray[tuple[int], np.dtype[np.int64]]
```
mrava87: we should always use  `npt.NDArray` for typing annotations, not pure `np.ndarray`

pylops/utils/typing.py
```diff
@@ -20,9 +20,13 @@
     import torch
 
 IntNDArray = npt.NDArray[np.int_]
+NumericNDArray = npt.NDArray[np.number]
```
mrava87: I was curious when I saw this and dit a little ChatGPT search... turns out that what we use in most places (``NDArray``) is very general and could potentially allow strings or objects as element of the numpy array... I can't say 100% for sure,  but I suspect this is never the case in PyLops as we always work with numeric.. so I would just suggest we change `NDArray=npt.NDArray[np.number]` instead of creating a new `NumericNDArray`

pylops/utils/typing.py
mrava87: Also to be consistent with numpy nomenclature I suggest. we use `Array` not `Vector`

Finally I am not convinced about these types, they seem to low level and not really what numpy recommends for types... see also this ChatGPT chat https://chatgpt.com/share/695843aa-e898-8011-a6a5-8cfaac4734c4

I suggest we do this change:

```
ArrayLike = npt.ArrayLike
NDArray = npt.NDArray[np.number]
IntNDArray = npt.NDArray[np.integer]
FloatingNDArray = npt.NDArray[floating]
```

and

```
SamplingLike = Union[Sequence[float], FloatingNDArray]
```

and use IntNDArray in spite of Int64Vector and FloatingNDArray in spite of Float64Vector for the spline files

pylops/signalprocessing/_interp_utils.py
```diff
@@ -0,0 +1,46 @@
...
+        warnings.warn(
+            f"At least one value in iava is beyond the penultimate sample index "
+            f"{last_sample_index}. Out-of-bound-values are forced below penultimate "
+            f"sample."
+        )
+
+        # NOTE: ``numpy.nextafter(x, -np.inf)`` gives the closest float-value that is
+        #       less than ``x``, i.e., this logic clips ``iava`` to the highest possible
+        #       value that is still below the last sample
+        iava[np.where(outside)] = np.nextafter(last_sample_index, -np.inf)
```
mrava87, line 50: Nice 😄

pylops/signalprocessing/interp.py
```diff
@@ -27,8 +24,8 @@ def _nearestinterp(
 ):
     """Nearest neighbour interpolation."""
     iava = np.round(iava).astype(int)
-    _checkunique(iava)
-    return Restriction(dims, iava, axis=axis, dtype=dtype), iava
+    ensure_iava_is_unique(iava=iava)
+    return (Restriction(dims, iava, axis=axis, dtype=dtype), iava)
```
mrava87: no need for parenthesis in return

pylops/signalprocessing/interp.py
```diff
@@ -81,7 +71,9 @@ def _sincinterp(
 ):
     """Sinc interpolation."""
     ncp = get_array_module(iava)
-    _checkunique(iava)
+
+    # TODO: is ``iava`` bound to an integer dtype
```
mrava87: What do you mean for this? See my other comment about `iva` not being an array of integers (and the need to fix the type hint

pylops/signalprocessing/interp.py
```diff
@@ -111,7 +103,7 @@ def Interp(
     dims: Union[int, InputDimsLike],
     iava: IntNDArray,
```
mrava87: This is wrong (my bad!) Can you please change it to `iava: SamplingLike` anywhere in this file (like you have in the `interpspline` file 😄

pylops/signalprocessing/interpspline.py
```diff
@@ -0,0 +1,997 @@
...
+      factorizations to solve the linear systems
+    - :math:`\mathbf{P}` is represented as a :py:class:`scipy.sparse.csr_matrix` which
+      involves a tiny bit of overhead during operator initialization, but allows for
+      a simple transpose.
+
+    """
+
+    def __init__(
+        self,
+        dims: Tuple,
```
mrava87: dims: Union[int, InputDimsLike],

pylops/signalprocessing/interpspline.py
```diff
@@ -0,0 +1,997 @@
...
+    the second derivatives at the first and last sampling point are both zero.
+
+    .. note:: The vector ``iava`` should contain unique values only. If the same
+      fractional index is present multiple times, an error will be raised. Elements that
+      exceed the last index ``dims[axis] - 1`` are clipped to the closest float value
+      right below ``dims[axis] - 1`` to avoid extrapolation.
+
+    Parameters
+    ----------
+    dims : :obj:`int`
```
mrava87: I think this should be the same as in `Interp`

```
dims : :obj:`list` or :obj:`int`
        Number of samples for each dimension.  
```

You can of course add `A cubic spline requires ``dims[axis] > 2``.` as it makes sense here.

pylops/signalprocessing/interpspline.py
```diff
@@ -0,0 +1,997 @@
...
+        For complex input, both the real and the imaginary parts are interpolated
+        separately.
+        Only double precision versions of ``numpy.inexact`` are supported, i.e., either
+        ``"float64"`` (default) or ``"complex128"``.
+        Multiplication of the operator with data with less precise data types will
+        result in a type promotion.
+    name : :obj:`str`, optional
+        Name of operator (to be used by :func:`pylops.utils.describe.describe`).
+
+    Returns
```
mrava87: No Returns on a class.... instead you should use Attributes with public facing class attributes (basically we report all but those named `self._*`

```
Attributes
----------
 dims : :obj:`tuple`
        Shape of the array after the adjoint, but before flattening.
        For example, ``x_reshaped = (Op.H * y.ravel()).reshape(Op.dims)``.
dimsd : :obj:`tuple`
        Shape of the array after the forward, but before flattening.
        For example, ``y_reshaped = (Op * x.ravel()).reshape(Op.dimsd)``.
shape : :obj:`tuple`
        Operator shape.
lhs_matrix_lu **ADD type and description**
lhs_matrix_transposed_lu **ADD type and description**
```

pylops/signalprocessing/interpspline.py
```diff
@@ -0,0 +1,997 @@
...
+    .. math::
+        \left(\mathbf{S}\mathbf{x}\right)_{j} =
+        \mathbf{x}_{i}\cdot p_{j,0}\left(t_{j}\right) +
+        \mathbf{x}_{i + 1}\cdot p_{j,1}\left(t_{j}\right) +
+        \mathbf{m}_{i}\cdot p_{j,2}\left(t_{j}\right) +
+        \mathbf{m}_{i + 1}\cdot p_{j,3}\left(t_{j}\right)
+
+    The adjoint operator :math:`\mathbf{S}^{H}` can be derived by rearranging the
+    involved operators. All of them are purely real and consequently, a transpose is
+    sufficient. This yields
```
mrava87, line 777: This doesnt render well, I think you need an empty line here

pylops/signalprocessing/interpspline.py
```diff
@@ -0,0 +1,997 @@
...
+    mapped to the knot-to-knot interval
+    :math:`\mathbf{k}_{i}\le t_{j} < \mathbf{k}_{i + 1}`, the base polynomials are:
+
+    - :math:`p_{j,0}\left(t\right) = \mathbf{k}_{i + 1} - t`
+    - :math:`p_{j,1}\left(t\right) = t - \mathbf{k}_{i}`
+    - :math:`p_{j,2}\left(t\right) = \frac{1}{6}\cdot\left(\left(\mathbf{k}_{i + 1} - t\right)^{3} - \left(\mathbf{k}_{i + 1} - t\right)\right)`
+    - :math:`p_{j,3}\left(t\right) = \frac{1}{6}\cdot\left(\left(t - \mathbf{k}_{i}\right)^{3} - \left(t - \mathbf{k}_{i}\right)\right)`
+
+    These base polynomials then need to be linearly combined using the coefficients
+    :math:`\mathbf{c} = \mathbf{F}\mathbf{y}`. Here, :math:`\mathbf{F}` is a
```
mrava87: Is this y or x? Above you say the forward is Sx=PFx so I would assume c=Fx?

pylops/signalprocessing/interpspline.py
```diff
@@ -0,0 +1,997 @@
...
+    pylops.signalprocessing.Interp : General interpolation operator
+    :py:class:`scipy.interpolate.CubicSpline` : An equivalent implementation of the forward operator :math:`\mathbf{S}\mathbf{x}`
+
+    References
+    ----------
+
+    .. [1] Wikipedia (German), Spline Interpolation
+       https://de.wikipedia.org/wiki/Spline-Interpolation#Kubisch_(Polynome_3._Grades)
+
+    Notes
```
mrava87, line 633: Very well written 😄

mrava87 (changes requested):
@MothNik great job!

I went through the PR once again and left a few minor comments. Once those are addressed I think this PR is finally ready to be merged 🥳

mrava87:
@MothNik let me know if you have time to make these final changes or if you want me to go ahead and do them before I can merge 😄

MothNik:
@mrava87 Thanks for the thorough review!
I will go over it when I'm back from vacation after this week.
Once I'm done, I will drop a note an convert this draft to a PR.

pylops/signalprocessing/interpspline.py
MothNik: It was `x`.

pylops/signalprocessing/interpspline.py
MothNik: I re-wrote the Notes because it was too much back and forth. Now, I think it is easier to follow. In this go, I fixed this one as well.

pylops/signalprocessing/interpspline.py
MothNik: I made those private since they should not be exposed.

pylops/utils/typing.py
MothNik: I changed it. That makes sense ✅

pylops/signalprocessing/interp.py
MothNik: Yes, this was a remainder that popped up on every push. It's now resolved.

pylops/signalprocessing/interp.py
MothNik: I changed it. That caused a cascade of further type check warnings downstream, e.g., in `Restriction`, but I think this PR is not the place to resolve those.

MothNik:
@mrava87 
I implemented your changes and also increased the version in the last commit to `2.7.0` (I guess?).
I made a changelog entry and converted this draft to an actual PR.

MothNik:
There is definitely also some follow-up work to do. The current $LU$-decomposition logic for the spline interpolation is borrowed from [this `scipy` issue](https://github.com/scipy/scipy/issues/20948), but I think this should at one point belong to `scipy` and not be part of `pylops`.

mrava87:
Fantastic, I will do one final review but I think this is likely ready to be merged 😄 and yes, if the logic of the LU-decomposition were to move to scipy that would be even better.. for now, happy to host it as part of the spline interpolator

PR #722, opened 2025-12-29
Fix Intel CI

### LINKED ISSUE #721: CI: Intel CI broken
@rohanbabbar04 the Intel oneAPI Math Kernel Library(oneMKL) CI seems broken (https://github.com/PyLops/pylops/actions/runs/20544483329/job/59013557790?pr=719)... could you please take a look 😄

rohanbabbar04:
Fixes #721 

- Remove Python 3.10 from `build-mkl.yaml`

### DIFF
--- a/.github/workflows/build-mkl.yaml
+++ b/.github/workflows/build-mkl.yaml
@@ -7,7 +7,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest]
-        python-version: ["3.10", "3.11", "3.12", "3.13"]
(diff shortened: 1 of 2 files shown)


rohanbabbar04:
It was an issue with the intel scipy version for `python=3.10`, removing it from the ci file solves this problem. I also think that both scipy and numpy now usually check for `>=3.11`, which is what we should do as well.

mrava87:
Thanks @rohanbabbar04. Yes, no problem with dropping 3.10.

PR #725, opened 2026-01-02
Doc: use Literal type hint

mrava87:
This PR improves type hints across the entire library by using `Literal` when a set of unique values can be provided to an input variable.


pylops/basicoperators/spread.py
```diff
@@ -172,7 +172,7 @@ def __init__(
         dtable: Optional[NDArray] = None,
         fh: Optional[Callable] = None,
         interp: Optional[bool] = None,
-        engine: str = "numpy",
+        engine: Literal["numpy", "numba"] = "numpy",
```
mrava87: Use `Tengine2`

PR #726, opened 2026-01-11
CI: added py314 and py314t to GA

mrava87:
Add python 3.14 and 3.14 to Github Action CI


mrava87:
So far I can reproduce this locally... likely due to `devito` forcing pip/conda to install numpy v2.2.6... If I remove devito and force numpy=2.3.3, the errors go away.

The one in `kirchhoff.test_traveltime_ana` seems to be related to `dist_srcs = trav_srcs * vel` which also modifies `trav_srcs`... if I do `dist_srcs = trav_srcs.copy() * vel`, for numpy v2.2.6 the tests pass, but with numpy v2.3.3 I can just keep what I have and it's all fine.

So far, if I remove devito from the requirements (and its test), the CI works 😄 

**Leaving here until devito allows higher numpy version**

PR #728, opened 2026-01-12
test: add check on error message for all pytest.raises

### LINKED ISSUE #727: Improve tests with pytest.raises
Our test suite has a number of tests with `pytest.raises`; however, we do not check that the error raised is really what we expect. Following what suggested in https://github.com/PyLops/pylops/pull/719#issuecomment-3704749198, we should add checks on the actual error message, eg:

```
with pytest.raises(ValueError) as exception_info:
        ....
        
 error_message = str(exception_info.value)
 assert "text in message" in error_message
```

mrava87:
Fixes https://github.com/PyLops/pylops/issues/727

### DIFF
--- a/pylops/basicoperators/spread.py
+++ b/pylops/basicoperators/spread.py
@@ -127,10 +127,10 @@ class Spread(LinearOperator):
 
     Raises
     ------
-    KeyError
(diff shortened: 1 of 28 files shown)


mrava87:
@MothNik FYI 😄

PR #741, opened 2026-03-12
Build/CI: modernize build/doc/ci

mrava87:
This PR mimics https://github.com/PyLops/pyproximal/pull/241 for the PyLops repository:

- [x] update `pyproject.toml`
- [x]  add `uv.lock`*
-  [x] add targets in `Makefile`
-  [x] update GA (do `intel` in separate PR  as it seems tricky to enable it in pyproject.toml)
-  [x] update Azure Pipelines 
-  [x] update readthedocs

\* Currently created on OSX (Apple M3) with the following set of commands (having some issues with test_radon, but not the case in CI...):
```
uv venv --python 3.13
uv lock
uv sync --locked  --extra advanced  --extra stat --extra deep --all-groups 
```

It also modernizes the overall tooling:
- [x] move content of `setup.cfg` to `pyproject.toml`
- [x] remove `requirements*` files (must keep a few intel-GA is converted to uv)
- [x] switch from black/flake8/isort to ruff
- [x] bring down errors of ruff to zero (which including fixing some code patterns and type hints standards)
- [x] adapt coverage support with upload to codacy

Finally, it contains various improvements to the documentation:

-  [x] add installation instructions for uv
-  [x] use `sphinx-design` to have tabs in installation instructions for conda/pip/uv
-  [x] added emojii in main section titles


mrava87:
@cako this is quite big but closely follows what I did in pyproximal (plus a lot of minor systematic changes to files in `pylops` to get zero errors for ruff... we will likely never get that for mypy here 😢). 

Probably the most important change I would like to hear from you is `pyproject.toml`... I tried to make it work with also the `intel` environment but so far did not succeed (will leave it for a PR on its own so I don't keep messing up and reverting here...)

PR #748, opened 2026-05-16
Pin minconda version in Intel MKL Tests

rohanbabbar04:


rohanbabbar04:
Okay, the issue is with the latest version of `miniconda`. It is also an open [issue](https://github.com/conda-incubator/setup-miniconda/issues/501). I have added a workaround for this, and it seems to solve this issue.

mrava87 (approved):
Looks good 😄 we can monitor https://github.com/conda-incubator/setup-miniconda/issues/501 and see if they fix it without needing to pin a version...

PR #751, opened 2026-05-29
Fix elementwise variance accumulation in lsqr (closes #639)

### LINKED ISSUE #639: [Bug] `pylops.lsqr(..., calc_var=True)` returns wrong variances
## General description

While working on [this SciPy issue](https://github.com/scipy/scipy/issues/22232), I've found that the variances `var` returned by [`pylops.lsqr`](https://pylops.readthedocs.io/en/stable/api/generated/pylops.optimization.basic.lsqr.html) are flawed.

I wrote the following reproducible example:

```python
import numpy as np
from scipy.sparse import rand
from scipy.sparse.linalg import aslinearoperator
from scipy.sparse.linalg import lsqr as scipy_lsqr
from pylops import LinearOperator
from pylops import lsqr as pylops_lsqr

np.random.seed(0)
A = np.random.randn(20, 10)
A_as_linop = aslinearoperator(A)
b = np.random.rand(A.shape[0])

# solving with SciPy lsqr
scipy_lsqr_result = scipy_lsqr(
    A_as_linop,
    b,
    damp=0.0,
    atol=1e-8,
    btol=1e-8,
    calc_var=True,
    iter_lim=10_000,
)
scipy_lsqr_solution = scipy_lsqr_result[0]
scipy_ata_inv_main_diagonal = scipy_lsqr_result[9]

# solving with PyLops lsqr
pylops_lsqr_result = pylops_lsqr(
    LinearOperator(A_as_linop),
    b,
    damp=0.0,
    atol=1e-8,
    btol=1e-8,
    calc_var=True,
    niter=10_000,
)
pylops_lsqr_solution = pylops_lsqr_result[0]
pylops_ata_inv_main_diagonal = pylops_lsqr_result[9]

# checking against a dense solver
print(f"Rank of A: {np.linalg.matrix_rank(A)}")
print(f"Condition number of A: {np.linalg.cond(A)}", end="\n\n")
ata_inv_dense = np.linalg.inv((A.T @ A))
ata_inv_main_diagonal_from_dense = np.diag(ata_inv_dense)
dense_solution = ata_inv_dense @ A.T @ b

print(f"Solution of dense solver:\n{dense_solution}")
print(f"Residuals of dense solver:\n{A @ dense_solution - b}", end="\n\n")
print(f"Solution of SciPy lsqr:\n{scipy_lsqr_solution}")
print(f"Residuals of SciPy lsqr:\n{A @ scipy_lsqr_solution - b}", end="\n\n")
print(f"Solution of PyLops lsqr:\n{pylops_lsqr_solution}")
print(f"Residuals of PyLops lsqr:\n{A @ pylops_lsqr_solution - b}", end="\n\n")

assert np.allclose(scipy_lsqr_solution, dense_solution)  # passes
assert np.allclose(pylops_lsqr_solution, dense_solution)  # passes

print(
    f"Main diagonal of A.T @ A inverse from dense solver:\n{ata_inv_main_diagonal_from_dense}",
    end="\n\n",
)
print(
    f"Main diagonal of A.T @ A inverse from SciPy lsqr:\n{scipy_ata_inv_main_diagonal}",
    end="\n\n",
)
print(
    f"Main diagonal of A.T @ A inverse from PyLops lsqr:\n{pylops_ata_inv_main_diagonal}",
    end="\n\n",
)
```

This gives

```bash
Rank of A: 10
Condition number of A: 4.802155388010249

Solution of dense solver:
[ 0.20475999  0.47231818 -0.15393537 -0.10020819  0.03660469 -0.12091888
 -0.21749234  0.26228669 -0.05870751  0.28511243]
Residuals of dense solver:
[-0.72362251 -0.36987063  0.01528345  0.39789717 -0.28985235 -0.78734416
 -0.0801831  -0.45719266 -0.12740786  0.27938473 -0.29133013  0.13924614
 -0.94945666 -1.29884266 -0.1150791  -0.4779664   0.1087797  -0.77155471
 -0.38756407 -0.47259969]

Solution of SciPy lsqr:
[ 0.20475999  0.47231818 -0.15393537 -0.10020819  0.03660469 -0.12091888
 -0.21749234  0.26228669 -0.05870751  0.28511243]
Residuals of SciPy lsqr:
[-0.72362251 -0.36987063  0.01528345  0.39789717 -0.28985235 -0.78734416
 -0.0801831  -0.45719266 -0.12740786  0.27938473 -0.29133013  0.13924614
 -0.94945666 -1.29884266 -0.1150791  -0.4779664   0.1087797  -0.77155471
 -0.38756407 -0.47259969]

Solution of PyLops lsqr:
[ 0.20475999  0.47231818 -0.15393537 -0.10020819  0.03660469 -0.12091888
 -0.21749234  0.26228669 -0.05870751  0.28511243]
Residuals of PyLops lsqr:
[-0.72362251 -0.36987063  0.01528345  0.39789717 -0.28985235 -0.78734416
 -0.0801831  -0.45719266 -0.12740786  0.27938473 -0.29133013  0.13924614
 -0.94945666 -1.29884266 -0.1150791  -0.4779664   0.1087797  -0.77155471
 -0.38756407 -0.47259969]

Main diagonal of A.T @ A inverse from dense solver:
[0.08555239 0.11526442 0.14234157 0.04237757 0.13102765 0.07080344
 0.115843   0.08861612 0.10540813 0.16258423]

Main diagonal of A.T @ A inverse from SciPy lsqr:
[0.08555239 0.11526442 0.14234157 0.04237757 0.13102765 0.07080344
 0.115843   0.08861612 0.10540813 0.16258423]

Main diagonal of A.T @ A inverse from PyLops lsqr:
[1.05981851 1.05981851 1.05981851 1.05981851 1.05981851 1.05981851
 1.05981851 1.05981851 1.05981851 1.05981851]
```

## Suspected Cause

On closer inspection, it turns out that the variances of `pylops.lsqr` are all identical.
I inspected the code and found ([Link](https://github.com/PyLops/pylops/blob/dev/pylops/optimization/cls_basic.py#L821-L824)):

```python
        if self.calc_var:
            self.var = self.var + to_numpy_conditional(
                self.var, self.ncp.dot(self.dk, self.dk)
            )
```

When one compares this to the equation on page 53 of the original publication ([Link](https://www-leland.stanford.edu/group/SOL/software/lsqr/lsqr-toms82a.pdf#page=11)):

Image

one can see an index dependency for `dk`, i.e., each variance value needs to updated individually which makes sense.

## Suggested Fix

The product `self.ncp.dot(self.dk, self.dk)` is the same as `square(self.dk).sum()` since `self.dk` should be a 1d vector. The result of this is a scalar and this explains why all variances are identical.
Looking at the publication and [the SciPy code](https://github.com/scipy/scipy/blob/v1.15.1/scipy/sparse/linalg/_isolve/lsqr.py#L473-L474), the update should probably be `self.var = self.var + square(self.dk)` where `square` should be linked to `self.ncp` if required.

gaoflow:
## Problem

`lsqr`'s optional `var` output estimates the diagonal of $(A^H A)^{-1}$ — i.e. one variance per unknown. The accumulation step used

```python
self.var = self.var + to_numpy_conditional(self.var, self.ncp.dot(self.dk, self.dk))
```

but `ncp.dot(self.dk, self.dk)` is the **scalar** sum of squares of `dk`, so the *same* number is added to every entry of `var`. The result is that all returned variances are identical (and wrong), as reported in #639.

### Reproduction (from the issue)

```python
import numpy as np
from scipy.sparse.linalg import aslinearoperator, lsqr as sp_lsqr
from pylops import LinearOperator, lsqr

np.random.seed(0)
A = np.random.randn(20, 10); b = np.random.rand(20)
Aop = LinearOperator(aslinearoperator(A))

var_pylops = lsqr(Aop, b, atol=1e-8, btol=1e-8, calc_var=True, niter=10_000)[9]
var_scipy  = sp_lsqr(aslinearoperator(A), b, atol=1e-8, btol=1e-8, calc_var=True, iter_lim=10_000)[9]
var_dense  = np.diag(np.linalg.inv(A.T @ A))
```

Before this PR `var_pylops` is a constant vector (`[1.0598, 1.0598, ...]`), while `var_scipy` and `var_dense` agree.

## Fix

Accumulate the **elementwise** square instead:

```python
self.var = self.var + to_numpy_conditional(self.var, self.ncp.square(self.dk))
```

This matches the recurrence in Paige & Saunders (1982, p.53) and [scipy's `lsqr`](https://github.com/scipy/scipy/blob/v1.15.1/scipy/sparse/linalg/_isolve/lsqr.py#L473-L474). `ncp.square` is available across the NumPy / CuPy backends used elsewhere in this method, and the solution path is untouched — only the `var` output changes. Credit to @IruNikZe for the diagnosis in #639.

> Note: like scipy, `var` is an *iterative* estimate of $\mathrm{diag}((A^H A)^{-1})$ and converges as iterations proceed; the identity-operator/one-iteration edge case raised later in the issue is inherent to the LSQR estimator (scipy behaves the same) and is not addressed here.

## Tests

Added `test_lsqr_calc_var` (overdetermined real, full column rank) asserting the variances are **not** all identical (guards against the regression) and match both scipy's `lsqr` and the dense `diag(inv(A^H A))`.

Verified locally: `pytest pytests/test_solver.py` (98 passed), `ruff check` clean, flake8 (max-line 88) clean on changed files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+# 2.8.0
+* Fixed element-wise variance accumulation in `pylops.optimization.cls_basic.LSQR`
+  and `pylops.lsqr`, which previously assigned the same value to every entry of
+  `var` (#639)
+
 # 2.7.0
 * Added cubic spline interpolation operator via
   `pylops.signalprocessing.interpspline.InterpCubicSpline` (also interfaceable via
@@ -17,9 +22,6 @@ Changelog
   and `pylops.waveeqprocessing.Marchenko`
 * Improved typing annotations across all operators (and enforced use of
   Literal for parameters with multiple options)
-* Fixed element-wise variance accumulation in `pylops.optimization.cls_basic.LSQR`
-  and `pylops.lsqr`, which previously assigned the same value to every entry of
-  `var` (#639)
 
 # 2.6.0
 * Added `pylops.medical` module and `pylops.medical.CT2D` operator
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ Changelog
   and `pylops.waveeqprocessing.Marchenko`
 * Improved typing annotations across all operators (and enforced use of
   Literal for parameters with multiple options)
+* Fixed element-wise variance accumulation in `pylops.optimization.cls_basic.LSQR`
+  and `pylops.lsqr`, which previously assigned the same value to every entry of
+  `var` (#639)
 
 # 2.6.0
 * Added `pylops.medical` module and `pylops.medical.CT2D` operator
--- a/pylops/optimization/cls_basic.py
+++ b/pylops/optimization/cls_basic.py
@@ -1248,8 +1248,13 @@ class LSQR(Solver):
             self.ncp.add(self.v, self.w, out=self.w)
         self.ddnorm = self.ddnorm + self.ncp.linalg.norm(self.dk) ** 2.0
         if self.calc_var:
+            # ``var`` estimates the diagonal of (A^H A)^-1, so it must accumulate
+            # the *element-wise* square of ``dk`` (one variance per unknown), as in
+            # Paige & Saunders (1982) eq. on p.53 and scipy's lsqr. Using
+            # ``dot(dk, dk)`` instead adds the same scalar (sum of squares) to every
+            # entry, yielding identical, wrong variances (issue #639).
             self.var = self.var + to_numpy_conditional(
-                self.var, self.ncp.dot(self.dk, self.dk)
+                self.var, self.ncp.square(self.dk)
             )
 
         # use a plane rotation on the right to eliminate the
--- a/pylops/optimization/cls_basic.py
+++ b/pylops/optimization/cls_basic.py
@@ -1248,11 +1248,7 @@ class LSQR(Solver):
             self.ncp.add(self.v, self.w, out=self.w)
         self.ddnorm = self.ddnorm + self.ncp.linalg.norm(self.dk) ** 2.0
         if self.calc_var:
-            # ``var`` estimates the diagonal of (A^H A)^-1, so it must accumulate
-            # the *element-wise* square of ``dk`` (one variance per unknown), as in
-            # Paige & Saunders (1982) eq. on p.53 and scipy's lsqr. Using
-            # ``dot(dk, dk)`` instead adds the same scalar (sum of squares) to every
-            # entry, yielding identical, wrong variances (issue #639).
+            # accumulate the element-wise square of dk (one variance per unknown)
             self.var = self.var + to_numpy_conditional(
                 self.var, self.ncp.square(self.dk)
             )
--- a/pytests/test_solver.py
+++ b/pytests/test_solver.py
@@ -379,11 +379,10 @@ def test_lsqr_pylops_scipy(par):
 
 @pytest.mark.parametrize("par", [(par3), (par4)])
 def test_lsqr_calc_var(par):
-    """LSQR ``var`` estimates the diagonal of (A^H A)^-1 element-wise (issue #639).
+    """LSQR ``var`` estimates the diagonal of (A^H A)^-1 element-wise.
 
-    Each unknown must get its own variance; a regression here would make all
-    entries identical (the previous ``dot(dk, dk)`` bug). Compare against scipy's
-    LSQR and the dense ``diag(inv(A^H A))`` on a full-column-rank system.
+    Each unknown must get its own variance; a regression would make all entries
+    identical. scipy's LSQR is used as the reference implementation.
     """
     np.random.seed(10)
 
@@ -391,17 +390,15 @@ def test_lsqr_calc_var(par):
     Aop = MatrixMult(A, dtype=par["dtype"])
     y = Aop * (np.ones(par["nx"]))
 
-    niter = 10 * par["nx"]
+    # niter is only a ceiling (atol/btol stop earlier); use scipy's default of
+    # 2 * nx so both implementations run for the same number of iterations.
+    niter = 2 * par["nx"]
     var = lsqr(Aop, y, x0=None, niter=niter, atol=1e-8, btol=1e-8)[9]
-    var_sp = sp_lsqr(
-        Aop, y, iter_lim=niter, atol=1e-8, btol=1e-8, calc_var=True
-    )[9]
-    var_dense = np.diag(np.linalg.inv(np.conj(A.T) @ A))
+    var_sp = sp_lsqr(Aop, y, iter_lim=niter, atol=1e-8, btol=1e-8, calc_var=True)[9]
 
-    # not all identical (the bug produced a constant vector)
+    # the dot(dk, dk) bug produced a constant vector instead (issue #639)
     assert not np.allclose(var, var[0])
     assert_array_almost_equal(var, var_sp, decimal=6)
-    assert_array_almost_equal(var, var_dense, decimal=6)
 
 
 @pytest.mark.parametrize(
--- a/pytests/test_solver.py
+++ b/pytests/test_solver.py
@@ -382,11 +382,8 @@ def test_lsqr_pylops_scipy(par):
 )
 @pytest.mark.parametrize("par", [(par3), (par4)])
 def test_lsqr_calc_var(par):
-    """LSQR ``var`` estimates the diagonal of (A^H A)^-1 element-wise.
-
-    Each unknown must get its own variance; a regression would make all entries
-    identical. scipy's LSQR is used as the reference implementation.
-    """
+    """Compare PyLops and scipy LSQR variance computation for the diagonal of
+    (A^H A)^-1 (issue #639)."""
     np.random.seed(10)
 
     A = np.random.normal(0, 1, (par["ny"], par["nx"]))
@@ -399,7 +396,6 @@ def test_lsqr_calc_var(par):
     var = lsqr(Aop, y, x0=None, niter=niter, atol=1e-8, btol=1e-8)[9]
     var_sp = sp_lsqr(Aop, y, iter_lim=niter, atol=1e-8, btol=1e-8, calc_var=True)[9]
 
-    # the dot(dk, dk) bug produced a constant vector instead (issue #639)
     assert not np.allclose(var, var[0])
     assert_array_almost_equal(var, var_sp, decimal=6)
 
--- a/pytests/test_solver.py
+++ b/pytests/test_solver.py
@@ -377,6 +377,9 @@ def test_lsqr_pylops_scipy(par):
         assert_array_almost_equal(xinv_sp, x, decimal=4)
 
 
+@pytest.mark.skipif(
+    int(os.environ.get("TEST_CUPY_PYLOPS", 0)) == 1, reason="Not CuPy enabled"
+)
 @pytest.mark.parametrize("par", [(par3), (par4)])
 def test_lsqr_calc_var(par):
     """LSQR ``var`` estimates the diagonal of (A^H A)^-1 element-wise.
--- a/pytests/test_solver.py
+++ b/pytests/test_solver.py
@@ -377,6 +377,33 @@ def test_lsqr_pylops_scipy(par):
         assert_array_almost_equal(xinv_sp, x, decimal=4)
 
 
+@pytest.mark.parametrize("par", [(par3), (par4)])
+def test_lsqr_calc_var(par):
+    """LSQR ``var`` estimates the diagonal of (A^H A)^-1 element-wise (issue #639).
+
+    Each unknown must get its own variance; a regression here would make all
+    entries identical (the previous ``dot(dk, dk)`` bug). Compare against scipy's
+    LSQR and the dense ``diag(inv(A^H A))`` on a full-column-rank system.
+    """
+    np.random.seed(10)
+
+    A = np.random.normal(0, 1, (par["ny"], par["nx"]))
+    Aop = MatrixMult(A, dtype=par["dtype"])
+    y = Aop * (np.ones(par["nx"]))
+
+    niter = 10 * par["nx"]
+    var = lsqr(Aop, y, x0=None, niter=niter, atol=1e-8, btol=1e-8)[9]
+    var_sp = sp_lsqr(
+        Aop, y, iter_lim=niter, atol=1e-8, btol=1e-8, calc_var=True
+    )[9]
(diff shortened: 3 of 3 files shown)


mrava87:
@IruNikZe I was wondering if you got anywhere with this in scipy (and pylops), and if you will be willing to review this PR?

MothNik:
@mrava87 I didn't get anywhere yet, unfortunately. Seems the variances converge (way) slower than the solution and there is not so much that can be done about that.

@gaoflow @mrava87 I'm however happy to review this PR this week.

mrava87:
@MothNik thanks a lot for the feedback and willingness to review this PR.

What you say makes somehow sense to me, as usually also in VI type of solvers the mean always converges faster than the std, so not surprised this may be also the case for a solver like LSQR

pylops/optimization/cls_basic.py
```diff
@@ -1248,8 +1248,13 @@ def step(self, x: NDArray, show: bool = False) -> NDArray:
             self.ncp.add(self.v, self.w, out=self.w)
         self.ddnorm = self.ddnorm + self.ncp.linalg.norm(self.dk) ** 2.0
         if self.calc_var:
+            # ``var`` estimates the diagonal of (A^H A)^-1, so it must accumulate
```
MothNik: I think this extensive comment is not needed here. 
There is no change in, e.g., a convention, that would require documentation about previous behaviour and why the fix was needed. In this case, it's just an exchange of one function for another to fix a bug, which should then ultimately be gone.

Mentioning the issue in the tests below feels more natural imo.

pytests/test_solver.py
```diff
@@ -377,6 +377,33 @@ def test_lsqr_pylops_scipy(par):
         assert_array_almost_equal(xinv_sp, x, decimal=4)
 
 
+@pytest.mark.parametrize("par", [(par3), (par4)])
+def test_lsqr_calc_var(par):
+    """LSQR ``var`` estimates the diagonal of (A^H A)^-1 element-wise (issue #639).
+
+    Each unknown must get its own variance; a regression here would make all
+    entries identical (the previous ``dot(dk, dk)`` bug). Compare against scipy's
+    LSQR and the dense ``diag(inv(A^H A))`` on a full-column-rank system.
```
MothNik: Is the comparison against the dense case testing the desired functionality?
It is not necessarily guaranteed that LSQR can estimate the variance elements accurately when compared to the true dense solution.

A test against the `scipy` implementation should be sufficient because that shows whether the algorithm has the same behaviour as the reference implementation.

pytests/test_solver.py
```diff
@@ -377,6 +377,33 @@ def test_lsqr_pylops_scipy(par):
...
+    A = np.random.normal(0, 1, (par["ny"], par["nx"]))
+    Aop = MatrixMult(A, dtype=par["dtype"])
+    y = Aop * (np.ones(par["nx"]))
+
+    niter = 10 * par["nx"]
+    var = lsqr(Aop, y, x0=None, niter=niter, atol=1e-8, btol=1e-8)[9]
+    var_sp = sp_lsqr(
+        Aop, y, iter_lim=niter, atol=1e-8, btol=1e-8, calc_var=True
+    )[9]
+    var_dense = np.diag(np.linalg.inv(np.conj(A.T) @ A))
```
MothNik: See [this comment](https://github.com/PyLops/pylops/pull/751/changes#r3357415866)

pytests/test_solver.py
```diff
@@ -377,6 +377,33 @@ def test_lsqr_pylops_scipy(par):
...
+
+    # not all identical (the bug produced a constant vector)
+    assert not np.allclose(var, var[0])
+    assert_array_almost_equal(var, var_sp, decimal=6)
+    assert_array_almost_equal(var, var_dense, decimal=6)
```
MothNik: See [this comment](https://github.com/PyLops/pylops/pull/751/changes#r3357415866)

pytests/test_solver.py
```diff
@@ -377,6 +377,33 @@ def test_lsqr_pylops_scipy(par):
...
+    entries identical (the previous ``dot(dk, dk)`` bug). Compare against scipy's
+    LSQR and the dense ``diag(inv(A^H A))`` on a full-column-rank system.
+    """
+    np.random.seed(10)
+
+    A = np.random.normal(0, 1, (par["ny"], par["nx"]))
+    Aop = MatrixMult(A, dtype=par["dtype"])
+    y = Aop * (np.ones(par["nx"]))
+
+    niter = 10 * par["nx"]
```
MothNik: This high `niter` suggests a very high number of iterations and thus high accuracy of the result.
However, when I run
```python
    res = lsqr(Aop, y, x0=None, niter=niter, atol=1e-8, btol=1e-8)
    print(f"{niter = }; {res[2] = }")
```
here, I get `niter = 110; res[2] = 11`. So the algorithm terminates before reaching `niter`, which serves solely as a maximum ceiling for the number of iterations (for safety).

So, I'd suggest to either:
1. add a comment saying that `niter` is set very high to avoid premature termination, but not to improve accuracy
2. remove this one and go with the defaults, which is `2 * par["nx"]` (see [here](https://github.com/scipy/scipy/blob/v1.17.0/scipy/sparse/linalg/_isolve/lsqr.py#L330-L331))

pytests/test_solver.py
```diff
@@ -377,6 +377,33 @@ def test_lsqr_pylops_scipy(par):
...
+    var = lsqr(Aop, y, x0=None, niter=niter, atol=1e-8, btol=1e-8)[9]
+    var_sp = sp_lsqr(
+        Aop, y, iter_lim=niter, atol=1e-8, btol=1e-8, calc_var=True
+    )[9]
+    var_dense = np.diag(np.linalg.inv(np.conj(A.T) @ A))
+
+    # not all identical (the bug produced a constant vector)
```
MothNik: I think the issue number can be mentioned here as well because now "`the bug`" is a bit out of context.

pytests/test_solver.py
```diff
@@ -377,6 +377,33 @@ def test_lsqr_pylops_scipy(par):
         assert_array_almost_equal(xinv_sp, x, decimal=4)
 
 
+@pytest.mark.parametrize("par", [(par3), (par4)])
+def test_lsqr_calc_var(par):
+    """LSQR ``var`` estimates the diagonal of (A^H A)^-1 element-wise (issue #639).
```
MothNik: I think the `(issue #639)` can be removed from here because this test makes sense in general and not just because the issue was observed.

MothNik (changes requested):
Please consider also updating the `Changelog.md` (see [here](https://github.com/PyLops/pylops/blob/dev/CHANGELOG.md)) to reflect your changes.

pytests/test_solver.py
MothNik: Maybe it even makes sense to forcefully clip `niter` to a too low value to make sure `pylops` and `scipy` use the exact same number of iterations instead of maybe differing by 1 step due to some floating point difference. 🤔

binggao1230:
Thanks @MothNik, all addressed in the latest commit:

- Condensed the inline comment in `LSQR` to a single line.
- Dropped the dense `diag(inv(A^H A))` comparison — you're right that LSQR isn't guaranteed to match it; the test now compares only against scipy's LSQR as the reference.
- Good catch on `niter`: switched to `2 * nx` (scipy's default) and added a note that it's only a ceiling since `atol`/`btol` stop earlier.
- Moved the `#639` reference out of the docstring into the specific "constant vector" comment.
- Added a `CHANGELOG.md` entry under 2.7.0.

binggao1230:
Pushed 515e51a to fix the CuPy CI failure: `test_lsqr_calc_var` compares against scipy's `lsqr` as the reference, but scipy can't implicitly convert cupy arrays (`TypeError: Implicit conversion to a NumPy array is not allowed`). Guarded it with the same `TEST_CUPY_PYLOPS` skipif that `test_lsqr_pylops_scipy` already uses for the same reason. Passes on the numpy backend; skipped under CuPy.

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ Changelog
   and `pylops.waveeqprocessing.Marchenko`
 * Improved typing annotations across all operators (and enforced use of
   Literal for parameters with multiple options)
+* Fixed element-wise variance accumulation in `pylops.optimization.cls_basic.LSQR`
```
mrava87: No, this should be already added to a 2.8.0 section

pytests/test_solver.py
```diff
@@ -377,6 +377,33 @@ def test_lsqr_pylops_scipy(par):
         assert_array_almost_equal(xinv_sp, x, decimal=4)
 
 
+@pytest.mark.skipif(
+    int(os.environ.get("TEST_CUPY_PYLOPS", 0)) == 1, reason="Not CuPy enabled"
+)
+@pytest.mark.parametrize("par", [(par3), (par4)])
+def test_lsqr_calc_var(par):
+    """LSQR ``var`` estimates the diagonal of (A^H A)^-1 element-wise.
```
mrava87: Do not mention regression, just say `compare PyLops and scipy LSQR variance computation for the diagonal of (A^H A)^-1`

pytests/test_solver.py
```diff
@@ -377,6 +377,33 @@ def test_lsqr_pylops_scipy(par):
...
+    Aop = MatrixMult(A, dtype=par["dtype"])
+    y = Aop * (np.ones(par["nx"]))
+
+    # niter is only a ceiling (atol/btol stop earlier); use scipy's default of
+    # 2 * nx so both implementations run for the same number of iterations.
+    niter = 2 * par["nx"]
+    var = lsqr(Aop, y, x0=None, niter=niter, atol=1e-8, btol=1e-8)[9]
+    var_sp = sp_lsqr(Aop, y, iter_lim=niter, atol=1e-8, btol=1e-8, calc_var=True)[9]
+
+    # the dot(dk, dk) bug produced a constant vector instead (issue #639)
```
mrava87: remove this comment, just mention issue #639 in the docstring on the test.

mrava87 (changes requested):

mrava87:
@gaoflow left a few final comments 😄

CHANGELOG.md
binggao1230: Done — moved this entry into a new `# 2.8.0` section at the top of the changelog.

pytests/test_solver.py
binggao1230: Done — docstring now reads: compare PyLops and scipy LSQR variance computation for the diagonal of (A^H A)^-1.

pytests/test_solver.py
binggao1230: Done — removed the inline comment; the test docstring now just references issue #639.

mrava87 (approved):

PR #753, opened 2026-05-30
ci: fix cupy numba PTX version mismatch on self-hosted runner

hongyx11:
## Problem

The CuPy CI (`buildcupy.yaml`) had **19 failing tests** in `test_fourierradon`, `test_kirchhoff`, and `test_nonstatconvolve`, all with:

```
numba.cuda.cudadrv.driver.LinkerError: [222] Call to cuLinkAddData results in CUDA_ERROR_UNSUPPORTED_PTX_VERSION
ptxas application ptx input, line 9; fatal : Unsupported .version 8.8; current version is '8.7'
```

## Root cause

numba compiles `@cuda.jit` kernels using the `libNVVM` it finds via `CUDA_HOME`. On the self-hosted runner that pointed at a **system CUDA 12.9** toolkit, whose NVVM emits PTX `.version 8.8`. The GPU **driver is 12.8**, and its in-driver JIT linker (`cuLinkAddData`) only accepts PTX ≤ `8.7` → the kernels fail to link.

This regressed when the CuPy CI switched to `uv sync` (commit 7697d6e); the previous pip flow never pulled a 12.9 toolchain. cupy itself is unaffected — only numba kernels.

## Fix

Force numba to use the **pip-installed CUDA 12.8 toolkit** (matches the 12.8 driver, emits PTX 8.7):

- **`pyproject.toml`**: pin `nvidia-cuda-nvcc-cu12<12.9` in the `gpu-cu12` extra (+ relock `uv.lock` → 12.8.93).
- **`buildcupy.yaml` test step**: `export CUDA_HOME` to the wheel's `cuda_nvcc` dir.
- **`buildcupy.yaml` install step**: add a versioned `libnvvm.so.4` symlink — numba's `find_lib` regex only matches `libnvvm.so.N`, but the wheel ships a bare `libnvvm.so`.

## Verification

Ran the exact workflow steps on an L40S GPU node:

- Before: `19 failed, 2139 passed, 881 skipped`
- After: **`2156 passed, 0 failed, 881 skipped`**

Offline check confirms the wheel NVVM emits `.version 8.7` (was `8.8`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mrava87:
Solved in https://github.com/PyLops/pylops/pull/754

PR #773, opened 2026-07-19
Feat: UDCT operator

mrava87:
## Objective

This PR introduces the `UDCT` operator.

## Description

The `UDCT` operator is introduced as a thin wrapper around the `curvelets.numpy.UDCT` class in https://curvelets.readthedocs.io. 

Since the curvelets library exposes only a class that applies the operator over all dimensions of a nd-array, the `UDCT` operator follows the same rationale. If one wants to apply it to a subset of dimensions, this can be easily achieved by leveraging the Kronecker (and Transpose) operators in PyLops as shown in the example.

This approach triggered interest in supporting multiprocess/multithread in `matmat`/`rmatmat` like in the stack operators - see


cako:
@mrava87 [ahead of time coefficient implemented as of 1.2](https://curvelets.readthedocs.io/en/stable/auto_examples/plot_01_zone_plate.html#inspecting-coefficient-shapes-ahead-of-time)!

mrava87:
> @mrava87 [ahead of time coefficient implemented as of 1.2](https://curvelets.readthedocs.io/en/stable/auto_examples/plot_01_zone_plate.html#inspecting-coefficient-shapes-ahead-of-time)!

Great, let me add it and make 1.2 the minimum version we support 😄

mrava87:
@cako did is ready for you to review. I now use `coefficient_shapes` to avoid a dry-run to identify the size of coeffs.

Note that initially I kind of copy-pasted https://github.com/PyLops/curvelops/blob/main/curvelops/curvelops.py but when I was almost done I realized that a lot of boilerplate code was used to handle trivially extra dimensions (a naive for...loop) as internally curvelets do not handle them as far as I can tell - aka what define in `dims` is the shape of what you pass and all axes are transformed... this ultimately led me to leverage `Knonecker`+`Transpose` and I created this Github Issue that I will try to tackle soon to gain from the fact that  `Knonecker` is embarrassingly parallel like all of the stacking operators (https://github.com/PyLops/pylops/issues/775)

cako (approved):
LGTM! Thanks @mrava87 !!

PR #774, opened 2026-07-19
Feat: 3D extension to Structure Tensor slope estimation

mrava87:
This PR extended `slope_estimate` to 3D inputs.

It also includes some analytical tests both for the 2D and 3D scenario against plane waves with fixed slope/dip.


cako (approved):
Added a few changes, have a look! Otherwise looks good to me!

mrava87:
> Added a few changes, have a look! Otherwise looks good to me!

Thanks, looks good!

Definitely not a big fan of having different number and type of returns but without breaking backward compatibility I wasn't sure it was possible to avoid that... good addition with the overloads 😉

PR #780, opened 2026-07-26
Feat: added MultiOperator and multiproc/multithread support to matmat/rmatmat

mrava87:
This PR aims to extend the multiproc/multithread capabilities in PyLops. 

It implements:
- [x] unified base class `MultiOperator` for all operators with multiproc/multithread support (adopted by `BlockDiag`, `HStack`, `VStack`
- [x] multiproc/multithread support in matmat/rmatmat methods
- [x] multiproc/multithread support in `Kronecker`
- [x] tests for matmat/rmatmat method of a generic operator (eg MatrixMult) and `Kronecker` with multiproc/multithread

Some examples and timings for the new multiproc/multithread support in matmat/rmatmat and Kronecker are in https://github.com/PyLops/pylops_notebooks/blob/master/developement/Basic_multi.ipynb

### DIFF
--- a/pylops/_multioperator.py
+++ b/pylops/_multioperator.py
@@ -1,0 +1,18 @@
+import threading
+from collections.abc import Callable
+
+from pylops.utils.typing import NDArray
+
+
+def _matvec_rmatvec_map(op: Callable[[NDArray], NDArray], x: NDArray) -> NDArray:
+    """matvec/rmatvec for multiprocessing / multithreading"""
+    return op(x).squeeze()
+
+
+def _matvec_rmatvec_map_mt(
+    op: Callable[[NDArray], NDArray], x: NDArray, y: NDArray, lock: threading.Lock
+) -> None:
+    """rmatvec for multithreading with lock"""
+    ylocal = op(x).squeeze()
+    with lock:
+        y[:] += ylocal
--- a/pylops/linearoperator.py
+++ b/pylops/linearoperator.py
@@ -32,16 +32,12 @@ else:
     from scipy.sparse._sputils import isintlike, isshape
 
 from pylops import get_ndarray_multiplication
+from pylops._multioperator import _matvec_rmatvec_map
 from pylops.optimization.basic import cgls
 from pylops.utils.backend import get_array_module, get_module, get_sparse_eye
 from pylops.utils.decorators import count
 from pylops.utils.estimators import trace_hutchinson, trace_hutchpp, trace_nahutchpp
 from pylops.utils.typing import DTypeLike, InputDimsLike, NDArray, ShapeLike, Tpool
-
-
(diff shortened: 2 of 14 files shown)


mrava87:
@cako if you have time, appreciate any feedback on this 😄 

The initial motivation was to support multiproc/threading in `Kronecker` for the UDCT operator (or any operator that does not internally handle broadcasting over dimensions), but when I started working on it I realized that we had a lot of code repetitions over the few operators already supporting multiproc/threading and so did a bit of code restructuring with a new base class...

pylops/_multioperator.py
```diff
@@ -0,0 +1,18 @@
...
+    return op(x).squeeze()
+
+
+def _matvec_rmatvec_map_mt(
+    op: Callable[[NDArray], NDArray], x: NDArray, y: NDArray, lock: threading.Lock
+) -> None:
+    """rmatvec for multithreading with lock"""
+    ylocal = op(x).squeeze()
+    with lock:
+        y[:] += ylocal
```
cako, line 18: I know this logic already existed but I can see this becoming a bottleneck for large parallelization. I think a better strategy would be to let the multioperator return its local result and the main thread/process can finish summing. Because at that point its a single assessor writing, there's no need for a lock.

pylops/linearoperator.py
```diff
@@ -458,31 +461,101 @@ def _rmatvec(self, x: NDArray) -> NDArray:
...
+        return y
+
+    def _matmat_multiproc(self, X: NDArray, pool: Pool) -> NDArray:
+        """Matrix-matrix multiplication (multiprocess version)"""
+        ys = pool.starmap(
+            _matvec_rmatvec_map,
+            [(self._matvec, col.reshape(-1)) for col in X.T],
+        )
+        y = np.vstack(ys).T
+        return y
```
cako, line 505: Is this part missing `sp.sparse.issparse(X)` special treatment?

cako (commented):
Overall looks pretty good, I just left a couple of suggestions!

pylops/_multioperator.py
mrava87: Interesting! 

I have a feeling that what you suggest was my initial approach (basically copy-paste from multiprocess) but then I tried this one for threading and paid off... but I can't find any proper performance test I did with the two approaches.

Let me write Github Issue about this and I will try both options with a proper comparison - https://github.com/PyLops/pylops/issues/786

pylops/linearoperator.py
mrava87: Ops, I may have forgotten 😢  let me add it (and add a test!)

mrava87: https://github.com/PyLops/pylops/pull/780/changes/44ae88e0f2e723ea95a555e18f0752959520ead6

PR #794, opened 2026-08-30
Feat: added Downsample2D operator

mrava87:
This PR adds a new Claude skill called `newop` that can be used to ask Claude to create a PyLops operator from a detailed mathematical description and/or a plain implementation of forward/adjoint.

It is stress-tested with a rather popular operator, namely a downsampling operator for images called `Downsample2D`. The first implementation is created from the following prompt:

```
❯ Create a new operator in signalprocessing called Downsample2D that implements 2d downsampling as the combination
  of gaussian filtering and subsampling in both directions.
```

and slightly refined to match PyLops standards (mostly in the way parameters are checked and test suite).


mrava87:
@claude review.

PR #795, opened 2026-09-02
Fix: change convention for Downsample2d default sigma

mrava87:
This PR changes the convention for the default `sigma` in `Downsample2D` choosing a value where the frequency spectrum of the gaussian filter is -10dB at the new Nyquist frequency. The old convention was somehow empirical and not 'fixed' for different downsampling factors.

### DIFF
--- a/docs/source/api/others.rst
+++ b/docs/source/api/others.rst
@@ -39,6 +39,16 @@ Describe
 
     describe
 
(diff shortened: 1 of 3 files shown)


mrava87:
@claude review.

mrava87:
@claude review again.

PR #799, opened 2026-09-15
Fix: improved handling of dtype in LinearOperator

mrava87:
## Summary

When moving to scipy >= 1.18 in a local environment , I noticed that operators combined through `HStack`, `VStack` or `BlockDiag` without an explicit `dtype` ended up with a string dtype (`dtype('<U7')`). `BlendingGroup` and `BlendingHalf` with the default `dtype="float64"`, as an exampled, started to fail. 

This PR makes `LinearOperator.dtype` always a `numpy.dtype` and stops relying on scipy's private `_get_dtype` (in the entire library).

**NOTE**: this is somehow a preemptive change as right now we can't update `uv.lock` to use `scipy==v1.18.1` as this still leads to some installation errors in the free-threaded python CI.

## Problem

- `LinearOperator` stored `dtype` exactly as given. For example, `Shift` sets `Op.dtype = dtype` after construction, so with the default `dtype="float64"` its dtype was the Python string `"float64"`.
- `BlendingGroup`/`BlendingHalf` build `HStack(...)`/`BlockDiag(...)` without `dtype`, so the dtype was inferred with `scipy.sparse.linalg._interface._get_dtype`.
- scipy 1.18 rebuilt `_get_dtype` on the array API `xp_result_type`, which treats a Python string as a string scalar and returns `dtype('<U7')`. Older scipy used `np.result_type`, which parsed the string as a dtype name.
- Any code that creates inputs with `op.dtype` (e.g., the asv benchmarks) then fails inside the FFT with `DTypePromotionError`.
- The test suite did not catch it because `test_blending.py` only passed `np.float64`/`np.float32` (types, not strings).

## Changes

- **`LinearOperator.dtype` is now a property.** Its setter always stores `np.dtype(value)`, both at initialization and when `dtype` is assigned later (this covers `Shift`, `PhaseShift`, `MDC`, ...). If neither `dtype` nor `Op` is provided, `dtype` now defaults to `numpy.float64`. Previously it was left unset, so for example `print(LinearOperator(shape=(10, 20)))` raised `AttributeError`.
- **`_get_dtype` moved to `pylops/utils/_internal.py`.** It casts every input with `np.dtype` before `np.result_type` and no longer mutates the `dtypes` list passed to it. `HStack`, `VStack` and `BlockDiag` now use it instead of scipy's private `_get_dtype`.
- **Docs:** the `dtype` entry of the `LinearOperator` docstring describes the casting and the default. `docs/source/adding.rst` now only shows initialization through `super().__init__` and states that these attributes must be passed there rather than assigned on `self`.
- **`newop` skill:** same guidance; the template passes `dtype=dtype` (no `np.dtype` wrapping needed).

## Backward compatibility

- Unchanged: `op.dtype == "float64"`, `op.dtype in ("float32", "float64")`, `str(op.dtype)`, `np.ones(n, dtype=op.dtype)`.
- Changed (negative regression): a `dtype` assigned on `self` before calling `super().__init__()` without passing `dtype` is overwritten by the `numpy.float64` default. *Checked* to never be the case for any linear operator defined in the library itself.
- Changed (negative regression): for operators that previously stored a string, code that relies on `op.dtype` being a `str` now behaves differently: `isinstance(op.dtype, str)`, using it as a dict key or set member, string methods, `json.dumps(op.dtype)`.
- Changed (positive regression): custom subclasses passing an invalid dtype now raise `TypeError: data type 'foo' not understood` at construction, as built-in operators already did.

I believe that the benefits of this outweighs the two negative regressions - in both cases, in fact, the behavior that changes was anyways not encouraged (no code/example/tutorial in this codebase did it).

## Tests

- `pytests/test_linearoperator.py`: added `test_dtype_normalization` (string dtypes, assignment after creation, `_get_dtype`, `HStack`/`VStack`/`BlockDiag` with a string-dtype `Shift`) and `test_dtype_default`.
- `pytests/test_blending.py`: added a string-dtype case (`"float64"`) to `test_Blending_group` and `test_Blending_half`, asserting that the operator dtype is a `numpy.dtype`.
- Checked locally with scipy 1.18.1: the new tests fail on `dev` (`assert dtype('<U7') == dtype('float64')`) and pass with this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mrava87:
@cako can I ask your opinion on this 😄 

As I (and Claude) explained in the summary, this was a rather casual finding whilst working on some examples for the new https://github.com/PyLops/pylops-asv/tree/main repo, but I felt it was a significant enough soon-to-appear-bug (with the landing of scipy v1.18.1) to be ignored.

However as I started to work on it, I felt that there were a few things that could be polished further wrt dtype assignment, hence some of the additional modifications in `LinearOperator` that are technically **breaking changes** - whilst I do believe the likelihood of breaking someone's code are really minimal as it would have to come from some doing some rather unusual stuff with `Op.dtype`, I would appreciate another opinion on this before I pull the trigger 🔫

